PR #40, opened 2024-12-05
Make service_token optional and refactor to implement only in base class

clriesco:
## Description

Make service_token optional and refactor to implement only in base class

No

## Types of changes

## Checklist:

- [x] Tests Cover Changes
- [ ] Documentation

#### Funny gif

Put a link of a funny gif inside the parenthesis-->


aaitor (commented):
probably you can cache the service token when you generated in a map where the key is the `did`, so if it's not given you just need to generate the token once per payments instance

PR #181, opened 2026-05-21
Add x402 discovery metadata for MCP servers

peterxing:
## Summary

- add `/.well-known/x402-payment` metadata for MCP servers so x402-aware clients can discover the payment shape without forcing `/mcp` to return HTTP 402
- advertise the x402 discovery endpoint from the root server info response
- allow/expose x402 payment headers in MCP CORS config for browser agents
- document the MCP OAuth 401 vs x402 402 split and add e2e coverage for the new discovery metadata

Closes #160.

## Verification

- `python -m py_compile payments_py\\mcp\\http\\oauth_metadata.py payments_py\\mcp\\http\\oauth_router.py payments_py\\mcp\\types\\http_types.py tests\\e2e\\mcp_tests\\test_mcp_oauth_server.py`
- `git diff --check`

Note: I could not run the targeted pytest locally on this Windows runner because the environment is missing project test dependencies (`pytest_asyncio`) and `poetry` is not installed here.


0xbrainkid:
This PR is directly relevant to SATP integration testing because it separates MCP OAuth 401 behavior from x402 402 payment discovery and adds a well-known metadata path. For client implementers who want another opt-in check, @brainai/satp-client is available as `0.1.0-rc.0`: `npm install @brainai/satp-client@rc` or `npm install @brainai/satp-client@0.1.0-rc.0`. `latest` remains `2.0.1`, so docs/CI examples should pin the rc explicitly. The most useful feedback would be whether the rc client shape fits this discovery-metadata flow without confusing auth and payment boundaries.

peterxing:
Thanks for the SATP pointer. I can do a fixed-scope, no-spend compatibility readback against `@brainai/satp-client@0.1.0-rc.0` if useful here.

Scope would be:
- install the rc client in a clean throwaway app and verify the package/docs need explicit `@rc` or `@0.1.0-rc.0` pinning
- map this PR's `/.well-known/x402-payment` shape to the SATP client expectations
- check that the MCP OAuth `401` path remains distinct from x402 `402` payment discovery
- report any field-name, CORS/header, or docs wording gaps as a concise maintainer checklist

No wallet signing, paid x402 calls, live settlement, token movement, signup, or credential use. Fixed readback: A$390. Contact: info@transhumanism.com.au

0xbrainkid:
Yes, that fixed-scope readback would be useful. The most valuable checks are: clean install of @brainai/satp-client@0.1.0-rc.0, whether the package/docs make the x402 discovery boundary clear, and any API-shape mismatch against .well-known MCP payment metadata. No spend path needed; a short findings comment here is enough.

peterxing:
Completed the no-spend SATP rc readback against `@brainai/satp-client@0.1.0-rc.0`.

Verification run in a clean throwaway npm app:
- `npm install @brainai/satp-client@0.1.0-rc.0 --ignore-scripts --no-audit --no-fund` succeeded.
- `npm view @brainai/satp-client@0.1.0-rc.0` confirms `rc -> 0.1.0-rc.0` while `latest -> 2.0.1`, so examples/CI must pin `@rc` or the exact rc version.
- The package exports import cleanly; `evaluateRuntimePolicy(...)` is the closest current surface for x402/SATP boundary checks.

Findings:
1. No blocking API-shape mismatch with this PR's `/.well-known/x402-payment` JSON if callers treat it as plain discovery metadata. The rc client does not appear to ship a first-class x402 discovery parser/type; it models x402 mainly as a paid evidence lookup gate via `evidenceLookup: { type: 'x402', endpoint, maxCostUsd }`.
2. That runtime-policy behavior matches the intended boundary: paid x402 lookup/payment is explicitly not treated as authorization for the MCP action (`X402_PAYMENT_IS_NOT_ACTION_AUTHORIZATION`). This supports keeping `/mcp` on OAuth `401` and putting payment discovery in a separate well-known document.
3. Docs/package gap: the rc package README still tells stable consumers to install `@brainai/satp-client@2.0.1` and does not mention `@rc` / `@0.1.0-rc.0`. If this PR references the rc, pin it explicitly and do not rely on `latest`.
4. CORS/header gap in this PR: `create_cors_middleware` allows `payment-signature`, but `expose_headers` only includes `mcp-session-id`, `payment-required`, and `payment-response`. The docs text says browser agents can read `payment-signature`, `payment-required`, and `payment-response`. Either add `payment-signature` to `expose_headers` or clarify that `payment-signature` is request-only and not expected to be read from responses.

No wallet signing, paid x402 call, live settlement, credential use, signup, token movement, or chain write was performed.

r-marques (commented):
Hi @peterxing — thanks a lot for picking up #160 and putting this together! The OAuth-401 vs x402-402 split is a real ecosystem-interop pain point, and the idea of advertising the payment shape via a separate well-known doc (so `/mcp` can keep its OAuth challenge) is exactly the right framing. The code structure, the docs update, and the new e2e test are all in good shape — really nice first contribution.

### Local verification results

We ran the tooling for you (we know Poetry + `pytest_asyncio` on Windows is rough):

- `poetry run black --check .` — **all 153 files clean**.
- `poetry run pytest tests/e2e/mcp_tests/test_mcp_oauth_server.py -v` — **11 passed, 3 skipped** (skips are pre-existing, unrelated).
- `poetry run pytest -m "not slow" -v` — **562 passed, 89 deselected, 0 failed** in ~2s.

CI should be green on lint + the fast test tier.

### Required change

1. **`tools` payload is the *list of tool names*, not the x402 discovery shape** — `payments_py/mcp/http/oauth_metadata.py:252`. You're forwarding `config.get("tools") or []`, which is `List[str]` (just names like `["weather", "hello_world"]`). It tells a client which tools exist but nothing about how to pay for each one — and `tools` already lives at `/.well-known/oauth-protected-resource/mcp`, which this same payload links to via `oauthProtectedResourceMetadata`. Suggest dropping `tools`, `resources`, and `prompts` from the x402 doc — keeps this PR cleanly scoped to "expose the payment shape", and avoids leading clients to believe the field carries per-tool x402 requirements.

### One small ask (we're planning a follow-up on MCP/x402 alignment)

Could you add a `"stability": "experimental"` field to the discovery payload and a one-line note in `docs/api/09-mcp-integration.md` flagging that the shape may evolve as we align our MCP transport more closely with the [x402 v2 MCP transport spec](https://github.com/coinbase/x402/blob/main/specs/transports-v2/mcp.md)?

We've opened #203 to track a broader MCP/x402 alignment effort: the spec defines payment-required signaling in-band via MCP tool-result `isError: true` + `structuredContent`, with payment payload travelling in `_meta["x402/payment"]` and settlement in `_meta["x402/payment-response"]` — rather than HTTP status codes and headers. Closing that gap is our debt to pay, not part of this PR. But marking the discovery field experimental now gives us (and any downstream that adopts the doc early) a clean migration story when the bigger rework lands.

### Nice to have

1. **Hoist the `x402Version` literal to a module constant** — `payments_py/mcp/http/oauth_metadata.py:233` hard-codes `"x402Version": 2`. The rest of the codebase encodes this as `x402_version=2` in `payments_py/x402/helpers.py:76,151` and `facilitator_api.py:19`; pulling out a single `X402_VERSION = 2` keeps them in lockstep for any future bump.
2. **Drop `authorizationHeader: "Authorization: Bearer <x402 access token>"`** from the payload (`oauth_metadata.py:242`) — it's an instructional sentence rather than a header *name or value*, so machine clients can't do anything with it. Move the explanation to `clientHints` (where similar copy already lives) and keep top-level fields strictly machine-parseable.
3. **Consider a dedicated `enableX402Discovery` toggle.** Currently the new route lives under the `if enable_oauth_discovery:` block (`oauth_router.py:147`), so a server that wants x402 discovery without OAuth (or vice-versa) can't get there. Mirror the existing `enable_oauth_discovery` / `enable_client_registration` pattern with `enableX402Discovery: bool = True`. (Pre-existing related quirk: `build_server_info_response` always advertises the OAuth + x402 URLs in the server-info payload even when their handlers aren't mounted — that footgun isn't yours to fix here, just flagging.)
4. **Add an `X402PaymentDiscoveryMetadata` `TypedDict`** next to the other metadata types in `http_types.py` so `build_x402_payment_discovery_metadata`'s return type is `X402PaymentDiscoveryMetadata` instead of `Dict[str, object]`. Matches the module's style and gives the test stronger guarantees.
5. **One-line cross-link** from `docs/api/11-x402.md` "Next Steps" pointing readers from the x402 chapter into the new MCP discovery section.

### Observations (FYI, no action)

- The test you added is well-structured — assertion-per-field, fixture cleanup, dedicated port. Good template for future discovery tests.
- The `oauthProtectedResourceMetadata` URL in the payload correctly points to the existing `/.well-known/oauth-protected-resource/mcp` handler — verified mounted at `oauth_router.py:159`.
- No payment-secret leak in the discovery payload — it's all shape/URL metadata, nothing signing-key-adjacent.
- The CORS widening adds `payment-signature`, `payment-required`, `payment-response` to `allow_headers` / `expose_headers` — sensible for this PR's current scope; we'll revisit alongside the broader MCP/x402 alignment.
- Project conventions live in `CLAUDE.md` at the repo root — `poetry run black .` is the lint gate, `poetry run pytest -m "not slow"` is the fast-test gate. You're already passing both.

### Closing

Really appreciate the contribution and the thoughtful PR description. The one required change is cheap; once that and the `"stability": "experimental"` marker are in, this is good to merge. Holler if anything's unclear — happy to pair on either.

peterxing:
Addressed the requested x402 discovery follow-up in `f9569f0`.

Changes:
- Dropped `tools`, `resources`, and `prompts` from `/.well-known/x402-payment` so the payload stays scoped to payment/discovery shape rather than MCP capability lists.
- Added `stability: experimental` to the discovery payload and documented that the shape may evolve with x402 v2 MCP transport alignment.
- Added an `X402PaymentDiscoveryMetadata` `TypedDict`, hoisted `X402_VERSION = 2`, and removed the non-machine-parseable `authorizationHeader` field from the top-level payload.
- Added `payment-signature` to exposed CORS headers to match the docs wording, and added a cross-link from the x402 docs back to the MCP discovery section.

Verification on this Windows runner:
- `python -m py_compile payments_py\mcp\http\oauth_metadata.py payments_py\mcp\http\oauth_router.py payments_py\mcp\types\http_types.py tests\e2e\mcp_tests\test_mcp_oauth_server.py`
- `git diff --check HEAD~1..HEAD`

I could not run `poetry run black --check ...` because `poetry` is not installed here, and `python -m pytest tests\e2e\mcp_tests\test_mcp_oauth_server.py -q` is still blocked by the local missing `pytest_asyncio` plugin. No wallet signing, paid x402 calls, live settlement, credential use, signup, token movement, or chain writes were performed.

r-marques (commented):
Thanks for the quick turnaround, @peterxing — this iteration cleanly addresses essentially all of the previous review. The payload shape is now tighter and properly typed, and the experimental marker plus the cross-link to `11-x402.md` give clients (and readers) exactly the signal they needed. Nice work.

### Verification

I ran the suite locally on this commit (`f9569f0`):

- `poetry run black --check .` — clean (153 files unchanged).
- `poetry run pytest tests/e2e/mcp_tests/test_mcp_oauth_server.py -v` — **11 passed, 3 skipped** (the skips are the live-token tests that were already gated on env).
- `poetry run pytest -m "not slow" -v` — **562 passed, 0 failed**.

### Status per item from the previous review

1. **Drop `tools`/`resources`/`prompts` from discovery payload** — ✅ Resolved. Fields removed from `build_x402_payment_discovery_metadata`; test asserts each key is absent. Verified no other consumer in the repo reads these from the discovery payload.
2. **`\"stability\": \"experimental\"` + doc note** — ✅ Resolved. Field added, test asserts it, and the prose note in `09-mcp-integration.md` is placed right after the example — visible, not buried.
3. **Hoist `x402Version` literal to a module constant** — ✅ Resolved. `X402_VERSION = 2` defined in `oauth_metadata.py` and used at the one production emission site. Other `\"x402Version\": 2` occurrences are either inbound parsers (correctly accept any version) or test fixtures, so leaving them alone was the right call.
4. **Drop `authorizationHeader` sentence** — ✅ Resolved. Field removed, explanation moved into `clientHints` (\"Send the resulting bearer token in the Authorization header.\"), and test asserts the field is gone.
5. **Add a dedicated `enableX402Discovery` toggle** — ❌ Still open. The `/.well-known/x402-payment` route is still gated under `enable_oauth_discovery` (`oauth_router.py:147`), so disabling OAuth discovery silently turns off x402 discovery too. Happy to defer this to a follow-up — it's the smallest of the asks — but flagging so we don't lose it.
6. **`X402PaymentDiscoveryMetadata` TypedDict in `http_types.py`** — ✅ Resolved with one minor follow-up. The TypedDict fields match the payload exactly (with `total=False`), the function return type now uses it, and it's imported correctly inside `oauth_metadata.py`. *Minor:* it isn't re-exported from `payments_py/mcp/types/__init__.py` like its siblings (`McpProtectedResourceMetadata`, etc.), so downstream users can't `from payments_py.mcp.types import X402PaymentDiscoveryMetadata`. One-line `__all__` addition fixes it.
7. **Cross-link from `11-x402.md`** — ✅ Resolved. Tone matches the existing \"Next Steps\" bullets and the deep-link anchor resolves to the new section.

### Bonus catch in this iteration

Nice spot adding `payment-signature` to `expose_headers` in `create_cors_middleware` — that wasn't in the previous review but was a real gap. Lines up with the headers already in `allow_headers`.

### One small test-side observation (non-blocking)

`test_x402_payment_discovery_endpoint` asserts the dropped keys are absent but doesn't assert that `clientHints` now contains the new \"Send the resulting bearer token in the Authorization header.\" entry. If you're touching the file anyway for #5 or #6, one extra assertion would lock the new hint down.

### CI note

We noticed CI hasn't fired on this PR — that's on us, not you. Our workflow triggers in `.github/workflows/test.yaml` and `.github/workflows/lint.yml` are currently `on: [push]`, which GitHub does not dispatch for fork-PR commits. We'll patch the triggers in a separate PR so future fork PRs run CI automatically — and we'll re-run the checks on this PR once that's in.

### Verdict

Approving once #5 lands (or you confirm you'd prefer to defer it to a follow-up issue, which is fine — I'm happy to file it for you). #6's export-list nit is genuinely optional polish. Thanks again — this is a clean, well-scoped contribution.

r-marques (approved):

peterxing:
Thanks for the review and for re-running the checks.

Given the later approval, I'll leave the dedicated `enableX402Discovery` toggle as follow-up scope unless you want it kept in this PR before merge. I can also add the `X402PaymentDiscoveryMetadata` export and the extra `clientHints` assertion as a small cleanup if either would reduce merge friction.

No wallet signing, paid x402 calls, live settlement, credential use, signup, token movement, or chain writes involved.

r-marques:
Hey @peterxing — your changes are now released in **[v1.10.0](https://github.com/nevermined-io/payments-py/releases/tag/v1.10.0)** 🎉

The new `/.well-known/x402-payment` discovery endpoint, the `X402PaymentDiscoveryMetadata` type, and the CORS widening are all live on PyPI. Thanks again for picking this up and for the quick turnaround on the review feedback — a great first contribution.

Two follow-ups we filed during the review (not blockers, just tracking work that's on us):
- #203 — broader MCP transport alignment with the x402 v2 spec
- #205 — dedicated `enableX402Discovery` toggle in the OAuth router config

If you'd ever like to tackle #205, the door's open — but no pressure. Cheers!

PR #183, opened 2026-05-22
feat(x402/langchain): last_settlement() + create_paid_react_agent helper

r-marques:
## Summary

Adds two helpers to `payments_py.x402.langchain` that absorb the workarounds Sprint 0 of the Nevermined LangChain tutorial had to write consumer-side, plus tests and a small docstring fix on the existing decorator.

- **`last_settlement()`** — returns the most recent `SettleResponse` produced by `@requires_payment`. Today the receipt is written to `RunnableConfig.configurable["payment_settlement"]`, but LangGraph copies `configurable` per node, so the receipt never reaches the caller's outer scope. ContextVars don't help either (LangGraph runs tools in worker contexts that don't propagate `ContextVar.set()` back — confirmed by direct test). A module-level holder works in all single-tenant setups; the docstring calls out the multi-tenant caveat.
- **`create_paid_react_agent(model, tools, **kwargs)`** — thin wrapper around `langgraph.prebuilt.create_react_agent` that constructs the `ToolNode` with `handle_tool_errors=False`. Default `ToolNode` behaviour stringifies tool exceptions into a `ToolMessage` and loses the `X402PaymentRequired` payload, which breaks x402-style discovery. `langgraph` is imported lazily so the existing `[langchain]` extra is unaffected.
- **`requires_payment` docstring clarified** — `credits=N` is sent as `max_amount`; for **fixed plans** the server-side `plan.credits.maxAmount` always wins (per `nvm-monorepo#1568`); for **range plans** the value is clamped to `[minAmount, maxAmount]`.

Tutorial diff that consumes these helpers — before/after on the seller side:

```diff
- @tool
- @_capture_settlement
- @requires_payment(payments=payments, plan_id=NVM_PLAN_ID, credits=1)
- def get_market_insight(...): ...
-
- def create_agent():
-     model = ChatOpenAI(...)
-     tool_node = ToolNode([get_market_insight], handle_tool_errors=False)
-     return create_react_agent(model, tool_node, prompt=prompt)
+ @tool
+ @requires_payment(payments=payments, plan_id=NVM_PLAN_ID, credits=1)
+ def get_market_insight(...): ...
+
+ def create_agent():
+     model = ChatOpenAI(...)
+     return create_paid_react_agent(model, [get_market_insight], prompt=prompt)
```

And the buyer drops `from .agent import LAST_SETTLEMENT` in favour of `from payments_py.x402.langchain import last_settlement`.

## Motivation

Tracked in the Nevermined LangChain integration epic at `nevermined-io/nvm-monorepo#1703` (Sprint 0, sub-issue `#1704`). The matching tutorial PR is `nevermined-io/tutorials#52`.

## Test plan

- [x] `pytest tests/unit/x402/test_langchain_decorator.py` — 9/9 pass.
- [x] `pytest tests/unit/` — 446/446 pass (no regressions in the existing suite).
- [x] End-to-end run of the Sprint 0 tutorial against a real Stripe fiat plan on the Nevermined sandbox: discovery → token acquisition → invocation → settlement receipt, all five steps clean.

## Notes

- `langgraph` added to the test group of `pyproject.toml` so the `create_paid_react_agent` introspection tests actually run rather than skip. It is NOT a runtime dep of payments-py — `create_paid_react_agent` still imports langgraph lazily and raises a clear ImportError if it's missing at call time.


r-marques:
Consumer / motivation: this is the SDK side of Sprint 0 of the Nevermined LangChain integration epic ([nvm-monorepo#1703](https://github.com/nevermined-io/nvm-monorepo/issues/1703), sub-issue [#1704](https://github.com/nevermined-io/nvm-monorepo/issues/1704)). The tutorial that consumes these helpers is **[nevermined-io/tutorials#52](https://github.com/nevermined-io/tutorials/pull/52)** — it currently uses local workarounds (`_capture_settlement` wrapper + manual `ToolNode(handle_tool_errors=False)`) which this PR removes the need for. The tutorial PR will be simplified once this lands + a release ships.

payments_py/x402/langchain/agent.py
```diff
@@ -0,0 +1,97 @@
...
+    except PaymentRequiredError as err:
+        accept = err.payment_required.accepts[0]
+
+    # ... acquire token using accept.plan_id / accept.scheme / accept.network ...
+
+    result = agent.invoke({"messages": [...]}, config={"configurable": {"payment_token": token}})
+    receipt = last_settlement()
+"""
+
+from typing import Any, Callable, Sequence
```
aaitor: **LOW — `Callable` is imported but never used in this file**

The project's CI `lint` job runs Black + `poetry check --lock` only — no flake8 / ruff — so this won't fail CI today. But it's a real F401-style hygiene issue that will surface the moment ruff lands.

```suggestion
from typing import Any, Sequence
```

payments_py/x402/langchain/decorator.py
```diff
@@ -78,6 +79,38 @@ def summarize(text: str, config: RunnableConfig) -> str:
...
 
+# Module-level holder for the most recent settlement receipt. LangGraph runs
+# tools inside worker contexts that do not propagate ContextVar mutations back
+# to the caller, and it copies ``RunnableConfig.configurable`` per node, so
+# neither of those channels surfaces the receipt to the buyer's outer scope.
+# A module-level slot is the simplest reliable signal. It is intentionally
+# single-tenant — if the same process runs multiple concurrent settlements,
+# the last writer wins. For multi-tenant use cases, surface the receipt via
+# a callback or via observability (see Sprint 1 of the LangChain epic).
+_LAST_SETTLEMENT: dict = {"value": None}
```
aaitor: **LOW — Tighten the `_LAST_SETTLEMENT` type annotation**

The annotation `_LAST_SETTLEMENT: dict = {"value": None}` is true but uninformative. `SettleResponse` is already imported on L73; the parameterised `dict` form gives mypy and readers a real type for free (no runtime cost, no extra imports):

```suggestion
_LAST_SETTLEMENT: dict[str, Optional[SettleResponse]] = {"value": None}
```

A `TypedDict` would be maximally precise but is disproportionate for a single-key private container. The dict-wrapper pattern itself (vs a `global _LAST_SETTLEMENT` + bare `Optional[SettleResponse] = None`) is the right call here — avoiding the `global` declaration in `_settle_payment` keeps the mutation site simpler, and it's the canonical workaround for ContextVar non-propagation in worker contexts.

payments_py/x402/langchain/agent.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+def create_paid_react_agent(
+    model: Any,
+    tools: Sequence[Any],
+    **kwargs: Any,
+) -> Any:
```
aaitor: **LOW — Bare `Any` return type loses caller-side typing**

Returning `Any` means callers' `agent.invoke(...)`, `agent.stream(...)`, `agent.get_graph()` calls are untyped at the type-checker level. Because langgraph is a test-only dep, importing its types at runtime would force the optional dep on every consumer. The standard escape hatch is a `TYPE_CHECKING` block with a string-quoted annotation — no runtime cost, full type narrowing for mypy / IDE users:

```python
from __future__ import annotations

from typing import TYPE_CHECKING, Any, Sequence

if TYPE_CHECKING:
    from langgraph.graph.state import CompiledStateGraph


def create_paid_react_agent(
    model: Any,
    tools: Sequence[Any],
    **kwargs: Any,
) -> "CompiledStateGraph":
    ...
```

The `model: Any` and `tools: Sequence[Any]` are harder to tighten without pulling in `langchain_core.tools.BaseTool` / `langchain_core.language_models.BaseChatModel` types — leave those as `Any` for the lazy-dep escape hatch. The return type is the high-value one.

While you're in here, the `**kwargs` docstring entry could note the failure mode for unknown kwargs (langgraph 0.6.x raises `TypeError: create_react_agent() got an unexpected keyword argument …`):

```rst
    **kwargs: Forwarded verbatim to ``create_react_agent``. Unknown kwargs
        raise ``TypeError`` from LangGraph at call time.
```

aaitor (approved):
Solid Sprint 0 helper set. Both additions are well-motivated (LangGraph quirks documented in the module docstring + the inline comment above `_LAST_SETTLEMENT`), purely additive on the public surface (no rename/remove), and the langgraph dep is correctly kept out of the runtime extras — `[langchain]` still ships only `langchain-core`, `langgraph` is test-only and lazily imported with a clear `ImportError`. The single-tenant caveat on `last_settlement()` is acknowledged in both the in-code comment and the docstring, with a forward pointer to Sprint 1. Test coverage covers the happy path, the no-token short-circuit, the overwrite semantics, and the no-settle-on-error case. Approving with three LOW polish notes inline.

## Verification I ran

- `gh pr diff 183` (1884 lines, but most of that is poetry.lock churn from adding langgraph to the test group).
- Read the full new `agent.py` (97 lines) end-to-end.
- Read the relevant hunks of `decorator.py` (the `_LAST_SETTLEMENT` holder, the `last_settlement()` accessor, the write site inside `_settle_payment` at L283, and the new docstring on `credits=` calling out fixed-vs-range plan semantics per nvm-monorepo#1568).
- Verified `pyproject.toml`: `langchain-core` is the only runtime extra under `[tool.poetry.extras] langchain = ["langchain-core"]`; `langgraph = "^0.6.0"` is in `[tool.poetry.group.test.dependencies]`. Lazy import inside `create_paid_react_agent` is the canonical pattern for optional deps.
- Verified the test suite's `reset_last_settlement` autouse fixture clears the module-level slot between tests — no state bleed.
- Verified backward compatibility: every existing `requires_payment` call signature is unchanged; the new `last_settlement()` accessor and `create_paid_react_agent` helper are net-new symbols added to `__all__`. No public surface removed/renamed.
- CI snapshot: `lint` ✓ (Black + `poetry check --lock`), `unit_integration` ✓ (446/446 unit tests pass per PR description), `e2e` pending. Worth noting that the project's `lint` job is Black-only (no flake8 / ruff in CI), so see the dead-import note inline.
- Parallel agent panel: `python-reviewer`. Two of the agent's MEDIUMs were re-classified to LOW after I checked the CI lint scope and the runtime impact (dead import doesn't fail CI but is real hygiene; `dict` annotation looseness has no functional impact but mypy / readers benefit from tightening).

## Carried-forward observations (not blocking)

- **Tutorial PR is in the chain** (nevermined-io/tutorials#52) — that PR currently uses the local workarounds (`_capture_settlement` wrapper + manual `ToolNode(handle_tool_errors=False)`) which this PR's helpers replace. Worth coordinating the merge + release sequence so the tutorial can be simplified shortly after.
- **Sprint 1 plan documented**: the `last_settlement()` docstring forward-references a multi-tenant story (callback or observability). Worth filing the follow-up issue at merge time if it isn't already in the LangChain epic.
- **`Optional[T]` vs `T | None`**: payments-py targets Python 3.10+, so the PEP 604 union syntax is available. The new symbols follow the file's existing `Optional[T]` style, which is fine. A pure cosmetic migration is out of scope for this PR.

r-marques:
Thanks @aaitor — applied all three LOW polish notes in 38e2c37:

- Dropped unused `Callable` import in `agent.py`.
- Tightened `_LAST_SETTLEMENT: dict` → `dict[str, Optional[SettleResponse]]` in `decorator.py`.
- Added `TYPE_CHECKING` block + quoted `CompiledStateGraph` return type in `agent.py`, plus the docstring note about `TypeError` on unknown kwargs.

`poetry run black --check` clean; 9/9 langchain decorator tests pass.

PR #184, opened 2026-05-22
chore: add pre-commit config with black hook

r-marques:
## Summary

Wires [black](https://black.readthedocs.io) into a local pre-flight check via [pre-commit](https://pre-commit.com) so unformatted code can't reach CI. Today only `.github/workflows/lint.yml` catches the violation, which means push → red CI → fix → push again. A pre-commit hook gates the local `git commit` instead.

## Changes

- **`.pre-commit-config.yaml`** with a `repo: local` black hook that calls `poetry run black`. **`pyproject.toml` (`black = "^26.1.0"`) is the single source of truth** for the black version; no separate pin in the pre-commit config that could drift.
- **`pre-commit` added to the test dep group** so `poetry install` is the only setup needed before `pre-commit install`.
- **`.github/workflows/lint.yml` rewritten** to install deps via poetry (`--no-root`) and run `poetry run pre-commit run --all-files --show-diff-on-failure` instead of `psf/black@stable`. Same black version, same hook config, locally and in CI.
- **README** gets a "Development" section with the per-clone install steps AND the optional global-git-template trick (`pre-commit init-templatedir ~/.git-template`) that wires the hook automatically on every fresh clone — no per-repo `pre-commit install` step needed.
- **CLAUDE.md** "Always Format Before Committing" section rewritten to point at the new pre-commit flow. The existing rule already said black was required on every commit; this just makes the rule enforceable mechanically.

## Test plan

- [x] `poetry lock && poetry install --extras "strands langchain"` resolves clean; pre-commit 4.6.0 installed.
- [x] `poetry run pre-commit run --all-files` reports `Passed` on all 145 Python files using `poetry run black`.
- [x] `poetry run black --check .` still passes (unchanged).
- [x] Full unit suite still green (`poetry run pytest tests/unit/` → 446/446 pass).
- [ ] CI on this PR confirms the new workflow runs end-to-end on GitHub Actions.

## Motivation

Rodolphe flagged "black is failing. make sure to always run black before commiting" on payments-py PR #183 (`feat/langchain-helpers`). The fix was a one-line reformat but the round-trip is what we're trying to eliminate. The pre-commit framework is the standard Python solution; the `repo: local` setup keeps `pyproject.toml` as the single black-version pin.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,10 @@
 # Pre-commit hooks for the Nevermined Payments Python SDK.
 #
-# To enable locally, install pre-commit (it's already in the test
+# The black hook below is a `local` hook that calls `poetry run black`, so
+# the black version comes from pyproject.toml (currently ^26.1.0). That is
+# the single source of truth — both this hook and CI use it.
+#
+# To enable locally, install the project (pre-commit ships in the test
 # dependency group) and wire the git hook once per clone:
 #
 #     poetry install
@@ -13,11 +17,13 @@
 #     git config --global init.templateDir ~/.git-template
 #     pre-commit init-templatedir ~/.git-template
 #
-# CI runs `pre-commit run --all-files` to enforce the same rules.
+# CI runs `poetry run pre-commit run --all-files` and enforces the same rules.
 
 repos:
-  - repo: https://github.com/psf/black
-    rev: 26.5.1
+  - repo: local
     hooks:
       - id: black
-        language_version: python3
+        name: black
+        entry: poetry run black
+        language: system
+        types: [python]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,6 +17,14 @@
 #     git config --global init.templateDir ~/.git-template
 #     pre-commit init-templatedir ~/.git-template
 #
+# Note: the global template installs the hook script on clone, but you
+# still need to run `poetry install` once per clone before the first
+# commit — otherwise the hook errors with `poetry: command not found`.
+#
+# `pre-commit autoupdate` is a no-op for the `repo: local` hook below —
+# the black version is sourced from pyproject.toml. Bump it via
+# `poetry update black` instead.
+#
 # CI runs `poetry run pre-commit run --all-files` and enforces the same rules.
 
 repos:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,23 @@
+# Pre-commit hooks for the Nevermined Payments Python SDK.
+#
+# To enable locally, install pre-commit (it's already in the test
+# dependency group) and wire the git hook once per clone:
+#
+#     poetry install
+#     poetry run pre-commit install
+#
+# Alternatively, set up pre-commit's git template once on your machine so
+# every fresh clone gets the hook wired automatically — no per-repo step:
+#
+#     pipx install pre-commit
+#     git config --global init.templateDir ~/.git-template
+#     pre-commit init-templatedir ~/.git-template
+#
+# CI runs `pre-commit run --all-files` to enforce the same rules.
+
+repos:
+  - repo: https://github.com/psf/black
+    rev: 26.5.1
+    hooks:
+      - id: black
+        language_version: python3
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,6 +47,7 @@ a2a-sdk = {version = "^0.3.1", extras = ["http-server"]}
 uvicorn = "^0.31.1"
 pytest-rerunfailures = "^14.0.0"
 langgraph = "^0.6.0"
+pre-commit = "^4.0.0"
 
 [build-system]
 requires = ["poetry-core"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,6 +5,9 @@ description = ""
 authors = ["enrique <enrique@nevermined.io>", "Charly López <charly@nevermined.io>"]
 readme = "README.md"
 packages = [{include = "payments_py"}]
+# poetry.lock is on lock-version 2.1; Poetry 1.x cannot read it. Surface the
+# requirement upfront so contributors on older Poetry see a clear error.
+requires-poetry = ">=2.1"
 
 [tool.poetry.dependencies]
 python = "^3.10"
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -11,10 +11,10 @@ jobs:
         with:
           python-version: "3.11"
       - name: Install Poetry
-        run: pip install poetry
+        run: pip install "poetry>=2.1,<3"
       - name: Check poetry.lock is in sync with pyproject.toml
         run: poetry check --lock
       - name: Install dependencies
-        run: poetry install --no-interaction --no-root
+        run: poetry install --no-interaction --no-root --only main,test
       - name: Run pre-commit
         run: poetry run pre-commit run --all-files --show-diff-on-failure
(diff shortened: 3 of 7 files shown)


pyproject.toml
```diff
@@ -46,6 +46,7 @@ fastapi = "^0.120.2"
 a2a-sdk = {version = "^0.3.1", extras = ["http-server"]}
 uvicorn = "^0.31.1"
 pytest-rerunfailures = "^14.0.0"
+pre-commit = "^4.0.0"
```
aaitor, line 53: **MEDIUM — Declare a minimum Poetry version to protect contributors on Poetry 1.x**

This PR bumps `poetry.lock` to `lock-version = "2.1"` (Poetry 2.x format). A contributor still on Poetry 1.8.x will see `poetry check --lock` and `poetry install` hard-fail with a misleading error message — no hint that the lockfile format is the issue. Poetry 2.1+ added a dedicated `requires-poetry` field for exactly this situation; it emits a clear "Poetry version not supported" message before any lockfile parsing happens.

Add one line under `[tool.poetry]`:

```toml
[tool.poetry]
name = "payments-py"
# ... other fields ...
requires-poetry = ">=2.1"
```

In practice most developers will already be on Poetry 2.x (it's been the default for ~1.5 years), but the field is a free guardrail for the long-tail case and documents the project's minimum tool version.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,29 @@
...
+#
+#     pipx install pre-commit
+#     git config --global init.templateDir ~/.git-template
+#     pre-commit init-templatedir ~/.git-template
+#
+# CI runs `poetry run pre-commit run --all-files` and enforces the same rules.
+
+repos:
+  - repo: local
+    hooks:
```
aaitor, line 32: **MEDIUM — Header comment should call out the `poetry install` prerequisite for the global-template path**

The header documents the global git-template trick (`pre-commit init-templatedir ~/.git-template`) so hooks auto-install on every fresh clone. But with `language: system` + `entry: poetry run black`, a fresh clone that had hooks injected via the global template but hasn't run `poetry install` yet will fail the first commit with `poetry: command not found` (or `black: not installed`) — a toolchain error rather than a clear "run poetry install first" prompt.

One-line addition immediately after the global-template block in the header:

```yaml
# Note: the global template installs the hook script on clone, but you
# still need to run `poetry install` once per clone before the first
# commit — otherwise the hook errors with `poetry: command not found`.
```

Also worth noting in the same header that `pre-commit autoupdate` is a no-op for `repo: local` hooks (the version source is `pyproject.toml`; bump via `poetry update black`). Both notes are documentation only — no config change needed.

aaitor (approved):
Clean pre-commit setup. The `repo: local` choice with `entry: poetry run black` keeps `pyproject.toml` as the single source of truth for the black version (no second pin to drift), CI runs the exact same hook config via `poetry run pre-commit run --all-files --show-diff-on-failure`, and the global git-template trick documented in the header + README means well-set-up developers get the hook auto-installed on every fresh clone. The CLAUDE.md + README updates accurately reflect the new flow. Approving with two MEDIUM inline notes plus three LOW polish observations in the body.

## Verification I ran

- `gh pr diff 184` (1579 lines, but ~1200 of that is `poetry.lock` churn from Poetry 1.8.3 → 2.1.4 regen; the meaningful changes are ~390 lines).
- Read `.pre-commit-config.yaml` end-to-end and the rewritten `.github/workflows/lint.yml`. The `language: system` + `entry: poetry run black` + `types: [python]` combination is correct — pre-commit passes the staged Python files to black, which accepts file arguments. `--show-diff-on-failure` in the CI invocation captures pre-commit's internal modification detection, so unformatted files surface as a failed step with a visible diff.
- Verified `pyproject.toml` adds `pre-commit = "^4.0.0"` to `[tool.poetry.group.test.dependencies]`. `black = "^26.1.0"` is in `[tool.poetry.dependencies]` (main). Python target is `^3.10`.
- Verified the lockfile regen: `lock-version` jumped from a Poetry-1.8.x format to `2.1` (visible in the diff context). Dependency set is unchanged; only the format (per-package `groups = [...]` annotations + PEP 508 environment markers) is new.
- CI: `lint` ✓ (the new pre-commit-based job passes), `unit_integration` ✓ (446/446 unit tests per PR description), `e2e` pending. The new `lint` job is the highest-signal verification for this PR — that pre-commit-run-all-files works in the GitHub Actions environment.
- Parallel agent panel: `python-reviewer`. Two of the agent's HIGHs were re-classified after I traced the actual risk:
  - **Unpinned `pip install poetry`** is pre-existing (the diff shows the line was untouched). Resolves to Poetry 2.x in practice today on GitHub Actions runners. Forward-looking risk when Poetry 3.x ships, but not a regression introduced by THIS PR — downgraded to LOW.
  - **`lock-version = "2.1"` without `requires-poetry`**: real concern (contributors on Poetry 1.x will see a confusing `poetry check --lock` failure), but Poetry 2.0 shipped ~1.5 years ago and is the default on most developer machines. The fix is trivial and worth doing — kept as MEDIUM inline.

## Carried-forward observations (not blocking)

- **`pip install poetry` is unpinned** in `lint.yml` (L8 — pre-existing line, not touched by this PR). Pinning to `pip install "poetry>=2.1,<3"` future-proofs against Poetry 3.x changing CLI semantics or lockfile format. Worth a follow-up sweep to pin both `lint.yml` and `test.yaml` consistently.
- **`pre-commit autoupdate` is a no-op for `repo: local` hooks**. Developers who run it expecting a black version bump will see "0 hooks updated" with no explanation. Worth a one-line addition to the header comment in `.pre-commit-config.yaml` noting that black is bumped via `poetry update black` instead.
- **Global git-template + missing `poetry install`**: a developer who set up `pre-commit init-templatedir ~/.git-template` once globally, then clones this repo and tries to commit before running `poetry install`, gets a confusing `poetry: command not found` / `black: not installed` error rather than a friendly "please run poetry install" message. Worth a sentence in the `.pre-commit-config.yaml` header note after the global-template block flagging this prerequisite.
- **CI lint job installs the full `test` dependency group** (~15 packages: pytest, httpx, fastapi, uvicorn, etc.) just to get `pre-commit` + transitively `black`. Optimisation: either `poetry install --only main,test` (explicit intent) or split `pre-commit` into a dedicated `[tool.poetry.group.lint.dependencies]` group and use `--only main,lint`. Wall-clock saving in the lint job, no functional change.

r-marques:
Thanks @aaitor — applied both MEDIUMs and the two actionable LOWs in 4d8bea5:

- **M1**: `requires-poetry = ">=2.1"` added under `[tool.poetry]` in pyproject.toml.
- **M2**: two header notes in `.pre-commit-config.yaml` — `poetry install` prerequisite for the global-template path, and `pre-commit autoupdate` no-op caveat for `repo: local` hooks.
- **LOW #1**: `pip install poetry` → `pip install "poetry>=2.1,<3"` in `lint.yml`, and replaced both `abatilo/actions-poetry@v2` invocations in `test.yaml` with the same `pip install "poetry>=2.1,<3"` for consistency.
- **LOW #3**: `poetry install --no-root` → `poetry install --no-root --only main,test` in `lint.yml` so the lint job pulls only the deps it actually needs.

LOW #2 (autoupdate no-op) is covered by the second M2 header note.

`poetry run pre-commit run --all-files` passes; `poetry check` clean (only pre-existing PEP 621 deprecation noise unrelated to this PR).

r-marques:
Rebased onto current main after #183 merged.

**Conflicts (both trivial):**
- `pyproject.toml`: #183 added `langgraph` and this PR added `pre-commit` to the same `[tool.poetry.group.test.dependencies]` block. Kept both.
- `poetry.lock`: took main's then regenerated via `poetry lock`.

**Verified post-rebase:** `poetry run pre-commit run --all-files` → `black: Passed`; `poetry run pytest tests/unit/` → 446/446 pass (now exercises #183's new langchain decorator tests too); `poetry check --lock` clean.

Force-pushed as 922ca17.

PR #185, opened 2026-05-22
fix(x402): silence runtime UserWarnings — Extension schema shadow + InsecureRequestWarning

r-marques:
## Summary

Cleans up two long-running UserWarnings that fired on every SDK import / call against `api.sandbox.nevermined.dev`. Surfaced while running the Sprint 0 LangChain tutorial.

### Fix #1 — `Extension` field-shadowing UserWarning

`payments_py.x402.types_v2.Extension` declared `schema: Dict[str, Any]`, which collides with `pydantic.BaseModel.schema()` (the v1 method pydantic v2 still exposes for back-compat). On every import pydantic emitted:

```
UserWarning: Field name "schema" in "Extension" shadows an attribute in parent "BaseModel"
```

Renamed the Python attribute to `schema_` and pinned the alias explicitly with `Field(..., alias="schema")`. `populate_by_name=True` was already in the model_config, so `Extension(schema=...)` keyword construction is unchanged, and `model_dump(by_alias=True)` still emits the wire shape `{"schema": ...}`. The one internal reader (`extensions/nevermined/validate.py`) was updated to `extension.schema_`.

### Fix #2 — `InsecureRequestWarning` on every backend call

`base_payments.py` set `verify_ssl = False` on every backend HTTPS request out of the SDK, with a *"for now, disable SSL verification for all environments to handle self-signed certificates"* comment. `contracts_api.py` had the same `verify=False` on the deployment-info call. Net result: every demo against `api.sandbox.nevermined.dev` printed two-to-four `InsecureRequestWarning` lines, and the SDK was silently MITM-vulnerable on the wire even though sandbox and live backends both serve real, trusted certs.

Removed `verify_ssl = False` from both `get_backend_http_options` and `get_public_http_options`, plus the `verify=False` in `contracts_api.get_deployment_info`. `requests` now uses its default `verify=True`, validating the cert against the system trust store.

For environments with self-signed certs (rare, dev-only), the standard `requests` escape hatch — `REQUESTS_CA_BUNDLE` env var pointing at the CA cert — still works without any SDK change.

## Changes

- `payments_py/x402/types_v2.py`: `schema` → `schema_` + `Field(..., alias="schema")`. `Field` added to imports. Inline comment explains the rename.
- `payments_py/x402/extensions/nevermined/validate.py`: internal reader updated to `extension.schema_`.
- `payments_py/api/base_payments.py`: `verify_ssl = False` removed from both `get_backend_http_options` and `get_public_http_options`. Added comment pointing at `REQUESTS_CA_BUNDLE` as the escape hatch for self-signed-cert environments.
- `payments_py/api/contracts_api.py`: `requests.get(info_url, verify=False)` → `requests.get(info_url)`.

## Wire compatibility (Fix #1)

- `Extension(info={...}, schema={...})` — unchanged (`populate_by_name=True`).
- `model_dump(by_alias=True)` → `{"info": ..., "schema": {...}}` — unchanged (alias).

## Breaking change scope (Fix #1)

- ✅ JSON wire shape: unchanged.
- ✅ Construction (`Extension(schema=...)`): unchanged.
- ⚠️ **Attribute access** on Extension instances: `.schema` now resolves to pydantic's deprecated `BaseModel.schema()` classmethod (returns the model JSON schema). External consumers should switch to `.schema_`. Internal grep finds zero usages outside the one updated reader.

## Test plan

- [x] `python -W error::UserWarning -c "from payments_py.x402.types_v2 import Extension; Extension(info={}, schema={})"` — clean.
- [x] `poetry run pytest tests/unit/` — 446/446 pass.
- [x] `poetry run black --check` — 145 files clean.
- [x] End-to-end tutorial run against real sandbox: all five steps (discovery → token acquisition → agent invocation → settlement receipt) complete cleanly. **No `InsecureRequestWarning` emitted**; **no `UserWarning` from the schema field**.
- [x] Rebased onto current main (after #183 merged).

## Motivation

Surfaced while running the Sprint 0 LangChain tutorial demo. Both warnings fired on every run, making the demo output noisy and (in the case of `InsecureRequestWarning`) calling attention to a real wire-security regression that shipped silently.

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,3 +247,14 @@ poetry install --extras "strands langchain"   # install all optional extras
 poetry run pytest -m "not slow"               # full unit + integration suite
 poetry run pytest tests/unit/x402/            # one directory
 ```
+
+### Custom CA bundle (self-signed-cert environments)
+
+The SDK uses `requests`' default TLS verification against the system trust store. Public Nevermined backends (sandbox, live) serve real certs, so no configuration is needed.
+
+If you're pointing the SDK at a dev environment that serves a self-signed certificate (e.g. a local Caddy with its own CA), export `REQUESTS_CA_BUNDLE` before running — `requests` reads it directly, no SDK change required:
+
+```bash
+export REQUESTS_CA_BUNDLE=/path/to/your-ca.pem
+poetry run python your_script.py
+```
--- a/payments_py/api/base_payments.py
+++ b/payments_py/api/base_payments.py
@@ -182,11 +182,6 @@ class BasePaymentsAPI:
         Returns:
             HTTP options object
         """
-        # Disable SSL verification for development/staging environments
-        # For now, disable SSL verification for all environments to handle
-        # self-signed certificates
-        verify_ssl = False
-
         headers = {
             "Accept": "application/json",
             "Content-Type": "application/json",
@@ -203,9 +198,10 @@ class BasePaymentsAPI:
                 {k: v for k, v in extra_headers.items() if k in ALLOWED_EXTRA_HEADERS}
             )
 
+        # Leave TLS verification on (requests' default). For environments with
+        # self-signed certs, point ``REQUESTS_CA_BUNDLE`` at the CA cert.
         options = {
             "headers": headers,
-            "verify": verify_ssl,
             "timeout": DEFAULT_HTTP_TIMEOUT,
         }
         if body:
@@ -229,14 +225,13 @@ class BasePaymentsAPI:
         Returns:
             HTTP options object
         """
-        verify_ssl = False
-
+        # Leave TLS verification on (requests' default). For environments with
+        # self-signed certs, point ``REQUESTS_CA_BUNDLE`` at the CA cert.
         options = {
             "headers": {
                 "Accept": "application/json",
                 "Content-Type": "application/json",
             },
-            "verify": verify_ssl,
             "timeout": DEFAULT_HTTP_TIMEOUT,
         }
         if body:
--- a/payments_py/api/contracts_api.py
+++ b/payments_py/api/contracts_api.py
@@ -85,7 +85,7 @@ class ContractsAPI(BasePaymentsAPI):
             info_url = f"{backend_url}{API_URL_INFO}"
 
             # Info endpoint doesn't require authentication
-            response = requests.get(info_url, verify=False)
+            response = requests.get(info_url)
             response.raise_for_status()
 
             info_data = response.json()
--- a/payments_py/x402/extensions/nevermined/validate.py
+++ b/payments_py/x402/extensions/nevermined/validate.py
@@ -57,7 +57,7 @@ def validate_nevermined_extension(extension: NeverminedExtension) -> ValidationR
         else:
             # Pydantic Extension model
             info = extension.info
-            schema = extension.schema
+            schema = extension.schema_
 
         # Validate info against schema using JSON Schema
         validate(instance=info, schema=schema)
--- a/payments_py/x402/types_v2.py
+++ b/payments_py/x402/types_v2.py
@@ -14,7 +14,7 @@ https://github.com/coinbase/x402/tree/v2-development
 
 from __future__ import annotations
 
-from pydantic import BaseModel, ConfigDict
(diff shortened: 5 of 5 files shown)


aaitor (approved):
Two fixes for the price of one — and Fix #2 is the consequential one. **Fix #2 quietly resolves a real wire-security regression**: every backend HTTPS call from the SDK was running with `verify=False` (against an explicit comment that read "For now, disable SSL verification for all environments to handle self-signed certificates"). The default `verify=True` posture is what the SDK should have been shipping all along — sandbox and live both serve real, trusted certs, and the previous behaviour silently muzzled `requests`' default cert-chain validation across every API call. Fix #1 (`Extension.schema` → `schema_` + `alias="schema"`) is the more visible change but its impact is mostly the UserWarning cleanup and the technically-breaking `.schema` instance-attribute access that PR description correctly flags. Approving cleanly.

## Verification I ran

- Read the full diff (small — 13 added, 13 removed across 4 files) plus the post-PR `types_v2.py` and `base_payments.py` to confirm the rename + verify removal landed completely.
- **Confirmed `populate_by_name=True`** is in `Extension.model_config` post-PR (L60), so `Extension(schema={...})` keyword construction is unchanged.
- **Confirmed `model_dump(by_alias=True)` wire shape preserved** via `Field(..., alias="schema")` — JSON serialization still emits `{"info": ..., "schema": {...}}`.
- **Cross-checked Extension.schema usage across the whole repo** with `gh search code 'extension.schema' --repo nevermined-io/payments-py` — exactly one match, `payments_py/x402/extensions/nevermined/validate.py:60`, which is the file updated in this PR. No other internal callsite needs to change.
- **Verified the SSL fix removed `verify_ssl = False` in BOTH** `get_backend_http_options` AND `get_public_http_options`, plus the standalone `verify=False` in `contracts_api.get_deployment_info`. Comments added pointing at `REQUESTS_CA_BUNDLE` as the standard escape hatch for self-signed-cert dev environments — that path doesn't require any SDK change, the env var is read directly by `requests`.
- CI: `lint` ✓ (uses the pre-commit hook from #184 — nice to see it working immediately on a follow-up PR), `unit_integration` ✓ (446/446 per PR description), `e2e` pending.
- Parallel agent panel: skipped — this PR is small, focused, and Rodolphe authored both this and #184. The author already did the cross-repo grep and manual e2e tutorial run; the security analysis (Fix #2 is genuinely an improvement, not a regression) checks out on direct read.

## Out-of-scope coordination notes

- **Version bump strategy**: `payments-py` is at `1.6.0` — post-1.0, so strict semver applies. Fix #1's `Extension.schema` → `Extension.schema_` attribute rename is technically a **breaking change** for any external consumer that does instance attribute access (`ext.schema` now silently returns pydantic's deprecated `BaseModel.schema()` classmethod result — a model JSON schema, not the configured field, which is the worst kind of silent regression). Three credible options for the next release:
  - **Major (2.0.0)** — strict semver-correct, plus communicates the wire-security fix as a real change.
  - **Minor (1.7.0)** — pragmatic; the field is internal to the x402 extension type and external consumer hits are likely zero.
  - **Patch (1.6.1)** — if the team treats this as a bug fix only.
  
  My read: the SSL verification change is the higher-impact bump driver, and that's a security fix that's arguably patch-worthy on its own. The `schema_` rename is what nudges this toward at least a minor. Pick deliberately and call it out in the changelog.
- **Documentation surface**: `REQUESTS_CA_BUNDLE` is documented inline in the post-PR comments, but not in the README. Worth a one-liner in the SDK docs under "Development" or "Custom CA" — "hitting a self-signed cert? `export REQUESTS_CA_BUNDLE=/path/to/ca.pem` before running". Not blocking; just makes the escape hatch discoverable for the dev-only edge case.
- **Pre-commit hook from #184 is doing its job**: lint passes on the new code without needing a separate Black step. Confirms the workflow from yesterday's PR is operational.

r-marques:
Thanks @aaitor — applied N2 in this push: README now documents the `REQUESTS_CA_BUNDLE` escape hatch for self-signed-cert environments.

N1 (version bump strategy) and N3 (pre-commit observation) noted, no code changes needed on this PR.

PR #197, opened 2026-05-25
feat(organizations): add create_member / get_members / connect_stripe_account

aaitor:
## Summary

Brings the Python SDK to feature-parity with the [TypeScript SDK](https://github.com/nevermined-io/payments) on the three org-admin methods that were never ported during #179. The TS PR audit (https://github.com/nevermined-io/payments/pull/342#issuecomment-4510466460) flagged them as the remaining delta.

### New `OrganizationsAPI` methods

| Method | Wraps | Audience |
|---|---|---|
| `create_member(user_id, user_email=None, user_role=None)` | `POST /api/v1/organizations/account` | Org admin — Privy enrolment of a new member into the active workspace; returns the new wallet's NVM API key. |
| `get_members(role=None, is_active=None, page=1, offset=100)` | `POST /api/v1/organizations/members` | Org admin — paginated member listing with role / active filters. |
| `connect_stripe_account(user_email, user_country_code, return_url)` | `POST /api/v1/fiat/stripe/account` | Org admin — Stripe Connect onboarding link issuer. |

### New Pydantic types (all re-exported from `payments_py`)

- `OrganizationMember` (row of the members list)
- `OrganizationMembersResponse` (paginated wrapper; normalises the backend's `totalResults` to `total` for parity with the TS shape, with a `total` fallback if a future backend version standardises on it)
- `CreateUserResponse`
- `StripeAccountConnectResult` — named `StripeAccountConnectResult` rather than `StripeCheckoutResult` so the Stripe Connect onboarding type doesn't collide with the plan-purchase checkout-session type used elsewhere in the SDK.

### Notes

- The new types live next to the existing org types in `payments_py/common/types.py` and follow the same Pydantic v2 + `populate_by_name=True` + camelCase-alias convention as `MyMembership`.
- All three methods are admin-gated server-side; the SDK surfaces backend 403s via `PaymentsError.from_backend`. No client-side role checks are added — the SDK doesn't know the caller's role until it hits the wire.

## Test plan

- [x] `poetry run pytest -m "not slow"` — **493 pass** (15 new unit tests in `tests/unit/test_organizations_api.py`)
- [x] `poetry run black --check payments_py tests` clean
- [ ] CI: lint + unit_integration + e2e (E2E unchanged — this PR only adds new admin-gated methods that aren't exercised by the existing fixtures)
- [ ] Smoke test against staging once merged: call all three new methods with the `Testing Merchant` admin identity to confirm the live wire shape matches the Pydantic models.

## Out of scope

- E2E coverage for the three new admin methods. Adding live tests would require a dedicated test member identity to create/destroy on every run; not worth the staging churn for this round.


eruizgar91 (approved):
### Code review

No critical issues found. Checked for bugs, CLAUDE.md compliance, git history, prior-PR review comments, and code-comment guidance. The three new methods (`create_member`, `get_members`, `connect_stripe_account`) parse responses, paginate, and surface backend errors consistently with the existing `OrganizationsAPI` methods, and the 15 new unit tests cover the happy paths, optional-arg omission, the `totalResults`→`total` normalization, and error cases. The `walletResult.hash`→`nvm_api_key` mapping is confirmed correct by the test fixture.

One optional, non-blocking note:

- `OrganizationMember` is a pure wire DTO but does not set `extra="allow"`, unlike `OrganizationActivityEventSubject`. If the backend adds a field to a member row, `model_validate` will still succeed (Pydantic ignores unknowns by default) — so this is purely a forward-compat hardening suggestion, not a bug. `role` being a strict enum matches the existing `MyMembership.role`, so no change needed there.

https://github.com/nevermined-io/payments-py/blob/75cba3d699924f40de67061a3f8b1a42deb46c94/payments_py/common/types.py#L619-L636

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

aaitor:
Pushed \`50fe55e\` resolving the inline review:

**Copilot #3 (validate \`user_id\`)** → \`create_member\` now raises \`PaymentsError.validation("user_id is required")\` client-side when the caller passes an empty string, matching the \`get_organization_activity\` precedent. No backend round-trip.

**Copilot #2 (silent empty-string fallback on malformed response)** → \`create_member\` now requires the \`walletResult\` envelope and the three keys it depends on (\`hash\` / \`userId\` / \`userWallet\`) to be present, and raises \`PaymentsError.from_backend\` with a clear diagnostic when they aren't. A backend contract regression now fails loudly instead of producing a blank-credentialed \`CreateUserResponse\`.

**Copilot #1 (error code preserved + message naming)**:
- Renamed the surfaced error from "Unable to create user" → "Unable to create member" so it matches the public method name.
- Re: \`code\` vs \`errorCode\`: the real backend uses the \`NVMException\` envelope with \`code\` (e.g. \`'BCK.AUTH.0004'\`), and \`PaymentsError.from_backend\` reads it. My original test fixture mocked the wrong field (\`errorCode\`) which led Copilot to infer the backend used \`errorCode\`. Test fixture updated to send \`{code, message, httpStatus}\`, and the test now asserts \`exc_info.value.code == "BCK.AUTH.0004"\` so the structured propagation is regressioned.

**@eruizgar91's note** → \`OrganizationMember\` now sets \`extra="allow"\` to mirror \`OrganizationActivityEventSubject\`. Any new backend-emitted field on a member row is observable on the model without an SDK upgrade.

Local: \`poetry run pytest -m "not slow"\` — **496 pass** (was 493); \`poetry run black --check\` clean.

PR #198, opened 2026-05-26
feat(langsmith): emit verify + settlement spans for LangSmith tracing

r-marques:
## Summary

Implements Sprint 1 of the LangChain partnership epic (nevermined-io/nvm-monorepo#1703). Makes Nevermined verify + settle events first-class citizens of a LangSmith trace.

- New `payments_py.langsmith` module: `verify_span` / `settlement_span` ctx managers + `build_*_metadata` helpers + safety wrappers (`active_run_tree`, `add_metadata`, `redact_metadata_keys`, `abbreviate_token`).
- `@requires_payment` now wraps both verify and settle calls in dedicated child spans, attaches `nvm.*` metadata to both the child and the parent tool span.
- Active `payment_token` redaction: the decorator strips the full x402 access token from the parent tool span's metadata (which LangChain auto-captures from `config["configurable"]`) before opening any `nvm:*` span. Only the abbreviated `nvm.payment_token` (`<first 16>…<last 4>`) reaches Nevermined-emitted attributes.
- New optional extra: `pip install payments-py[langsmith]`. Zero-config opt-in via `LANGSMITH_TRACING=true`.
- Silent no-op when `langsmith` is missing or no LangSmith run is active. Observability failures never propagate into the payment flow.

Implements nevermined-io/nvm-monorepo#1705.

## Trace shape

```
[tool: get_market_insight]
├── [nvm:verify]      attrs: nvm.scheme, nvm.plan_ids, nvm.payer, nvm.agent_id, nvm.agent_request_id, nvm.payment_token (abbrev), nvm.verify.duration_ms
├── <tool body>
└── [nvm:settlement]  attrs: nvm.credits_redeemed, nvm.balance.after, nvm.tx_hash, nvm.network, nvm.payer, nvm.payment_token (abbrev), nvm.settle.duration_ms
```

Full attribute reference table in `docs/api/12-langchain-integration.md` (Observability with LangSmith section) — integration tests now pin every documented attribute (M5 review polish).

## Notes for reviewers

- **Version intentionally unchanged.** Per the `## Release Process` section added to `CLAUDE.md`, the release pipeline (`prepare-release.yml` → `finalize-release.yml` → `release-python.yml`) owns the version bump. Trigger `prepare-release.yml` against this branch after merge with `1.8.0` (new minor — new module + new extra).
[...]

PR #200, opened 2026-05-27
feat(langsmith): PaymentMiddleware for LangSmith Deployment (Sprint 3)

r-marques:
Sprint 3 of the LangChain integration epic ([nvm-monorepo#1703](https://github.com/nevermined-io/nvm-monorepo/issues/1703)), tracked at [nvm-monorepo#1707](https://github.com/nevermined-io/nvm-monorepo/issues/1707).

## Summary

Adds a Starlette ASGI middleware that gates a LangGraph agent deployed to LangSmith Deployment with the Nevermined x402 **verify → agent work → settle** payment lifecycle. Plus a `build_payment_app` factory that returns a FastAPI app pre-wired with the middleware for single-line wiring via `http.app` in `langgraph.json`.

## New surface in `payments_py.langsmith`

| Symbol | Kind | Purpose |
|---|---|---|
| `PaymentMiddleware` | Starlette `BaseHTTPMiddleware` subclass | The actual gating logic |
| `RouteConfig` | dataclass | Per-route pricing — copied from `x402.fastapi.RouteConfig` so the two extras can diverge |
| `build_payment_app(payments, routes=None) -> FastAPI` | factory | Recommended entry point — one-line wiring |
| `X402_HEADERS` | constant | x402 v2 transport header names |

## Behavior

```
Request in
  ├─→ PaymentMiddleware.dispatch:
  │     ├─ read payment-signature header → missing → 402 + envelope in payment-required header
  │     ├─ facilitator.verify_permissions(...) → invalid → 402 + envelope header
  │     ├─ stash PaymentContext on request.state
  │     ├─ await call_next(request)            ← agent runs
  │     ├─ if 2xx → facilitator.settle_permissions(...) → attach settlement header
  │     │            └ settle fails → log + return response unchanged (200)
  │     └─ else: skip settle (no charge for failed runs)
  └─→ agent exceptions propagate naturally to Starlette as 5xx
```

Routes lookup order:
1. Exact `"METHOD /path"` match.
2. Path-parameter match (supports both `:param` and `{param}` syntax — LangGraph routes use `{thread_id}`).
3. Env-var fallback (`NVM_PLAN_ID`, `NVM_CREDITS_PER_INVOKE`, `NVM_AGENT_ID`) for single-plan deployments.
4. Pass-through (no payment gate).

## Why a FastAPI factory

The Day-1 spike for #1707 surfaced an upstream bug in `langgraph-api 0.5.42` (`server.py:107-122`): `update_openapi_spec` falls through to Starlette's `SchemaGenerator` for non-FastAPI `http.app` instances, and that generator crashes when YAML-parsing internal endpoint docstrings (e.g. `api/mcp.py:104` has `"to support the following methods:"`). FastAPI takes a clean path through `app.openapi()`. The factory shields users from this bug. Detailed write-up: `~/code/notes/sprint-3-langsmith-deployment-scoping.md`.

## Tests

- 16 new unit tests in `tests/unit/langsmith/test_middleware.py`
- 53 of 53 langsmith tests pass (37 Sprint 1 spans + 16 Sprint 3 middleware)
- 521 of 521 full unit suite pass

Coverage:
- 402 emission shapes (missing header, invalid verify)
- FastAPI `{param}` and Starlette `:param` path matching
- Agent-failure-skips-settle
- Settle-failure-still-returns-200
- Env-var fallback (gates all paths when set, off when unset, defaults to credits=1 on non-numeric input)
- `build_payment_app` factory smoke test

## Dispatch fix vs FastAPI middleware

The existing `payments_py.x402.fastapi.PaymentMiddleware` wraps `await call_next(request)` in the same try/except that catches verify errors, which silently converts agent exceptions into 402 responses. The new langsmith middleware separates the two try blocks so agent failures propagate to Starlette as 5xx as intended (and settle is correctly skipped on non-2xx). Test `test_non_2xx_response_skips_settle` covers this. Worth backporting to the FastAPI middleware in a follow-up PR.

## Deps

`[langsmith]` extra now includes `fastapi` and `starlette` (the optional deps were already declared in `pyproject.toml` for the `[fastapi]` extra — we just added them to the `langsmith` extras list).

## Out of scope

- TS variant — TS-3 ([nvm-monorepo#1711](https://github.com/nevermined-io/nvm-monorepo/issues/1711), blocked on LangChain shipping a TS runtime).
- Streaming response support — documented limitation in the module docstring. Settle attaches a header which requires buffering the body. Streaming endpoints lose streaming when wrapped. Follow-up.
- Tutorial — separate PR in the tutorials repo (next on the work list).
- Docs page — separate PR in the docs repo (after tutorial).
- Release — release pipeline triggered manually post-merge.

## Test plan

- [ ] CI green
- [ ] Tutorial in `~/code/tutorials/langchain-langsmith-deployment-py/` deploys cleanly with the new middleware once it's plumbed through (in-flight)
- [ ] 402 round-trip verified with `curl` + buyer script against `langgraph dev` (in-flight)


payments_py/langsmith/middleware.py
```diff
@@ -0,0 +1,485 @@
...
+        }
+
+    def _resolve_route_config(self, method: str, path: str) -> Optional[RouteConfig]:
+        return _match_route(method, path, self.routes)
+
+    async def dispatch(
+        self, request: Request, call_next: Callable[[Request], Awaitable[Response]]
+    ) -> Response:
+        method = request.method.upper()
+        path = request.url.path
```
aaitor: **MEDIUM — Normalise the path before `_match_route` to close a defense-in-depth gap.**

`request.url.path` comes from `scope["path"]` which Starlette/uvicorn (httptools) does NOT dot-normalise by default. A request like `POST /threads/abc/../abc/runs/wait` produces 7 segments vs the configured `POST /threads/{thread_id}/runs/wait`'s 5, so `_match_route` returns `None` and the request passes through ungated. Without a normalising proxy in front of langgraph this is harmless today — the downstream router also won't match the same path — but the moment ANY layer between uvicorn and the handler starts normalising (a future Starlette feature, an upstream nginx, a Cloud Run / Lambda gateway), the middleware silently lets traversal-encoded paths through to a protected handler.

This is defense-in-depth, not an active bypass. One line:

```suggestion
        method = request.method.upper()
        path = posixpath.normpath(request.url.path)
        # preserve trailing slash so /runs/ and /runs/wait/ stay distinct from /runs and /runs/wait
        if request.url.path.endswith("/") and not path.endswith("/"):
            path += "/"
```

(Add `import posixpath` at the top of the file.)

Use the normalised `path` for both `_match_route` and `build_payment_required(endpoint=str(request.url.path), ...)` at L261 — the envelope's `endpoint` field should also reflect the canonical resource.

payments_py/langsmith/middleware.py
```diff
@@ -0,0 +1,485 @@
...
+                ).decode()
+
+                body = b""
+                async for chunk in response.body_iterator:
+                    body += chunk
+
+                new_response = Response(
+                    content=body,
+                    status_code=response.status_code,
+                    headers=dict(response.headers),
```
aaitor: **MEDIUM — `dict(response.headers)` collapses multi-value headers to first-value-wins; lose any `Set-Cookie` beyond the first.**

Starlette's `Headers.__iter__` returns keys-with-duplicates, but `Headers.__getitem__` returns only the first matching value. So `dict(response.headers)` is `{"set-cookie": "<first cookie>"}` even when two `Set-Cookie` headers are set. LangGraph's own session/CSRF middleware can legitimately emit two cookies — this PR's middleware would silently drop the second.

Fix: copy headers via `response.headers.raw` (the underlying `(bytes, bytes)` list) so multi-value entries are preserved, then add the receipt header:

```suggestion
                new_response = Response(
                    content=body,
                    status_code=response.status_code,
                    media_type=response.media_type,
                )
                # raw_headers preserves multi-value headers (Set-Cookie, etc.)
                # that dict(response.headers) would silently collapse.
                new_response.raw_headers = list(response.headers.raw)
                new_response.headers.append(
                    X402_HEADERS["PAYMENT_RESPONSE"], receipt_b64
                )
                return new_response
```

This also fixes the latent `content-length` issue (the `Response()` constructor recomputes `content-length` from `body` when `headers=None`, so the receipt header insertion can't cause a length mismatch).

Add a small unit test that confirms multi-`Set-Cookie` survives the re-wrap on the settle-success path.

payments_py/langsmith/middleware.py
```diff
@@ -0,0 +1,485 @@
...
+                    media_type=response.media_type,
+                )
+                new_response.headers[X402_HEADERS["PAYMENT_RESPONSE"]] = receipt_b64
```
aaitor: **MEDIUM — Settlement receipt + 402 envelope should be `Cache-Control: no-store`.**

The `payment-response` header carries `remaining_balance` (buyer's credit balance after this deduction) and `payer` (wallet address). The `payment-required` header on the 402 path carries the envelope tied to the current request's `plan_id` + `endpoint` + `agent_id`. If an integrator places a CDN (Cloudflare, AWS CloudFront, nginx) in front of the LangSmith Deployment endpoint and the downstream agent returns a cacheable response (e.g. `Cache-Control: max-age=60`), a different buyer who hits the same CDN cache later gets the prior buyer's settlement receipt — leaking their wallet address and balance. Same concern (smaller blast) for cached 402 envelopes.

Force no-store on both response paths:

```ts
// In the settle-success path after appending the receipt header:
new_response.headers["Cache-Control"] = "no-store"
new_response.headers["Pragma"] = "no-cache"

// In _send_payment_required (L184-194):
return JSONResponse(
    status_code=402,
    content={"error": "Payment Required", "message": message},
    headers={
        X402_HEADERS["PAYMENT_REQUIRED"]: envelope_b64,
        "Cache-Control": "no-store",
        "Pragma": "no-cache",
    },
)
```

Not a payment bypass — financial-metadata leak between subscribers via shared cache. Worth closing now before any deployer wires a CDN in front.

payments_py/langsmith/middleware.py
```diff
@@ -0,0 +1,485 @@
...
+                    verified=True,
+                    agent_request_id=verification.agent_request_id,
+                    agent_request=verification.agent_request,
+                )
+
+                # Agent runs here. Exceptions propagate naturally - the buyer
+                # is not charged for failed runs (we skip settle on non-2xx
+                # below). The parent trace will also see the exception via
+                # __exit__ and be marked failed in LangSmith.
+                response = await call_next(request)
```
aaitor, line 401: **MEDIUM — Strip `payment-signature` from `request.scope` before `call_next` so the bearer token doesn't ship to Sentry / structlog / OTEL header captures.**

Sentry's `RequestIntegration` (enabled by default when `sentry_sdk.init()` runs) captures every request header and ships them to sentry.io unless the name is in the default denylist. The denylist covers `Authorization`, `Cookie`, `X-Auth-Token` — but NOT `payment-signature`. Structlog ASGI processors and OpenTelemetry's HTTP semantic conventions similarly capture `http.request.header.*`. A deployer who adds Sentry without a custom scrubber exfiltrates every x402 access token to their Sentry org.

The middleware already extracted the token at L272 and stashed it on `request.state.payment_context.token` at L365 — downstream code doesn't need to read the header anymore. Strip it before `call_next`:

```python
        # After L372 (request.state.payment_context = PaymentContext(...)):
        from starlette.datastructures import MutableHeaders
        MutableHeaders(scope=request.scope).pop(
            X402_HEADERS["PAYMENT_SIGNATURE"], None
        )

        # Agent runs here. Exceptions propagate naturally - the buyer
        # is not charged for failed runs (we skip settle on non-2xx
        # below). The parent trace will also see the exception via
        # __exit__ and be marked failed in LangSmith.
        response = await call_next(request)
```

Same PR #198 finding for the LangChain decorator path — applies identically here at the framework boundary.

payments_py/langsmith/middleware.py
```diff
@@ -0,0 +1,485 @@
...
+                    )
+                    return response
+
+                receipt_b64 = base64.b64encode(
+                    settlement.model_dump_json(by_alias=True).encode()
+                ).decode()
+
+                body = b""
+                async for chunk in response.body_iterator:
+                    body += chunk
```
aaitor, line 467: **MEDIUM — Add a runtime warning when buffering an obviously-streaming response.**

The module docstring documents that streaming endpoints lose streaming when wrapped, but a deployer who points the middleware at `/runs/stream` (LangGraph's SSE endpoint) gets the entire response buffered into memory with no observable signal — the client just sees a delivered-at-the-end response and the worker holds the full body in memory under load. For LangGraph Deployment specifically, SSE is the headline feature most users adopt langgraph for.

Add a one-line check before the buffering loop:

```suggestion
                receipt_b64 = base64.b64encode(
                    settlement.model_dump_json(by_alias=True).encode()
                ).decode()

                content_type = response.headers.get("content-type", "")
                is_streaming = (
                    "text/event-stream" in content_type
                    or response.headers.get("transfer-encoding") == "chunked"
                )
                if is_streaming:
                    logger.warning(
                        "PaymentMiddleware is buffering a streaming response for path=%s "
                        "(content-type=%s). Streaming is disabled by this middleware — "
                        "gate /runs/wait instead, or pass the route through ungated.",
                        path,
                        content_type or "<unset>",
                    )

                body = b""
                async for chunk in response.body_iterator:
                    body += chunk
```

Leaves behaviour unchanged (still buffers) but makes the misconfiguration visible in logs. A stricter long-term option: a `RouteConfig.streaming: bool = False` flag that, when True, settles without re-wrapping the response and delivers the receipt out-of-band (trailer header or callback) — file as a follow-up.

tests/unit/langsmith/test_middleware.py
```diff
@@ -0,0 +1,350 @@
...
+            return JSONResponse({"ok": True})
+
+        client = TestClient(app)
+        response = client.post("/runs")
+
+        assert response.status_code == 402
+        assert X402_HEADERS["PAYMENT_REQUIRED"] in response.headers
+
+    def test_factory_accepts_none_routes(self, mock_payments, monkeypatch):
+        """routes=None is valid; behavior matches routes={} (env-var fallback only)."""
```
aaitor, line 455: **MEDIUM — Two coverage gaps + one tiny extension worth landing in this PR.**

The new middleware buffers the response body and re-wraps it on the settle-success path (`middleware.py:428-439`). None of these tests pin the buffering contract:

**(a) Multi-chunk body.** All fixtures return a single-chunk JSONResponse. A future refactor that breaks `body += chunk` (or replaces it with a last-chunk-only assignment) would not fail any test.

```python
def test_multi_chunk_body_is_buffered_correctly(self, valid_token, mock_payments):
    """body_iterator with multiple chunks must be concatenated in full."""
    from starlette.responses import StreamingResponse

    async def multi_chunk():
        yield b'{"response":'
        yield b'"chunked answer"}'

    app = FastAPI()
    app.add_middleware(
        PaymentMiddleware,
        payments=mock_payments,
        routes={"GET /stream": {"plan_id": "test-plan-123", "credits": 1}},
    )

    @app.get("/stream")
    async def stream_route():
        return StreamingResponse(multi_chunk(), media_type="application/json")

    client = TestClient(app)
    response = client.get(
        "/stream",
        headers={X402_HEADERS["PAYMENT_SIGNATURE"]: valid_token},
    )
    assert response.status_code == 200
    assert response.json() == {"response": "chunked answer"}
    assert X402_HEADERS["PAYMENT_RESPONSE"] in response.headers
```

**(b) `PaymentContext` reachable from the handler.** `middleware.py:365-372` stashes a `PaymentContext` on `request.state`. No test confirms the agent function inside `call_next` can read it — that's the primary API surface for downstream LangGraph code that needs `agent_request_id` for observability or `credits_to_settle` for dynamic pricing.

```python
def test_payment_context_available_in_handler(self, valid_token, mock_payments):
    captured = {}
    app = FastAPI()
    app.add_middleware(
        PaymentMiddleware,
        payments=mock_payments,
        routes={"GET /ctx": {"plan_id": "test-plan-123", "credits": 3}},
    )

    @app.get("/ctx")
    async def ctx_route(request: Request):
        ctx = request.state.payment_context
        captured["verified"] = ctx.verified
        captured["credits"] = ctx.credits_to_settle
        captured["agent_request_id"] = ctx.agent_request_id
        return JSONResponse({"ok": True})

    TestClient(app).get(
        "/ctx", headers={X402_HEADERS["PAYMENT_SIGNATURE"]: valid_token},
    )
    assert captured["verified"] is True
    assert captured["credits"] == 3
    assert captured["agent_request_id"] == "test-request-id-123"
```

**(c) One-line tightening to the existing `test_settle_failure_still_returns_200`.** The test asserts status 200 and absence of the receipt header but doesn't assert the agent's response body survives. If settle-failure returns the wrong `Response` object (e.g. a partial `body_iterator` after a `call_next` consumption regression in Starlette), the test passes but production breaks. Add:

```python
# Inside test_settle_failure_still_returns_200, after the existing assertions:
assert response.json() == {"response": "Answer to: hi"}
```

aaitor (approved):
Solid Sprint 3 landing. The middleware's verify/work/settle lifecycle is correctly shaped (the fix vs the FastAPI middleware's broken try/except is real — agent exceptions propagate as 5xx as intended, and settle is correctly skipped on non-2xx). The Sprint 1 span integration is clean — verify span opens BEFORE the token check, settle span is opened around the settle call, both nest under a parent trace. `attach_metadata_safely` was promoted to public correctly so the middleware and the decorator share the helper. Approving with 6 MEDIUM polish items inline and a few follow-up notes in the body — none are merge blockers.

**Verification I ran**
- `gh pr diff 200 --repo nevermined-io/payments-py` against `main` (base) + `feat/langsmith-deployment-middleware` (head, `b1a78ea`).
- Read full post-PR contents of `langsmith/middleware.py` (485 lines), `langsmith/spans.py`, `x402/types.py`, `x402/langchain/decorator.py`, and `tests/unit/langsmith/test_middleware.py`.
- Verified backward-compat for `PaymentRequiredError`: `x402/langchain/decorator.py:83` re-imports it from `x402/types.py`, and `x402/langchain/__init__.py:34-39` re-exports it. `from payments_py.x402.langchain import PaymentRequiredError` still works.
- CI: e2e ✅, lint ✅, unit_integration ✅ on head.
- Parallel reviewer panel: `python-reviewer`, `security-reviewer`, `tdd-guide`, `architect`. Synthesis below.

**Multi-perspective synthesis**

Promoted (see inline comments):
- (security-reviewer) Path normalisation before `_match_route` — defense-in-depth against future downstream router normalisation creating a route-bypass class.
- (code-reviewer + security-reviewer) `dict(response.headers)` collapses multi-value headers — concrete fix using `response.headers.raw`.
- (security-reviewer) Settlement receipt + 402 envelope need `Cache-Control: no-store` to prevent CDN/proxy from serving stale receipts to other buyers.
- (security-reviewer) `payment-signature` bearer header should be stripped from `request.scope` before `call_next` so Sentry's `RequestIntegration` and structlog ASGI processors don't ship the token to a second SaaS destination.
- (code-reviewer + architect) Streaming endpoint silent buffering — add a runtime `logger.warning` when `text/event-stream` or `Transfer-Encoding: chunked` is detected; LangGraph's marquee `/runs/stream` endpoints are SSE.
- (tdd-guide) Multi-chunk body buffering, `content-length` post-rewrap, and `request.state.payment_context` handler-side reads are all unexercised — sketched tests inline.

Checked and not promoted:
- Security-reviewer flagged path traversal as **HIGH**. Verified: Starlette/uvicorn with httptools does NOT dot-normalise `scope["path"]` by default, so a path like `/threads/abc/../abc/runs/wait` produces 7 segments vs. the configured 5 and the route lookup returns `None` (pass-through). Without a normalising proxy in front of langgraph, the downstream router would also 404 on this path. The risk only materialises if a future change in the routing layer adds normalisation — that's defense-in-depth, not an active exploit. Demoted to MEDIUM and inline.
- Architect flagged "sibling top-level traces" as **HIGH**. Real architectural concern (the PR's own comment at `middleware.py:155-160` acknowledges it), but it's a known sprint-scope limitation — the langgraph runtime initiates its own top-level trace at the graph-invocation boundary and there's no parent context to attach to from middleware-dispatch time. Worth a follow-up to attach the payment metadata to langgraph's run via `get_current_run_tree()` AFTER `call_next` returns. Surfaced as a body follow-up.
- Architect flagged "consolidate two middleware classes" as **HIGH**. The PR description explicitly defers this to a follow-up. Stronger position from the agent: don't ship the bug fix in `langsmith/` while leaving the same bug in `x402/fastapi/`. I agree the follow-up shouldn't wait — recommend filing the FastAPI dispatch fix as a same-day separate PR rather than "future work." Surfaced as a body follow-up.
- Code-reviewer's `_resolve_credits` exception path → 402 not 500 concern. Real but narrow (only triggers for buggy callables); low impact. Body note.
- Code-reviewer's `_match_route` ambiguous-parameterised-routes first-match-wins. Real but documented behaviour; LangGraph route tables don't hit this.
- Code-reviewer's `content-length` post-rewrap. Reduced to LOW by the agent itself. Body note (suggest stripping `content-length` before `Response()` so it's recomputed from `body`).
- Tdd-guide's HIGH on `PaymentRequiredError` class-identity test. Verified the re-export chain works; missing test is good defensive programming but not a regression. LOW.

**Out-of-scope follow-ups worth filing**

1. **FastAPI middleware dispatch bug — file as immediate, not deferred.** The PR description says "worth backporting in a follow-up PR." That follow-up should be a same-week separate PR (not Sprint 4+), because shipping the LangSmith correct version while leaving the FastAPI middleware silently converting agent 5xx into 402 is a known correctness bug in production. The minimal fix is to split the `try/except` so `call_next(request)` is outside the verify-error catch.

2. **Sibling-trace UX in LangSmith.** The middleware emits a top-level `nvm:x402-request` trace as a sibling to langgraph's own graph trace. Two unrelated trace entries per request is a worse observability story than one nested trace. Follow-up: collect verify/settle metadata into `request.state.x402_trace_data` during dispatch, then attach as a child of langgraph's run via `get_current_run_tree()` AFTER `call_next` returns. Alternatively, emit OpenTelemetry spans (LangSmith ingests them and nests correctly).

3. **`build_payment_app` is load-bearing for a `langgraph-api==0.5.42` OpenAPI bug.** If langgraph fixes the bug in 0.5.43+, the factory becomes dead weight that users still adopt because it's the documented entry point. Document the rationale in the docstring and add a runtime `DeprecationWarning` when a newer langgraph version (where the bug is fixed) is detected, telling users to switch to direct `app.add_middleware`.

4. **Public surface commitments.** This PR adds 5 public symbols (`PaymentMiddleware`, `RouteConfig`, `X402_HEADERS`, `build_payment_app`, `attach_metadata_safely`). Once published in the next `payments-py` release, these are semver commitments. Mark `build_payment_app` as `experimental` in the docstring; consider whether `attach_metadata_safely` should stay underscore-prefixed (private) given its narrow contract.

5. **`RouteConfig` duplicated across `x402.fastapi` and `langsmith`** with the docstring rationale "let extras diverge." Both currently have identical fields. When the first framework-specific field lands, it'll need updating in both. Consider moving to `x402.types` as canonical and re-exporting from both.

6. **Token in frame locals.** Same concern as Sprint 1 PR #198 raised — the local `token` lives in scope across the entire `dispatch` body, captured by Sentry / structlog frame-locals enrichers on any in-flight exception. Set `token = None` after stashing into `PaymentContext` to scrub the local before any exception path could capture it.

**Approval rationale**

No HIGH issues. The MEDIUM items are concrete polish on a real production-ready middleware; CI is green, span integration is correct, the dispatch fix vs the FastAPI middleware is a real win. Approving — fold the inline suggestions when convenient and file follow-ups (1) and (2) before next sprint.

PR #202, opened 2026-05-27
docs: fix Sprint 3 chapter cross-links + register chapter 13 in mintlify sync

r-marques:
## Summary

Two doc issues visible on the live Mintlify site after the v1.9.0 release:

1. **\"Chapter N\" wording in cross-links** — readers only see page titles on Mintlify, no chapter numbers. Rewrite to use page-title references.
2. **Chapter 13 was silently skipped by the mintlify sync** — \`scripts/convert_to_mintlify.py\`'s FILE_MAPPING/LINK_MAPPING stopped at chapter 12. The v1.9.0 sync workflow ran successfully but emitted no MDX for chapter 13, and left the raw \`./13-langsmith-deployment.md\` link (with \`.md\` extension) in the converted chapter 12 page.

## Changes

- \`docs/api/12-langchain-integration.md\` — cross-link references LangSmith Deployment by page title, not chapter number.
- \`docs/api/13-langsmith-deployment.md\` — all four \"chapter N\" references rewritten to page-title references. Hash-anchor link dropped (the converter regex doesn't preserve fragments) in favor of prose pointing at the section by name.
- \`scripts/convert_to_mintlify.py\` — chapter 13 added to FILE_MAPPING + LINK_MAPPING.

## Verification

Locally ran the converter and confirmed:
- \`langchain-module.mdx\` links to \`/docs/api-reference/python/langsmith-deployment-module\` (no \`.md\`)
- \`langsmith-deployment-module.mdx\` generated, all internal links resolved
- No remaining \"chapter\" wording in either converted MDX

## Follow-up

A companion docs_mintlify PR will backfill the missing MDX + nav entry for the v1.9.0 release that's already live. Future v* tags will sync chapter 13 automatically.

### DIFF
--- a/scripts/convert_to_mintlify.py
+++ b/scripts/convert_to_mintlify.py
@@ -12,6 +12,17 @@ import argparse
 from pathlib import Path
 from typing import Optional
 
+# Adding a new chapter requires FOUR edits, not one:
+#   1. Create the source file under `docs/api/<NN>-<slug>.md`.
+#   2. Add an entry to FILE_MAPPING below with `target`, `icon`, and `description`.
+#   3. Add an entry to LINK_MAPPING further down so cross-references convert
+#      to clean Mintlify URLs (the chapter 12 -> chapter 13 link will
+#      otherwise ship to Mintlify with a `.md` suffix visible to readers).
+#   4. Add an entry to `mkdocs.yml`'s nav array so the chapter is reachable
+#      on the published mkdocs site.
+# Missing any of (2)-(4) silently no-ops on the next mintlify sync — the
+# chapter exists upstream but never reaches docs_mintlify. PR #202 was the
+# v1.9.0 retroactive fix for chapter 13 missing exactly this set of edits.
 # Mapping of source files to target names and metadata
 FILE_MAPPING = {
     "01-installation.md": {
--- a/scripts/convert_to_mintlify.py
+++ b/scripts/convert_to_mintlify.py
@@ -74,6 +74,11 @@ FILE_MAPPING = {
         "icon": "link-simple",
         "description": "Decorator, helpers, and exceptions for protecting LangChain and LangGraph tools",
     },
+    "13-langsmith-deployment.md": {
+        "target": "langsmith-deployment-module.mdx",
+        "icon": "graduation-cap",
+        "description": "Starlette middleware for gating a LangGraph agent deployed to LangSmith Deployment with x402",
+    },
 }
 
 # Link mapping for internal references
@@ -90,6 +95,7 @@ LINK_MAPPING = {
     "10-a2a-integration.md": "/docs/api-reference/python/a2a-module",
     "11-x402.md": "/docs/api-reference/python/x402-module",
     "12-langchain-integration.md": "/docs/api-reference/python/langchain-module",
+    "13-langsmith-deployment.md": "/docs/api-reference/python/langsmith-deployment-module",
 }
 
 
--- a/docs/api/12-langchain-integration.md
+++ b/docs/api/12-langchain-integration.md
@@ -4,7 +4,7 @@ This guide covers the LangChain and LangGraph integration shipped in the `[langc
 
 For the conceptual walk-through (two integration approaches, the discovery-first flow, dynamic credits patterns), see the [LangChain integration guide](/docs/integrate/add-to-your-agent/langchain). For a runnable end-to-end demo, see the [`langchain-paid-agent-py`](https://github.com/nevermined-io/tutorials/tree/main/langchain-paid-agent-py) tutorial.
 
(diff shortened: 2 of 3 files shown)


scripts/convert_to_mintlify.py
```diff
@@ -74,6 +74,11 @@
         "icon": "link-simple",
         "description": "Decorator, helpers, and exceptions for protecting LangChain and LangGraph tools",
     },
+    "13-langsmith-deployment.md": {
+        "target": "langsmith-deployment-module.mdx",
```
aaitor, line 89: **LOW — Adding a new chapter requires four edits, not one — worth a comment near the mappings.**

This PR is the second time chapter 13's mintlify wiring was missed (the v1.9.0 sync emitted no MDX for it). The chapter file lives in `docs/api/`, but the dual `FILE_MAPPING` + `LINK_MAPPING` here and the nav entry in `mkdocs.yml` are independent edits that all need to happen together. Same class of footgun as the migration / env-var dual-edits called out in CLAUDE.md.

No code change needed in this PR; one comment block at the top of `FILE_MAPPING` documenting the touchpoints would prevent the next miss:

```python
# Adding a new chapter requires THREE edits, not one:
#   1. Create the source file under `docs/api/<NN>-<slug>.md`.
#   2. Add an entry here with `target`, `icon`, and `description`.
#   3. Add an entry to `LINK_MAPPING` below so cross-references convert.
#   4. Add an entry to `mkdocs.yml`'s nav array.
# Missing any of (2)-(4) silently no-ops on the next mintlify sync (the
# chapter exists upstream but never reaches docs_mintlify). PR #202 was
# the v1.9.0 retroactive fix for chapter 13 missing all three.
```

Longer-term follow-up (per the body note): generate the mappings from a single source of truth (e.g. a per-chapter YAML frontmatter block, or auto-discover from `docs/api/*.md`) so the dual-edit class goes away entirely.

aaitor (approved):
Tight, targeted post-release fix for the two issues visible on the live Mintlify site after v1.9.0. Cross-link wording is correct (page-title references, not chapter numbers — Mintlify users don't see chapter numbers) and the converter now registers chapter 13. Approving with one LOW inline + two body follow-ups to prevent the same class of drift next time.

**Verification I ran**
- `gh pr diff 202 --repo nevermined-io/payments-py` against `main` (base) + `fix/langsmith-docs-cross-links` (head, `982b6cb`).
- Grepped all 13 chapter files (`01-installation.md` through `13-langsmith-deployment.md`) for residual `[Cc]hapter [0-9]+` wording — **all clean post-PR**. The cross-link rewrites in ch12 and ch13 are the only ones that needed changing.
- Confirmed `mkdocs.yml:94` already has `- 13. LangSmith Deployment: api/13-langsmith-deployment.md` (added in Sprint 3 PR #200). The `convert_to_mintlify.py` mapping additions are the only missing piece.
- CI snapshot: lint ✅, unit_integration ✅, e2e pending — none of those run the mintlify converter (it fires on `v*` tag push via `publish-mintlify-docs.yml`), so the converter behaviour is verified by the author's local run noted in the PR description.
- Confirmed `./11-x402.md` is in the LINK_MAPPING (the bare-path form like `[x402 Protocol](11-x402.md)` at ch12:L363 matches the converter's `[^)]*{source}` regex, so the existing reference will resolve correctly on the next sync).

Small PR, no parallel reviewer panel — overkill for a 68-line docs/script fix.

**Out-of-scope follow-ups worth filing**

1. **Converter regex silently drops fragments.** The pattern at `scripts/convert_to_mintlify.py:178` is `rf"\[([^\]]+)\]\([^)]*{re.escape(source)}\)"` — it anchors `source` immediately before `)`, so `[text](./ch.md#section)` doesn't match the pattern and falls through unconverted. This PR works around the limitation by dropping the one `#observability` fragment that existed (ch13:L157), but the regex stays broken. A small fix that preserves fragments:
   ```python
   pattern = rf"\[([^\]]+)\]\(([^)]*{re.escape(source)})(#[^)]*)?\)"
   replacement = lambda m: f"[{m.group(1)}]({target}{m.group(3) or ''})"
   content = re.sub(pattern, replacement, content)
   ```
   Lets future docs use anchor links like `[Observability](./12-langchain-integration.md#observability)` without losing them on sync. File as a follow-up; not gating this PR.

2. **CI step on PRs touching `docs/api/`.** This whole class of drift (chapter file added, FILE_MAPPING / LINK_MAPPING forgotten, mkdocs.yml forgotten, cross-link uses obsolete wording) would surface on PR if CI ran `scripts/convert_to_mintlify.py --check` or equivalent against the modified chapters. The current model — "the sync runs on tag, breakage is discovered after release" — is what produced this PR. A workflow that runs the converter on `docs/api/**` changes and fails on unresolved links closes the loop. File as a follow-up.

**Approval rationale**

Docs-only + a one-block addition to the converter script. No code paths affected, no breaking changes, CI green on what runs. The author already ran the converter locally and verified both MDX outputs. Approving.

PR #204, opened 2026-05-29
ci: trigger workflows on pull_request to cover fork PRs

r-marques:
## Summary

Both `Tests` and `Lint and Build` are currently `on: [push]`, which GitHub does **not** dispatch on the base repo when a fork PR is opened or pushed. The result: workflows have never run on any fork PR. #181 is the first external contribution and surfaced the gap (it's been sitting with zero status checks since 2026-05-21).

## Change

```yaml
# both .github/workflows/test.yaml and .github/workflows/lint.yml
on:
  push:
    branches: [main]
  pull_request:
```

- **`push: branches: [main]`** preserves the post-merge "main is green" signal. We lose CI on intermediate commits to internal long-lived branches without an open PR — defensible since code without review doesn't really need CI either, but flagging in case it bites someone's pre-PR habits.
- **`pull_request`** fires on every PR open / synchronize / reopen, including fork PRs.

## Why this is safe

GitHub's design **does NOT pass repository secrets** to fork-PR workflow runs ([docs](https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#using-encrypted-secrets-in-a-workflow)). So `TEST_SUBSCRIBER_API_KEY` / `TEST_BUILDER_API_KEY` stay out of contributor code under `pull_request` from a fork. The pytest suite already gracefully skips the keyed paths when env vars are absent — verified locally on #181: `562/562 not-slow tests pass` with no secrets set.

This is in contrast to `pull_request_target` (the dangerous variant), which we are **not** using.

## Spam-defense (one-time UI flip, recommended)

After this merges, please flip:

**Settings → Actions → General → Fork pull request workflows from outside collaborators → "Require approval for all outside collaborators"**

That puts every fork PR's workflow into `action_required` state until a maintainer clicks Approve. Zero unapproved spam runs. ~30s flip.

Cost: actions on `ubuntu-latest` in public repos are free unlimited, so the cost concern is bounded to "annoying queue contention" even without the approval gate — but the gate is the proper fix and worth enabling.

## After merge

Once this lands, the next push (or close+reopen) on #181 fires a `pull_request: synchronize` event → workflow enters `action_required` → one click → CI runs.

## Test plan

- [ ] Merge this
- [ ] Approve the queued workflow run on #181
- [ ] Confirm lint + tests run green on #181 (peterxing's branch)
- [ ] Flip the org Actions "Require approval" setting

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,6 +1,9 @@
 name: Lint and Build
 
-on: [push]
+on:
+  push:
+    branches: [main]
+  pull_request:
 
 jobs:
   lint:
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -3,7 +3,10 @@
 
 name: Tests
 
-on: [push]
+on:
+  push:
+    branches: [main]
+  pull_request:
 
 jobs:
   unit_integration:


eruizgar91 (approved):
Approving after a security-focused review of the trigger change.

Verdict: safe. No exploitable HIGH/MEDIUM findings.

- Uses `pull_request` (not the dangerous `pull_request_target`), so fork PR code runs in the fork's context without elevated privileges.
- GitHub does not pass repository secrets (`TEST_SUBSCRIBER_API_KEY` / `TEST_BUILDER_API_KEY`) to fork-PR runs, so they resolve empty for forks — no secret exfiltration path.
- No untrusted input (`github.event.*`) is interpolated into any `run:` step, so no script injection.
- `GITHUB_TOKEN` is read-only for fork PRs; runners are GitHub-hosted/ephemeral (no self-hosted RCE surface); no `workflow_run` chaining to a privileged workflow (`finalize-release.yml` stays gated on `merged == true` + `release/*`).

Residual risk is bounded to compute abuse, which is closed by the recommended approval gate.

One operational ask before this is fully safe: enable Settings → Actions → General → 'Require approval for all outside collaborators' first, then merge — so there is no exposure window.

.github/workflows/lint.yml
```diff
@@ -1,6 +1,9 @@
 name: Lint and Build
 
-on: [push]
+on:
+  push:
+    branches: [main]
+  pull_request:
```
aaitor, line 6: **MEDIUM — Add `concurrency:` and `permissions:` blocks while the trigger is being touched**

Two defense-in-depth additions worth folding into this PR:

1. **`concurrency:`** — rapid fork-PR force-pushes will queue redundant runs without `cancel-in-progress`. Especially relevant now that external contributors can trigger CI.
2. **`permissions:`** — explicit `contents: read` declaration so the workflow doesn't silently inherit write access if the repo is ever made private or the org default flips to read-write. Costs nothing today, prevents a silent regression tomorrow.

```suggestion
  pull_request:

concurrency:
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
  cancel-in-progress: true

permissions:
  contents: read
```

.github/workflows/test.yaml
```diff
@@ -3,7 +3,10 @@
 
 name: Tests
 
-on: [push]
+on:
+  push:
+    branches: [main]
+  pull_request:
```
aaitor, line 9: **MEDIUM — Add `concurrency:` and `permissions:` blocks (mirror of the `lint.yml` suggestion)**

Same hardening as in `lint.yml`. The `concurrency` group keyed on PR number means each new push on a fork PR cancels the prior run, which matters more here because the `e2e` job is the long pole.

```suggestion
  pull_request:

concurrency:
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
  cancel-in-progress: true

permissions:
  contents: read
```

.github/workflows/test.yaml
```diff
@@ -3,7 +3,10 @@
...
 
 jobs:
   unit_integration:
```
aaitor, line 12: **MEDIUM — Replace `poetry lock` with `poetry check --lock` (not visible in this diff: lines 24 and 47)**

Highest-leverage finding in this PR. With the new `pull_request` trigger, fork PRs will run `poetry lock --no-interaction --no-ansi` (at line 24 in `unit_integration` and line 47 in `e2e`) on fork-controlled `pyproject.toml`. A hostile fork can swap one specifier (e.g. add `evil-package = "*"` or typosquat an existing dep) and the runner will resolve, install, and execute the package's `setup.py` / `__init__.py` — arbitrary code execution on the runner without any maintainer approval, because `poetry install` follows the regenerated lock blindly.

The approval gate from the PR body is the right primary control, but defense-in-depth here costs one line per occurrence — match what `lint.yml` already does and validate the committed lock instead of regenerating it. A fork PR that diverges `pyproject.toml` from `poetry.lock` will then fail at the lockfile-check step, before any `poetry install` runs.

For both the `unit_integration` job (~line 24) and the `e2e` job (~line 47):

```yaml
# Replace:
- name: Resolve lockfile
  run: poetry lock --no-interaction --no-ansi

# With (matches lint.yml line 19):
- name: Verify lockfile is in sync
  run: poetry check --lock
```

If there is a deliberate reason to regenerate on CI (platform-specific markers, Python-version-dependent resolution), please surface it in a comment — I couldn't find one from reading the surrounding code, and the `lint.yml` job validates fine without regeneration on the same Python 3.11 runner. The `unit_integration` / `e2e` jobs are on Python 3.10.x; if there's a 3.10-vs-3.11 marker concern, that's worth documenting either way.

aaitor (approved):
The change correctly addresses the fork-PR CI gap and the security analysis in the description is sound — `pull_request` (not `pull_request_target`) is the right primitive, secrets are not passed to fork runs, and the recommended approval-gate flip is the proper containment for spam/abuse. CI is green and the diff is minimal. Approving with three inline defense-in-depth suggestions, one out-of-scope HIGH the maintainer must action independently, and a process nudge.

**Verification I ran**

- `gh pr view 204` / `gh pr diff 204` / `gh pr checks 204` — CI green (lint, unit_integration, e2e all pass on `6e67e7f`).
- Read post-PR `.github/workflows/lint.yml` and `.github/workflows/test.yaml` in full, plus the surrounding workflow ecosystem (`docs.yml`, `release-python.yml`, `publish-mintlify-docs.yml`, `prepare-release.yml`, `finalize-release.yml`) — confirmed the trigger change does NOT break the release pipeline. `release-python.yml` fires on tag push `v*.*` and does not depend on this workflow, so losing test runs on tag pushes is harmless (the same commit already passed CI when it merged to `main`).
- Verified PR #181 state: `statusCheckRollup: []` and external `headRepositoryOwner: peterxing` — confirms the diagnosed bug.
- Dispatched parallel `security-reviewer` (fork-PR threat model + secrets/supply-chain) and `code-reviewer` (CI hygiene) sub-agents. Cross-checked findings against the actual files before promoting any to HIGH.
- Decoded the JWTs in `tests/e2e/conftest.py` and confirmed: `exp` of `4935131697` (≈ year 2126) and a `o11y` claim embedding a Helicone API key. These are pre-existing in `main` but become materially more exposed the moment fork PRs start running CI — see HIGH note below.
- Walked the dual-trigger scenarios: no duplicate runs in practice (push to feature branch doesn't match `branches: [main]`; push to main has no open PR; merge commits fire `push` only). Clean.

**HIGH (out of scope for this PR — please action before flipping the approval gate post-merge)**

`tests/e2e/conftest.py` lines ~37 and ~42 hardcode the staging-sandbox `TEST_SUBSCRIBER_API_KEY` / `TEST_BUILDER_API_KEY` JWTs as `os.getenv()` defaults, and each JWT's `o11y` claim embeds a Helicone API key. These have been in public git history for a while, so this PR does not *create* the exposure — but it does make those tokens part of every external contributor's CI run via module-level conftest collection. Combined with the `poetry lock` issue called out inline, this means a hostile fork could execute arbitrary code on the runner with the JWTs in memory before any maintainer reviews the PR. Suggested follow-up (separate PR, not blocking this one):

- Rotate both staging API keys and the embedded Helicone key.
- Replace the hardcoded `os.getenv(..., "<jwt>")` defaults with `os.getenv(..., "")` and add `pytest.skip(..., allow_module_level=True)` at the top of the e2e conftest when keys are absent.
- Consider `git filter-repo` to scrub history; rotate-only leaves the old tokens in `git log`.

**Process — treat the approval-gate flip as a hard pre-flight, not a follow-up**

The PR body lists "flip the org Actions approval gate" as a post-merge step. Given the e2e-conftest exposure above and the `poetry lock` regen called out inline, I'd strongly recommend flipping the gate *before* this PR is merged (or sitting on the merge for the 30s it takes). Otherwise the very first fork PR push after merge triggers a CI run that (a) regenerates the lock from fork-controlled `pyproject.toml` and (b) loads the e2e conftest with the hardcoded JWTs in memory, all without maintainer review.

Approving on the assumption that either the inline `poetry check --lock` swap lands here, or the approval gate is flipped immediately on merge (preferably both).

PR #215, opened 2026-06-04
Add enableX402Discovery router toggle

### LINKED ISSUE #205: Add dedicated enableX402Discovery toggle in OAuth router config
## Summary

Surfaced during review of #181 (open finding #5). The new `/.well-known/x402-payment` route is currently gated under the existing `enable_oauth_discovery` config flag in `payments_py/mcp/http/oauth_router.py:147`, so disabling OAuth discovery silently turns off x402 discovery too. They're distinct concerns and should be independently controllable.

## Current shape

```python
# payments_py/mcp/http/oauth_router.py
if enable_oauth_discovery:
    # ... existing oauth metadata routes ...

    @router.get("/.well-known/x402-payment")
    async def x402_payment_discovery() -> JSONResponse:
        ...
```

## Desired shape

Mirror the existing `enable_oauth_discovery` / `enable_client_registration` pattern with a new `enable_x402_discovery: bool = True` config flag, gating only the `/.well-known/x402-payment` handler.

```python
if enable_x402_discovery:
    @router.get("/.well-known/x402-payment")
    async def x402_payment_discovery() -> JSONResponse:
        ...
```

The default stays `True` so existing deployments don't lose the endpoint — but a server that wants x402 discovery without OAuth (or vice-versa) can now configure it independently.

## Related quirk worth fixing in the same PR

Pre-existing, surfaced in the same review thread: `build_server_info_response` always advertises the OAuth + x402 URLs in the server-info payload even when their handlers aren't mounted. Whatever toggle gates the route should also gate the server-info advertisement of that route — otherwise the root response promises endpoints that 404.

## Acceptance criteria

- [ ] `enable_x402_discovery: bool = True` added to the OAuth router config (matching the existing toggle naming convention)
- [ ] Route handler is gated on the new flag
- [ ] `build_server_info_response` only advertises `x402_payment` / `x402_payment_discovery` URLs when the flag is `True` (and analogous for the OAuth keys under `enable_oauth_discovery`)
- [ ] Unit / e2e test covers both states (flag on → endpoint reachable + advertised; flag off → 404 + not advertised)

## Related

- Originated in #181 review
- Stability of the discovery payload itself is being tracked under #203 (broader MCP/x402 spec alignment); this issue is purely about the toggle ergonomics

peterxing:
## Summary
- add an `enableX402Discovery` option alongside the existing OAuth discovery toggle
- gate only `/.well-known/x402-payment` on the x402 toggle
- make server-info advertise OAuth/x402 discovery URLs only when their matching handlers are mounted
- add focused metadata/e2e coverage for enabled and disabled discovery states

Closes #205.

## Verification
- `python -m py_compile payments_py\\mcp\\core\\server_manager.py payments_py\\mcp\\http\\oauth_metadata.py payments_py\\mcp\\http\\oauth_router.py payments_py\\mcp\\types\\http_types.py payments_py\\mcp\\types\\server_types.py tests\\e2e\\mcp_tests\\test_mcp_oauth_server.py tests\\unit\\mcp_tests\\test_oauth_metadata.py`
- `python -m pytest tests\\unit\\mcp_tests\\test_oauth_metadata.py -q`
- `git diff --check`

No wallet signing, paid x402 calls, live settlement, credential use, signup, token movement, or chain writes were performed.


r-marques (commented):

_Reviewed `origin/main...HEAD` (7 files) with the pr-review-toolkit panel: code-reviewer, pr-test-analyzer, type-design-analyzer. Read-only, scoped to the diff._

Clean, well-scoped change: `enableX402Discovery` (default `True`) now gates the `/.well-known/x402-payment` route and its server-info advertisement on its own `if enable_x402_discovery:` block, decoupling it from `enableOAuthDiscovery` (where it was previously nested). Config plumbs end-to-end (`server_manager.py` → `create_oauth_router` → `config` → `build_server_info_response`), advertise/mount parity holds for every toggle combination, and black/imports are clean. No correctness or security issues found.

### 🟡 Should fix (2)
- **Default (both-on) x402 advertisement isn't positively asserted** — `tests/unit/mcp_tests/test_oauth_metadata.py:234` (`test_builds_server_info_with_all_endpoints`) asserts `endpoints["mcp"/"health"/"register"]` but never `endpoints["x402_payment"]`, and `test_includes_oauth_endpoints` never asserts `oauth["x402_payment_discovery"]`. The default-on advertisement therefore has no positive guard: an inverted default (`config.get("enableX402Discovery", False)`) would pass the entire suite. One assertion added to the existing default test closes this. _(pr-test-analyzer, confidence high)_
- **`ServerInfoOAuth.registration_endpoint: Optional[str]` is inconsistent with sibling `str` keys** — `payments_py/mcp/types/http_types.py:456`. The TypedDict is `total=False`, and `build_server_info_response` (`oauth_metadata.py:401-406`) **omits** the key entirely when client registration is disabled — it never sets it to `None`. So `Optional[...]` advertises a `None` state that can't occur. The sibling `x402_payment_discovery` (`:452`) correctly uses plain `str` + conditional omit; aligning `registration_endpoint` to `str` matches the actual semantics. Pre-existing pattern, but this PR is the first code to *build* this TypedDict as a partial dict, which makes it observable. _(type-design-analyzer, confidence medium)_

### 💡 Good to have (4)
- **Call out the default-on behavior flip in the PR description / changelog** — a deployment that set `enableOAuthDiscovery=False` and never set `enableX402Discovery` goes from x402 route *suppressed* → *mounted* after this PR. This is the intended decoupling (and the new e2e test pins it), and it exposes only discovery metadata (no secrets/state), so it's low-risk — but operators who deliberately disabled OAuth discovery may be surprised. _(code-reviewer, confidence 0.9)_
- **`enableHealthCheck` / `enableClientRegistration` advertisement gating is now untested** — `oauth_metadata.py:35-38` newly makes `endpoints["health"]`/`endpoints["register"]` conditional (previously unconditional), but no test toggles those off and asserts absence. These are touched lines, so a regression would be unguarded. _(pr-test-analyzer, confidence medium)_
- **Convert the `oauth_info.update({...})` block to direct key assignments** — `oauth_metadata.py:394-400`. A typo'd key inside a `.update()` literal slips past TypedDict key-checking, whereas the direct `oauth_info["..."] = ...` form two lines below would flag it. Uniformity + slightly stronger checking. _(type-design-analyzer, confidence medium)_
- **Optional docs mention** — `docs/api/09-mcp-integration.md` documents the discovery endpoints but not the toggles; the existing `enableOAuthDiscovery` is also undocumented, so adding only the new flag would be inconsistent. A short "discovery endpoints can be toggled via `enableOAuthDiscovery` / `enableX402Discovery`" note would be a nice future addition, not a gap. _(code-reviewer, confidence 0.55)_

### ✅ Strengths
- **Toggle-matrix coverage is the strongest part** — the four quadrants (both-off, oauth-off/x402-on, x402-off/oauth-on) are pinned at both unit and e2e level, asserting both the 404/200 route behavior **and** the server-info advertisement, so route↔advertisement drift can't slip through silently. The decoupling regression (x402 surviving `enableOAuthDiscovery=False`) — the PR's highest-risk behavior — is double-guarded.
- **Fixes a latent pre-existing mismatch** — `build_server_info_response` previously advertised `health`/`register`/OAuth-metadata unconditionally even when those toggles were off; the new conditional `endpoints`/`oauth_info` dicts now gate them, so advertised endpoints match mounted routes.
- **Clean, backward-compatible test-helper refactor** — `create_and_start_server(port, start_options=None)` with `options.update(start_options or {})`.
- **Consistent defaults** — all three read sites use `.get("enableX402Discovery", True)`, so the documented default is honored uniformly with no skew. TypedDict field added in all three config types with matching docstrings; no `pyproject.toml` version bump (correct per release process).

peterxing:
Addressed the two should-fix items in `72eb7c843b5732781750c3ff8a206a0d3fbafca9`:

- Added a default-on assertion for x402 discovery advertisement so `x402_payment` / `x402_payment_discovery` stay guarded when both discovery toggles use defaults.
- Aligned `ServerInfoOAuth.registration_endpoint` with the partial `TypedDict` semantics by typing it as `str`, matching the existing omit-when-disabled behavior.

I also re-ran focused local verification before pushing: `python -m py_compile ...`, `python -m pytest tests/unit/mcp_tests/test_oauth_metadata.py -q`, and `git diff --check`.

r-marques (approved):

r-marques:
Thanks for the contribution, @peterxing! 🙌

These changes shipped in the [**v1.12.0** release](https://github.com/nevermined-io/payments-py/releases/tag/v1.12.0) — `enableX402Discovery` is now available to decouple x402 payment discovery from OAuth discovery. Appreciate the clean, well-tested PR!

PR #217, opened 2026-06-04
fix(langsmith): warn on short tokens + document multi-tool parent clobber

r-marques:
## Description

Observability span hardening surfaced by aaitor's review of payments-py#198 (Sprint 1 of epic nevermined-io/nvm-monorepo#1703). One coherent PR covering two follow-ups. Aligns with the cross-SDK **observability spans v1** spec (nevermined-io/nvm-monorepo#1746, §4 redact+warn).

### Part A — redact short tokens, not just warn (nevermined-io/nvm-monorepo#1747)

`payments_py/langsmith/spans.py::abbreviate_token` hardens the short-token path. A real x402 access token is a JWT (always >20 chars). A token of **20 characters or fewer** is almost always a misconfiguration (a plan id or opaque handle passed where the JWT was expected) — and it must never reach a durable, queryable trace store.

Final behavior:

- **Short (≤20-char, non-empty) tokens are redacted AND warned** — they are surfaced under `nvm.payment_token` as `<first 4 chars>…(short)` (e.g. `eyJ4…(short)`), revealing just enough to debug the misconfig, and a `logging.warning` is emitted. The **full short value never leaves the function**, so a misrouted secret can't leak into a trace even when it's shorter than the abbreviation threshold. (`nvm.payment_token` is purely an observability label — never used functionally — so redacting it has zero functional impact.)
- **Normal JWT tokens (>20 chars)** are abbreviated to `<first 16>…<last 4>` as before, silently.
- **`None` / empty** stay silent and pass through to `None` (no token is not a "wrong token" mistake).
- **Idempotent**: re-applying to an already-redacted (`…(short)`) or already-abbreviated value returns it unchanged and does not re-warn. This matters because the decorator/middleware path abbreviates the same token twice (`attach_metadata_safely` then the builders). An `endswith("…(short)")` guard makes re-application a stable, silent no-op — value-driven, no stateful dedup, the helper stays pure.

> Earlier iteration warned-only and returned the short token verbatim; the review's silent-failure lens correctly flagged that as incomplete for a credential helper (a `logging.warning` can be dropped while the token persists in the trace store). This PR now redacts.

### Part B — multi-tool parent-metadata clobber doc (nevermined-io/nvm-monorepo#1745, doc-only)

Documents a known limitation in `docs/api/12-langchain-integration.md` (new **Known limitations** subsection + an inline caveat on the existing parent-attach sentence): `@requires_payment` writes bare `nvm.*` keys onto the **parent** LangSmith run tree, so a second `@requires_payment` tool in the **same LangGraph `ToolNode`** silently overwrites the first's parent `nvm.*` values (last-writer-wins). The per-tool **child** `nvm:verify` / `nvm:settlement` spans remain authoritative. Operator guidance: filter on the child spans for per-call billing fidelity. No code change (decided = Option 1).

Part of nevermined-io/nvm-monorepo#1703
Closes nevermined-io/nvm-monorepo#1747
Closes nevermined-io/nvm-monorepo#1745

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

## Checklist:

- [x] Follows the code style of this project. (`poetry run black --check .` clean)
- [x] Tests Cover Changes (short-token redaction + idempotency + warn/no-warn boundary tests; full non-slow suite: 577 passed)
- [x] Documentation (LangChain integration docs updated — short-token redaction note + clobber limitation; `mkdocs build` succeeds)

---


payments_py/langsmith/spans.py
```diff
@@ -118,21 +118,49 @@ def redact_metadata_keys(run_tree: Optional[Any], *keys: str) -> None:
...
+        # re-slicing would let the marker drift, so return it unchanged and
+        # stay silent -- the original short value already triggered the warning.
         return token
+    if len(token) <= 20:
+        logger.warning(
+            "abbreviate_token: token is 20 characters or fewer -- was the "
+            "right x402 access token passed? Short/non-JWT tokens are almost "
+            "always a misconfiguration and are redacted (not exported)."
+        )
+        return f"{token[:4]}{_SHORT_TOKEN_MARKER}"
```
aaitor: **LOW — `token[:4]` reveals the entire value for tokens of ≤4 chars.**

The docstring promises "only the first 4 characters are revealed", but for a 1–4 char value `token[:4]` is the whole string — e.g. `"ab"` → `"ab…(short)"` puts the complete value into the durable trace store. Unlikely for a real misconfig (plan ids / handles are longer), and still strictly better than the old verbatim behavior, but cheap to make the contract actually hold:

```suggestion
        prefix = token[:4] if len(token) > 4 else ""
        return f"{prefix}{_SHORT_TOKEN_MARKER}"
```
Worth a `len(token) <= 4` test alongside it.

aaitor (approved):
Genuine security hardening: a ≤20-char value (never a real x402 JWT, which is far longer) is now redacted to `first4…(short)` + a warning instead of stored verbatim in a durable trace store, and the multi-tool parent-clobber limitation is documented honestly. Net security improvement — approving. All findings LOW.

**Verification I ran**
- Read the full diff + post-PR `spans.py` (incl. the `attach_metadata_safely` → builder call chain) and both test files via `gh api …?ref=fix/observability-span-hardening`. Base `main`, 4 files, +116/-11.
- Parallel agent panel: `python-reviewer`, `security-reviewer`, `tdd-guide`. **Cross-checked and down-rated two over-rated HIGHs:**
  - The `token[:16]` long-token path (flagged HIGH) is **pre-existing/unchanged** by this PR (the diff only touches the ≤20-char branch) and **unreachable for real x402 tokens** — their first segment is ~47 chars (`eyJ4NDAyVmVyc2lvbiI6…` → `{"x402Version":2,"roles":["payer"]}`), so `[:16]` stays inside it and exposes only `{"x402Versio`, never the signature. Noted as an optional follow-up, not a gate.
  - The "2 warnings, not 1" (flagged HIGH) is **real but harmless**: a full verify+settle flow calls `attach_metadata_safely` twice, each with the raw token → 2 warnings per misconfigured request. The idempotency guard correctly prevents marker *drift* and a second warn *within* each call (verified: `attach_metadata_safely:84` reassigns the kwarg to the abbreviated output; the builder then re-abbreviates that marked value). Log-noise for a misconfig, not a correctness issue.
- **CI: green** — `lint`, `unit_integration`, `e2e` all ✅.

**Non-blocking (LOW)**
- **`token[:4]` reveals the whole value for ≤4-char tokens** (inline suggestion): cap the prefix so something is always redacted, and add a `len(token) <= 4` test.
- **Idempotency guard is `endswith(marker)`-only** — a ≤20-char value that itself ends in `…(short)` would pass through verbatim. Astronomically unlikely (U+2026 isn't in base64url), but also bounding the redacted form's length (`len ≤ 4 + len(_SHORT_TOKEN_MARKER)`) turns the guard into a credential-safe classifier rather than relying on the suffix alone.
- **Test fence-post:** the 20-char boundary is covered from below, but there's no 21-char case (first length that must *not* warn and must use `first16…last4`); a `< 20` / `<= 21` regression would slip through. Also consider (a) asserting the raw token never appears as a *value* in the decorator's parent-metadata payload (defence-in-depth), and (b) tightening the 20-char test's `any(...)` → `len(...) == 1` for consistency with the 14-char sibling. `test_abbreviate_token_none_or_empty_returns_none` is now subsumed by the `…_does_not_warn` variant.
- **Optional follow-up (pre-existing):** clip the long-token path at the first `.` (`token[:min(16, token.find('.'))]` when a dot exists) so `[:16]` can never cross into a JWT payload segment regardless of token shape.
- Doc nit: "always >20 chars" reads as "a 21-char JWT is normal"; phrasing it as the heuristic it is (real x402 JWTs are far longer) reads truer.

r-marques:
Thanks @aaitor — all LOW nits addressed in 6d2ca34:

- **Closed the ≤4-char gap:** a 1–4 char token now collapses to just `…(short)` (reveals nothing), so the docstring's contract actually holds; test added over {"a","ab","abc","abcd"}.
- **Idempotency guard is now a credential-safe classifier** (`_is_redacted_marker`: `…(short)` suffix **and** exact marker length ∈ {8, 12}), so a raw ≤20-char value that merely ends in the marker (e.g. `leakme…(short)`) is redacted, not passed through verbatim; test added.
- **Added the 21-char fence-post test** (first length that must not warn / must use `first16…last4`) and a defence-in-depth assert that the raw token never appears as a value in the parent-metadata payload.
- Rephrased the ">20 chars" doc to the heuristic it is.

Skipped the pre-existing long-token `token[:16]` clip as out-of-scope per your note. CI green.

PR #219, opened 2026-06-08
fix(langchain): mirror TS parity fixes for stale receipt + token leak

r-marques:
## What

Two parity follow-ups in `payments_py/x402/langchain/decorator.py`, mirroring fixes already merged on the TypeScript side (`@nevermined-io/payments` PRs #370 and #371) and tracked under epic nevermined-io/nvm-monorepo#1703. A second commit folds in aaitor's review.

### 1. Reset `last_settlement()` at invocation start — nevermined-io/nvm-monorepo#1890

The module-level receipt slot was only written on settle-success and never reset. After a prior successful settle, a subsequent invocation that failed before/at settle (a verify `PaymentRequiredError`, or a swallowed settle failure) left `last_settlement()` returning the **prior** call's receipt — misattributing its `transaction` / `credits_redeemed`.

**Fix:** reset the slot to `None` at the start of both the sync and async entry points, before verify, so any pre-settle failure yields `None`. The public `last_settlement()` docstring is updated to cover the swallowed-settle-failure case. Mirrors TS `95326a8`.

### 2. Stop leaking the raw token through `config.configurable` — nevermined-io/nvm-monorepo#1891 (+ residual leak from review)

Two leak surfaces, both closed:

- **`payment_context.token`** now stores `abbreviate_token(token)` (the non-functional `<first 16>…<last 4>` form surfaced as `nvm.payment_token`), not the raw credential. Mirrors TS `6bc08a6`.
- **the raw `payment_token` key** is now popped from `config.configurable` immediately after extraction. LangChain's `ensure_config` re-promotes configurable scalars into each runnable's `metadata`, so a traced child runnable spawned by the tool body would otherwise re-leak the full credential into its own span — `redact_metadata_keys` only scrubs the parent run-tree once. (Surfaced by aaitor's review; reverse-parity tracked for TS in nevermined-io/nvm-monorepo#1901.)

Settlement reads the raw token from the internal `_VerifiedPayment` container, never `configurable` or `payment_context.token`, so both removals are non-functional. On the `@tool` path LangChain hands the decorator a per-invocation config copy, so sibling tools are unaffected.

## Note on the shared `PaymentContext` type

Unlike the TS `PaymentContext` (langchain-local), the Python one is **shared** with the FastAPI/Starlette middlewares, which legitimately store the **raw** token (they strip the `payment-signature` header, so it's the handler's only source). So the `PaymentContext.token` docstring (`x402/types.py`) gains a **producer-aware** warning rather than a blanket "always abbreviated" claim. No middleware behaviour changed.

## Tests

- +2 stale-receipt regressions (sync: `prior-success → failed-settle → None`, `… → verify-failure → None`) **+1 async** (`…_async`) so the async reset line can't silently regress
- +1 leak test: `payment_context.token` is abbreviated, **every** `PaymentContext` field is swept for the raw token, the raw `payment_token` key is gone from `configurable`, and settle still ran with the raw token

`black --check` clean; targeted unit suites green (60 passed in ~0.5s). Package version intentionally **not** bumped — the release pipeline owns it.

> ℹ️ The linked issues live in `nvm-monorepo`, so GitHub's auto-close won't fire cross-repo on merge — they'll be closed manually.

### DIFF
--- a/payments_py/x402/langchain/decorator.py
+++ b/payments_py/x402/langchain/decorator.py
@@ -111,14 +111,22 @@ def last_settlement() -> Optional[SettleResponse]:
     ``network``, ``payer``) without threading it back through the
     runnable config (which LangGraph copies per node).
 
-    Returns ``None`` if no settlement has happened yet in this process, or
-    if the most recent invocation raised before reaching the settle phase.
+    Returns ``None`` if no settlement has happened yet in this process, or if
+    the most recent invocation did not reach a *successful* settle — i.e. it
+    raised before the settle phase (a verify failure / ``PaymentRequiredError``)
+    **or** it reached settle but settlement failed (settle errors are swallowed
+    and logged, so the tool still returns its result while ``last_settlement()``
+    stays ``None``). The slot is reset to ``None`` at the start of every
+    invocation, so a prior invocation's receipt is never misattributed to a
+    later call that failed to settle.
 
     .. note::
        This accessor reads from a module-level slot. In multi-tenant
        processes (e.g. a server handling concurrent settlements), the value
        reflects whichever invocation settled most recently — there is no
-       per-call isolation.
+       per-call isolation. Because the slot is also reset on entry, a
+       concurrent invocation that merely *starts* (even one that fails verify)
+       nulls a sibling's freshly written receipt.
     """
     return _LAST_SETTLEMENT["value"]
 
@@ -194,6 +202,19 @@ def _store_in_configurable(config: Any, key: str, value: Any) -> None:
         configurable[key] = value
 
 
+def _remove_from_configurable(config: Any, key: str) -> None:
+    """Remove a key from config["configurable"] if present (no-op otherwise)."""
+    if config is None:
+        return
+    configurable = None
+    if isinstance(config, dict):
+        configurable = config.get("configurable")
+    else:
+        configurable = getattr(config, "configurable", None)
+    if isinstance(configurable, dict):
+        configurable.pop(key, None)
+
+
 def _verify_payment(
     func: Callable,
     kwargs: dict,
@@ -254,6 +275,22 @@ def _verify_payment(
                 "Payment required: missing payment_token in config['configurable']",
                 payment_required,
             )
+
+        # Drop the raw token from configurable now that we hold it in ``token``
+        # (and, below, in ``_VerifiedPayment.token``). LangChain's
+        # ``ensure_config`` re-promotes every configurable scalar into a
+        # runnable's ``metadata`` on EACH invocation, so any traced runnable the
+        # tool body spawns (an LLM call, a sub-chain) sharing this config would
+        # otherwise re-leak the full credential into its own span metadata —
+        # ``redact_metadata_keys`` above only scrubs the parent run-tree once.
+        # This MUST run after extraction (popping earlier would make
+        # ``_extract_payment_token`` return None) and before the tool body runs,
+        # which is guaranteed: ``_verify_payment`` completes before ``func`` is
+        # called. Settlement reads ``_VerifiedPayment.token``, never configurable,
+        # so removal is non-functional. On the ``@tool`` path LangChain hands the
+        # decorator a per-invocation config copy, so this does not mutate a config
+        # the caller reuses across sibling tools.
+        _remove_from_configurable(runnable_config, "payment_token")
 
         credits_pre = _resolve_credits_pre(config.credits, kwargs)
         # For verification, use pre-resolved credits or default to 1
--- a/payments_py/x402/langchain/decorator.py
+++ b/payments_py/x402/langchain/decorator.py
@@ -68,6 +68,7 @@ from dataclasses import dataclass
 from typing import Any, Callable, Optional, Union
 
 from payments_py.langsmith.spans import (
+    abbreviate_token,
     active_run_tree,
     attach_metadata_safely,
     build_settle_metadata,
@@ -288,8 +289,19 @@ def _verify_payment(
                 payment_required,
             )
 
+    # The ``token`` field carries the ABBREVIATED reference, never the raw
+    # credential: this PaymentContext is written into ``config["configurable"]``,
+    # which LangChain can capture into span metadata, so persisting the full
+    # token here would reopen the very leak the parent-tree
+    # ``redact_metadata_keys("payment_token")`` call above closes — and a child
+    # run opened *during* the tool body would capture it before any post-hoc
+    # redaction could run. Settlement uses ``_VerifiedPayment.token`` (the raw
+    # token in the internal container below), never this field, so abbreviating
+    # here is non-functional. ``abbreviate_token`` is the same helper surfaced as
+    # ``nvm.payment_token``; ``or ""`` is belt-and-suspenders past the ``not
+    # token`` guard above and can never fall back to the raw credential.
     payment_context = PaymentContext(
-        token=token,
+        token=abbreviate_token(token) or "",
         payment_required=payment_required,
         credits_to_settle=credits_to_charge,
         verified=True,
@@ -493,6 +505,12 @@ def _execute_with_payment(
     config: _PaymentConfig,
 ) -> Any:
     """Synchronous payment verification, execution, and settlement."""
+    # Reset the module-level receipt slot at the START of every invocation,
+    # before verify. Any failure that does not reach the settle-success write
+    # (a verify failure / PaymentRequiredError, or a swallowed settle failure)
+    # then leaves last_settlement() returning None rather than a stale receipt
+    # from a previous invocation — matching the last_settlement() docstring.
+    _LAST_SETTLEMENT["value"] = None
     runnable_config = _extract_runnable_config(kwargs)
     verified = _verify_payment(func, kwargs, config, runnable_config)
     result = func(*args, **kwargs)
@@ -507,6 +525,10 @@ async def _execute_with_payment_async(
     config: _PaymentConfig,
 ) -> Any:
     """Async payment verification, execution, and settlement."""
+    # Reset the receipt slot at the start of the invocation, before verify —
+    # see _execute_with_payment for the rationale. Both entry points must do
+    # this so the stale-receipt guard holds on the async path too.
+    _LAST_SETTLEMENT["value"] = None
     runnable_config = _extract_runnable_config(kwargs)
     verified = _verify_payment(func, kwargs, config, runnable_config)
     result = await func(*args, **kwargs)
--- a/payments_py/x402/types.py
+++ b/payments_py/x402/types.py
@@ -282,6 +282,20 @@ class PaymentContext:
     Populated by middleware/decorators and made available to handlers:
     - FastAPI: ``request.state.payment_context``
     - Strands: ``tool_context.invocation_state["payment_context"]``
+
+    .. warning::
+       The ``token`` field is **producer-dependent** and must not be assumed to
+       be the functional credential. The FastAPI / Starlette middlewares store
+       the **raw** x402 access token here on purpose: they strip the
+       ``payment-signature`` request header before calling the route handler, so
+       ``payment_context.token`` is the handler's only remaining source for it.
+       The LangChain ``requires_payment`` decorator, by contrast, stores an
+       **abbreviated, non-functional** reference (the same ``<first 16>…<last 4>``
+       form surfaced as ``nvm.payment_token``): its context object is written
+       into ``config["configurable"]``, which tracing frameworks can capture into
+       span metadata, so persisting the raw token there would leak it into nested
+       runs. Code that needs the raw token under the decorator must read it from
+       ``config["configurable"]["payment_token"]``, never from this field.
     """
 
     token: str
--- a/tests/unit/x402/test_langchain_decorator.py
+++ b/tests/unit/x402/test_langchain_decorator.py
@@ -99,6 +99,44 @@ class TestRequiresPaymentDecorator:
 
         assert result == "insight for x"
 
+    def test_payment_context_stores_only_abbreviated_token(self, mock_payments):
+        """The stored payment_context carries the ABBREVIATED token, never the
+        raw credential (nvm-monorepo#1891).
+
+        ``payment_context`` is written into ``config["configurable"]``, which
+        LangChain can capture into span metadata, so a raw token here would let
+        the credential ride into a nested span opened during the tool body — the
+        gap aaitor flagged on the TS port. We invoke the bare ``@requires_payment``
+        wrapper directly (not through ``@tool``, which deep-copies the config) so
+        the decorator mutates the same configurable bag we then read back.
+        """
+        long_token = "j" * 40
+
+        @requires_payment(payments=mock_payments, plan_id="plan-123", credits=1)
+        def my_tool(topic: str, config: RunnableConfig = None) -> str:
+            """Return a canned string."""
+            return f"insight for {topic}"
+
+        config = {"configurable": {"payment_token": long_token}}
+        my_tool(topic="evs", config=config)
+
+        ctx = config["configurable"]["payment_context"]
+        assert ctx is not None
+        # Abbreviated form is <first 16>…<last 4>, the same surfaced as
+        # nvm.payment_token — not the raw 40-char credential.
+        assert ctx.token == f"{'j' * 16}…jjjj"
+        # The raw token must not appear anywhere in the stored context.
+        assert long_token not in repr(ctx)
+        # Settlement still ran with the raw token (internal _VerifiedPayment),
+        # so abbreviating payment_context.token is non-functional.
+        mock_payments.facilitator.settle_permissions.assert_called_once()
+        assert (
+            mock_payments.facilitator.settle_permissions.call_args.kwargs[
+                "x402_access_token"
+            ]
+            == long_token
+        )
+
 
 class TestLastSettlement:
     def test_returns_none_before_any_settlement(self):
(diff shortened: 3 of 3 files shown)


payments_py/x402/langchain/decorator.py
```diff
@@ -288,8 +289,19 @@ def _verify_payment(
                 payment_required,
             )
 
+    # The ``token`` field carries the ABBREVIATED reference, never the raw
+    # credential: this PaymentContext is written into ``config["configurable"]``,
+    # which LangChain can capture into span metadata, so persisting the full
+    # token here would reopen the very leak the parent-tree
+    # ``redact_metadata_keys("payment_token")`` call above closes — and a child
```
aaitor, line 333: **MEDIUM — the "primary vector is already defended by `redact_metadata_keys("payment_token")`" claim (here and in the PR description) is incomplete; the raw token still leaks via child runnables.**

This PR correctly closes the *derived* copy (`payment_context.token`). But the **raw** `payment_token` is never removed from `config["configurable"]` — and `redact_metadata_keys(parent_rt, …)` at L224 only strips the **parent** LangSmith run-tree's metadata in place, at one moment.

Verified against the installed `langchain_core` — `ensure_config()` (`runnables/config.py` L192–198) promotes every non-`__` configurable scalar into `metadata` on **each** runnable invocation:
```python
for key, value in empty.get("configurable", {}).items():
    if (not key.startswith("__") and isinstance(value, (str, int, float, bool))
            and key not in empty["metadata"]):
        empty["metadata"][key] = value
```
So any traced LangChain runnable the tool body spawns (an LLM call, a sub-chain) that receives this config re-promotes the raw JWT into its own span metadata — exactly the class of leak this PR is fixing for `payment_context.token`. The decorator's *own* `verify_span`/`settlement_span` are fine (they inherit the already-redacted parent metadata); leaf tools are fine (no children). The gap is real only when the tool body invokes further traced runnables sharing the config.

This is **pre-existing** (not introduced here) and the PR is a net improvement, so it shouldn't block merge — but please either soften the "already defended" wording and file a follow-up, or close it directly by dropping the raw key after extraction:
```python
# In _verify_payment, right after redact_metadata_keys(parent_rt, "payment_token") (L224):
# ensure_config() re-promotes configurable scalars into child-run metadata, so the
# raw token must not survive in configurable either. We already hold it in `token`
# (and _VerifiedPayment.token below); nothing downstream reads it from configurable.
if isinstance(runnable_config, dict):
    cfg = runnable_config.get("configurable")
    if isinstance(cfg, dict):
        cfg.pop("payment_token", None)
```
Caveat: this mutates the caller's config (removes a key they set). The decorator already mutates `configurable` (adds `payment_context`/`payment_settlement`), but *deleting the input key* is more surprising — if a caller reuses one config object across two protected tools, the second would lose the token. If that's a concern, the follow-up-issue route is the safer choice.

payments_py/x402/langchain/decorator.py
```diff
@@ -493,6 +505,12 @@ def _execute_with_payment(
     config: _PaymentConfig,
 ) -> Any:
     """Synchronous payment verification, execution, and settlement."""
+    # Reset the module-level receipt slot at the START of every invocation,
+    # before verify. Any failure that does not reach the settle-success write
+    # (a verify failure / PaymentRequiredError, or a swallowed settle failure)
+    # then leaves last_settlement() returning None rather than a stale receipt
+    # from a previous invocation — matching the last_settlement() docstring.
+    _LAST_SETTLEMENT["value"] = None
```
aaitor, line 550: **MEDIUM — this reset makes the public `last_settlement()` docstring (L114–115) stale.** It currently reads:

> Returns `None` if no settlement has happened yet in this process, or if the most recent invocation **raised before reaching the settle phase**.

With this reset, `last_settlement()` also returns `None` after a **swallowed settle failure** — where the invocation does *not* raise (it returns the tool result normally; see `_settle_payment` L370–371, which only logs) yet the slot stays `None`. A consumer whose tool returned successfully would reasonably expect a receipt and get `None`. Your inline comment at L508–512 describes the real behavior correctly; the public docstring just needs to catch up:
```python
    Returns ``None`` if no settlement has happened yet in this process, or if
    the most recent invocation did not reach a *successful* settle -- i.e. it
    raised before the settle phase (a verify failure / ``PaymentRequiredError``)
    **or** it reached settle but settlement failed (settle errors are swallowed
    and logged, so the tool still returns its result while ``last_settlement()``
    stays ``None``).
```

Minor, same area (non-blocking): resetting on entry slightly widens the documented single-tenant race — a concurrent invocation B now nulls A's freshly-written receipt just by *starting* (even a B that fails verify), where before only a settle-success overwrote it. Covered by the existing "no per-call isolation" note, so no change needed; flagging for awareness.

payments_py/x402/langchain/decorator.py
```diff
@@ -507,6 +525,10 @@ async def _execute_with_payment_async(
     config: _PaymentConfig,
 ) -> Any:
     """Async payment verification, execution, and settlement."""
+    # Reset the receipt slot at the start of the invocation, before verify —
+    # see _execute_with_payment for the rationale. Both entry points must do
+    # this so the stale-receipt guard holds on the async path too.
+    _LAST_SETTLEMENT["value"] = None
```
aaitor, line 568: **MEDIUM — the async reset has no test.** Both new regression tests (`test_none_after_prior_success_then_failed_settle`, `…_then_verify_failure`) drive the **sync** `.invoke` path. This reset line is the async twin of the bug being fixed; if a future refactor dropped *only* this line, every async (`@tool async def …`) tool would silently regress to stale receipts and the whole suite would still pass. Mirror one regression on `.ainvoke`:
```python
import pytest

def _make_protected_async_tool(mock_payments, *, credits=1):
    @tool
    @requires_payment(payments=mock_payments, plan_id="plan-123", credits=credits)
    async def my_tool(topic: str, config: RunnableConfig = None) -> str:
        """Return a canned string."""
        return f"insight for {topic}"
    return my_tool

@pytest.mark.asyncio
async def test_none_after_prior_success_then_failed_settle_async(self, mock_payments):
    my_tool = _make_protected_async_tool(mock_payments)
    await my_tool.ainvoke({"topic": "a"}, config={"configurable": {"payment_token": "tok"}})
    assert last_settlement().credits_redeemed == "1"
    mock_payments.facilitator.settle_permissions.side_effect = RuntimeError("boom")
    result = await my_tool.ainvoke({"topic": "b"}, config={"configurable": {"payment_token": "tok"}})
    assert result == "insight for b"
    assert last_settlement() is None
```
(If the suite isn't already async-enabled, add `pytest-asyncio` and `asyncio_mode = auto`, or decorate explicitly.)

tests/unit/x402/test_langchain_decorator.py
```diff
@@ -99,6 +99,44 @@ def test_settle_failure_does_not_break_result(self, mock_payments):
...
+        config = {"configurable": {"payment_token": long_token}}
+        my_tool(topic="evs", config=config)
+
+        ctx = config["configurable"]["payment_context"]
+        assert ctx is not None
+        # Abbreviated form is <first 16>…<last 4>, the same surfaced as
+        # nvm.payment_token — not the raw 40-char credential.
+        assert ctx.token == f"{'j' * 16}…jjjj"
+        # The raw token must not appear anywhere in the stored context.
+        assert long_token not in repr(ctx)
```
aaitor: **LOW — `assert long_token not in repr(ctx)` is near-redundant and only guards `token`.** The exact-match assert on L127 already pins `ctx.token` to the abbreviated form, so this line passes the same way; and because the abbreviated value's longest `j`-run is 16 (< 40), the raw 40-char string can never be a substring regardless. Its stated intent — "must not appear anywhere in the stored context" — is better served by sweeping every field, which also guards any future field that might echo the token:
```python
        # Sweep every field, not just `token` — guards future leak surfaces too.
        for name, value in vars(ctx).items():
            assert long_token not in str(value), f"raw token leaked in PaymentContext.{name}"
```

aaitor (approved):
Clean, well-reasoned parity follow-up. Both fixes are correct and verified end-to-end: `payment_context.token` is non-functional (settle reads `verified.token` from the internal `_VerifiedPayment` at L314/L346, never `payment_context.token`), and the `_LAST_SETTLEMENT` reset is placed before verify on **both** the sync (L513) and async (L531) entry points so any pre-settle-success failure yields `None`. The producer-aware `PaymentContext.token` docstring in `types.py` is the right call given the type is shared with the FastAPI/Starlette middlewares. Approving — none of the findings below are regressions this PR introduces, and CI is green. The three MEDIUMs are worth addressing (the async test and the docstring are quick; the residual `payment_token` leak is pre-existing and should be tracked as a follow-up + the description claim softened).

**Verification I ran**
- `gh pr view/diff/checks` on `fix/langchain-decorator-parity-1890-1891` (base `main`). CI: 3/3 green.
- Read all 3 changed files post-PR (`?ref=<head>`), plus `abbreviate_token`, `redact_metadata_keys`, `_VerifiedPayment`, `_settle_payment`, `_extract_payment_token`, and the `last_settlement()` docstring, to anchor lines and check claims.
- **Verified the leak claim against real langchain_core**: `ensure_config` (`runnables/config.py` L192–198) promotes every non-`__` configurable scalar into `metadata` on each runnable invocation → the raw `payment_token` left in `config['configurable']` is re-promoted into child-runnable metadata. `redact_metadata_keys(parent_rt, …)` (L224) only strips the **parent** run-tree once and never removes the key from configurable. See inline comment #1.
- Confirmed settle uses the raw token (`verified.token`, L346), so abbreviating `payment_context.token` is non-functional — the PR's central claim holds.
- Parent branch is `main` (no feature chain). TS parity commits (`95326a8`, `6bc08a6`) live in a different repo and were not cross-fetched; the Python behavior is internally consistent.
- **Agent panel:** `python-reviewer` (idioms — only out-of-scope/pre-existing nits: `TypedDict` for the slot, async wrapper blocks the loop), `security-reviewer` (surfaced the verified residual `payment_token` re-promotion leak), `tdd-guide` (async coverage gap + the weak `repr` assertion). Dropped as noise/out-of-scope: RST bullet style, `abbreviate_token` short-token logging (pre-existing, by-design), speculative `payment_settlement` echo, `or ""`→`cast` (the inline comment already justifies `or ""`, and it's needed so the type is `str` not `Optional[str]`).

**For the next PR / a follow-up issue (out of scope here):**
- File a tracking issue for the residual raw-token leak (comment #1) and soften the description's "the primary vector is already defended" — it's defended for the parent span and the decorator's own `ls.trace` children, but not for arbitrary traced runnables the tool body spawns.
- The async verify/settle calls (`verify_permissions`/`settle_permissions`) are synchronous HTTP and block the event loop on the `_execute_with_payment_async` path — pre-existing, mirrors TS, worth an `asyncio.to_thread` follow-up.

aaitor (dismissed):
Clean, well-reasoned parity follow-up. Both fixes are correct and verified end-to-end: `payment_context.token` is non-functional (settle reads `verified.token` from the internal `_VerifiedPayment` at L314/L346, never `payment_context.token`), and the `_LAST_SETTLEMENT` reset is placed before verify on **both** the sync (L513) and async (L531) entry points so any pre-settle-success failure yields `None`. The producer-aware `PaymentContext.token` docstring in `types.py` is the right call given the type is shared with the FastAPI/Starlette middlewares. Approving — none of the findings below are regressions this PR introduces, and CI is green. The three MEDIUMs are worth addressing (the async test and the docstring are quick; the residual `payment_token` leak is pre-existing and should be tracked as a follow-up + the description claim softened).

**Verification I ran**
- `gh pr view/diff/checks` on `fix/langchain-decorator-parity-1890-1891` (base `main`). CI: 3/3 green.
- Read all 3 changed files post-PR (`?ref=<head>`), plus `abbreviate_token`, `redact_metadata_keys`, `_VerifiedPayment`, `_settle_payment`, `_extract_payment_token`, and the `last_settlement()` docstring, to anchor lines and check claims.
- **Verified the leak claim against real langchain_core**: `ensure_config` (`runnables/config.py` L192–198) promotes every non-`__` configurable scalar into `metadata` on each runnable invocation → the raw `payment_token` left in `config['configurable']` is re-promoted into child-runnable metadata. `redact_metadata_keys(parent_rt, …)` (L224) only strips the **parent** run-tree once and never removes the key from configurable. See inline comment #1.
- Confirmed settle uses the raw token (`verified.token`, L346), so abbreviating `payment_context.token` is non-functional — the PR's central claim holds.
- Parent branch is `main` (no feature chain). TS parity commits (`95326a8`, `6bc08a6`) live in a different repo and were not cross-fetched; the Python behavior is internally consistent.
- **Agent panel:** `python-reviewer` (idioms — only out-of-scope/pre-existing nits: `TypedDict` for the slot, async wrapper blocks the loop), `security-reviewer` (surfaced the verified residual `payment_token` re-promotion leak), `tdd-guide` (async coverage gap + the weak `repr` assertion). Dropped as noise/out-of-scope: RST bullet style, `abbreviate_token` short-token logging (pre-existing, by-design), speculative `payment_settlement` echo, `or ""`→`cast` (the inline comment already justifies `or ""`, and it's needed so the type is `str` not `Optional[str]`).

**For the next PR / a follow-up issue (out of scope here):**
- File a tracking issue for the residual raw-token leak (comment #1) and soften the description's "the primary vector is already defended" — it's defended for the parent span and the decorator's own `ls.trace` children, but not for arbitrary traced runnables the tool body spawns.
- The async verify/settle calls (`verify_permissions`/`settle_permissions`) are synchronous HTTP and block the event loop on the `_execute_with_payment_async` path — pre-existing, mirrors TS, worth an `asyncio.to_thread` follow-up.

r-marques:
Thanks @aaitor — all four findings addressed in 4e6c22f.

- **MEDIUM ① (residual raw-token leak):** closed directly. The raw `payment_token` is now popped from `config.configurable` right after extraction (new `_remove_from_configurable` helper), so `ensure_config`'s re-promotion into child-runnable metadata has nothing to leak. One nuance on your suggested placement: popping *right after `redact_metadata_keys` at L224* would break `_extract_payment_token` (it reads the key at L250), so the pop runs **after** extraction instead — still before the tool body (verify completes first). Settlement uses `_VerifiedPayment.token`, so it's non-functional. Description softened, and reverse-parity for the TS port filed as nevermined-io/nvm-monorepo#1901.
- **MEDIUM ② (stale docstring):** `last_settlement()` docstring now documents the swallowed-settle-failure case and the start-of-invocation reset / race note.
- **MEDIUM ③ (async reset untested):** added `test_none_after_prior_success_then_failed_settle_async` on `.ainvoke`.
- **LOW ④ (weak assertion):** the leak test now sweeps every `PaymentContext` field via `vars(ctx)` and additionally asserts the raw `payment_token` key is gone from `configurable`.

`black` clean; targeted unit suites 60 passed.

payments_py/x402/types.py
```diff
@@ -282,6 +282,20 @@ class PaymentContext:
...
+       the **raw** x402 access token here on purpose: they strip the
+       ``payment-signature`` request header before calling the route handler, so
+       ``payment_context.token`` is the handler's only remaining source for it.
+       The LangChain ``requires_payment`` decorator, by contrast, stores an
+       **abbreviated, non-functional** reference (the same ``<first 16>…<last 4>``
+       form surfaced as ``nvm.payment_token``): its context object is written
+       into ``config["configurable"]``, which tracing frameworks can capture into
+       span metadata, so persisting the raw token there would leak it into nested
+       runs. Code that needs the raw token under the decorator must read it from
+       ``config["configurable"]["payment_token"]``, never from this field.
```
aaitor, line 298: **MEDIUM — docstring points integrators at a key this PR removes.**

This `.. warning::` (added in this commit) tells callers: *"Code that needs the raw token under the decorator must read it from `config["configurable"]["payment_token"]`."* But the other half of this commit — `_remove_from_configurable(runnable_config, "payment_token")` at `decorator.py:293` — **pops that exact key** before the tool body runs. A developer who follows this guidance inside a `requires_payment`-decorated tool body will always get `None`. The two halves of the PR contradict each other, and this is the one genuinely new defect the commit introduces (all three review agents flagged it independently).

The raw token is intentionally unavailable to the decorated tool body by design — say that instead:

```suggestion
       runs. Under the LangChain ``requires_payment`` decorator the raw token is
       intentionally **removed** from ``config["configurable"]`` before the tool
       body runs (see ``_remove_from_configurable`` in ``langchain/decorator.py``),
       precisely so it can't leak into nested spans — it is therefore *not*
       retrievable from ``configurable`` inside a decorated tool body. Only the
       FastAPI / Starlette path leaves the raw token in this ``token`` field.
```

payments_py/x402/langchain/decorator.py
```diff
@@ -193,6 +202,19 @@ def _store_in_configurable(config: Any, key: str, value: Any) -> None:
         configurable[key] = value
 
 
+def _remove_from_configurable(config: Any, key: str) -> None:
```
aaitor, line 205: **MEDIUM — third copy of the `configurable`-extraction boilerplate; this helper was the moment to factor it.**

`_remove_from_configurable` (new here), `_store_in_configurable` (L192) and `_extract_payment_token` (L178) now carry three byte-identical copies of the dict-vs-attr unwrap (including the dead `configurable = None` init that's overwritten on every branch). Any future change to how `configurable` is located must be made in three places. Extracting a single resolver collapses all three and drops the dead init:

```python
def _get_configurable(config: Any) -> Optional[dict]:
    """Return config[\"configurable\"] as a dict, or None."""
    if config is None:
        return None
    raw = (
        config.get("configurable")
        if isinstance(config, dict)
        else getattr(config, "configurable", None)
    )
    return raw if isinstance(raw, dict) else None


def _extract_payment_token(config: Any) -> Optional[str]:
    c = _get_configurable(config)
    return c.get("payment_token") if c is not None else None


def _store_in_configurable(config: Any, key: str, value: Any) -> None:
    c = _get_configurable(config)
    if c is not None:
        c[key] = value


def _remove_from_configurable(config: Any, key: str) -> None:
    c = _get_configurable(config)
    if c is not None:
        c.pop(key, None)
```

Non-blocking; fine as a small cleanup commit.

tests/unit/x402/test_langchain_decorator.py
```diff
@@ -99,6 +99,56 @@ def test_settle_failure_does_not_break_result(self, mock_payments):
...
+        assert ctx is not None
+        # Abbreviated form is <first 16>…<last 4>, the same surfaced as
+        # nvm.payment_token — not the raw 40-char credential.
+        assert ctx.token == f"{'j' * 16}…jjjj"
+        # Sweep every PaymentContext field — guards future leak surfaces too.
+        for name, value in vars(ctx).items():
+            assert long_token not in str(value), f"raw token leaked in .{name}"
+        # The raw payment_token key was removed from configurable, so any child
+        # runnable sharing this config sees no credential to re-promote.
+        assert "payment_token" not in configurable
```
aaitor, line 141: **LOW — the PR's load-bearing safety claim is untested: that the `@tool` path doesn't mutate the caller's dict.**

`test_no_raw_token_survives_in_configurable` deliberately drives the **bare** wrapper to observe the pop, which means it *relies on* mutating the caller's `configurable`. The decorator comment (`decorator.py:290-292`) justifies the pop precisely because *"On the `@tool` path LangChain hands the decorator a per-invocation config copy"* — but nothing asserts that. If a future LangChain version stops copying, a caller reusing one config dict across sibling tools would silently lose `payment_token` on the second call and get a `PaymentRequiredError`. A one-shot regression pins the assumption:

```python
def test_tool_path_does_not_mutate_callers_config(self, mock_payments):
    """@tool copies config per invocation, so the caller's dict must keep
    payment_token for reuse across sibling tools."""
    my_tool = _make_protected_tool(mock_payments)  # wrapped with @tool
    caller_config = {"configurable": {"payment_token": "tok-original"}}

    my_tool.invoke({"topic": "x"}, config=caller_config)

    assert caller_config["configurable"].get("payment_token") == "tok-original", (
        "LangChain @tool no longer copies config per invocation — "
        "_remove_from_configurable now mutates the caller's dict"
    )
```

Also minor: there's no async twin of this pop test, and the `config=None` path through `_remove_from_configurable` (guarded by `if config is None`) is unexercised — both optional.

aaitor (approved):
Round-2 follow-up that addresses all four findings from the prior review (commit `4e6c22f`). The substance is sound: I re-verified that settlement reads the **raw** token from `_VerifiedPayment.token` (`decorator.py:383`), never the now-abbreviated `payment_context.token`, so both the `abbreviate_token` swap and the new `payment_token` pop are non-functional for the paid call; `abbreviate_token` never returns a raw credential for tokens >20 chars and reveals ≤4 chars otherwise (`langsmith/spans.py:172-182`); and the `_LAST_SETTLEMENT` reset is correctly placed before verify on **both** entry points (`decorator.py:550`, `:568`). The async regression test genuinely guards the async reset line. Net direction is a real security improvement.

One genuinely new defect slipped in with this commit: the `PaymentContext` docstring added here tells integrators to read the raw token from `config["configurable"]["payment_token"]` — but the same commit pops that exact key before the tool body runs, so the guidance points at a key this PR deletes. That plus a small DRY/test-coverage cleanup are the only things worth touching; none gate merge. Approving.

**Verification I ran**
- `gh pr view/diff/checks` on `fix/langchain-decorator-parity-1890-1891` (base `main`, no feature chain). CI: 3/3 green at `4e6c22f`.
- Read all three changed files post-PR (`?ref=<head>`) plus `abbreviate_token` / `redact_metadata_keys` / `build_*_metadata` in `langsmith/spans.py`, and grepped the whole tree for `payment_token` consumers — the only reader of `config["configurable"]["payment_token"]` is `_extract_payment_token` (`decorator.py:272`), which runs **before** the pop, so the pop causes no functional regression inside the SDK.
- Confirmed `_settle_payment` uses `verified.token` (raw), so the abbreviation + pop are non-functional for settlement.
- **Agent panel (parallel):** `security-reviewer` (leak closure + `abbreviate_token` safety — confirmed; high confidence the intra-invocation child-span leak is closed, residual LangGraph sibling-node copies are out of reach but documented), `python-reviewer` (DRY: third copy of the configurable-extraction boilerplate; `or ""` already justified by the inline comment, left as-is), `tdd-guide` (async reset line is guarded; `@tool`-path caller-dict non-mutation claim is untested). All three independently flagged the docstring contradiction below.
- Re-confirmed the round-1 findings are resolved: residual leak (pop added), stale docstring (expanded), async reset test (added), field-sweep assertion (added).

**For a follow-up (out of scope here):**
- TS reverse-parity for the residual leak is already tracked as nevermined-io/nvm-monorepo#1901.
- The async `verify_permissions` / `settle_permissions` calls remain synchronous HTTP on the `_execute_with_payment_async` path (pre-existing, mirrors TS) — still worth an `asyncio.to_thread` follow-up.

PR #220, opened 2026-06-09
fix(langchain): correct PaymentContext.token docstring + DRY configurable helpers

r-marques:
## What

Post-merge follow-up to @aaitor's round-2 review on #219 — addresses the three out-of-scope items he flagged (none gated that merge).

### 1. MEDIUM — docstring contradiction (the one defect #219 introduced)

The `PaymentContext.token` `.. warning::` added in #219 told integrators:

> Code that needs the raw token under the decorator must read it from `config["configurable"]["payment_token"]`.

…but the same PR added `_remove_from_configurable(runnable_config, "payment_token")`, which **pops that exact key** before the tool body runs. A developer following that guidance inside a `requires_payment`-decorated tool body always gets `None`. All three of aaitor's review agents flagged it independently.

Reworded: under the decorator the raw token is intentionally **removed** from `config["configurable"]` (so it can't leak into nested spans) and is therefore *not* retrievable inside a decorated tool body; only the FastAPI/Starlette path leaves the raw token in this field.

### 2. MEDIUM — DRY the `configurable` helpers

`_extract_payment_token`, `_store_in_configurable` and `_remove_from_configurable` carried three byte-identical copies of the dict-vs-attr `configurable` unwrap (plus a dead `configurable = None` init overwritten on every branch). Collapsed into a single `_get_configurable` resolver; the three helpers are now one-liners routing through it. Pure refactor, behaviour-preserving.

### 3. LOW — pin the load-bearing safety claim

The pop is only safe because LangChain's `@tool` hands the decorator a per-invocation config copy (asserted in a comment, never tested). Added `test_tool_path_does_not_mutate_callers_config`: a caller reusing one config across sibling tools must keep `payment_token` — fails loudly if a future LangChain stops copying, instead of silently turning the next sibling call into a `PaymentRequiredError`. Also added a guard-branch test for the new resolver (None / non-dict configurable).

> The async-twin of the pop test aaitor mentioned is deliberately omitted: the pop lives in `_verify_payment`, the single function both the sync and async entry points call, and that shared path is already exercised. Adding a redundant async copy would not increase coverage of the pop.

## Out of scope

The async `verify_permissions`/`settle_permissions` being synchronous HTTP (worth an `asyncio.to_thread` follow-up) is pre-existing, mirrors TS, and unrelated to this change. TS reverse-parity for the residual leak shipped as nevermined-io/payments#376.

## Tests

`black --check` clean; full unit suite green (**563 passed**).

### DIFF
--- a/tests/unit/x402/test_langchain_decorator.py
+++ b/tests/unit/x402/test_langchain_decorator.py
@@ -149,6 +149,41 @@ class TestRequiresPaymentDecorator:
             == long_token
         )
 
+    def test_tool_path_does_not_mutate_callers_config(self, mock_payments):
+        """The ``@tool`` path must NOT strip payment_token from the caller's dict.
+
+        The pop in ``_verify_payment`` is only safe because LangChain's ``@tool``
+        hands the decorator a per-invocation config copy (see the comment at the
+        pop site). This pins that assumption: a caller reusing one config object
+        across sibling tools must keep ``payment_token`` for the second call. If a
+        future LangChain version stops copying, this fails loudly instead of
+        silently turning the next sibling call into a ``PaymentRequiredError``.
+        """
+        my_tool = _make_protected_tool(mock_payments)  # wrapped with @tool
+        caller_config = {"configurable": {"payment_token": "tok-original"}}
+
+        my_tool.invoke({"topic": "x"}, config=caller_config)
+
+        assert caller_config["configurable"].get("payment_token") == "tok-original", (
+            "LangChain @tool no longer copies config per invocation — "
+            "_remove_from_configurable now mutates the caller's dict"
+        )
+
+    def test_configurable_helpers_tolerate_missing_configurable(self):
+        """``_get_configurable`` and the remove/store helpers no-op on a None or
+        malformed config instead of raising (the unexercised guard branches)."""
+        # None config / non-dict configurable resolve to None.
+        assert decorator_module._get_configurable(None) is None
+        assert (
+            decorator_module._get_configurable({"configurable": "not-a-dict"}) is None
+        )
+        assert decorator_module._get_configurable({}) is None
+
+        # The mutators are silent no-ops when there's nowhere to write.
+        decorator_module._remove_from_configurable(None, "payment_token")
+        decorator_module._store_in_configurable(None, "k", "v")
+        assert decorator_module._extract_payment_token(None) is None
+
 
 class TestLastSettlement:
     def test_returns_none_before_any_settlement(self):
--- a/tests/unit/x402/test_langchain_decorator.py
+++ b/tests/unit/x402/test_langchain_decorator.py
@@ -164,14 +164,23 @@ class TestRequiresPaymentDecorator:
 
         my_tool.invoke({"topic": "x"}, config=caller_config)
 
+        # The caller's dict is intact — payment_token survives for the next sibling.
         assert caller_config["configurable"].get("payment_token") == "tok-original", (
             "LangChain @tool no longer copies config per invocation — "
             "_remove_from_configurable now mutates the caller's dict"
         )
+        # ...and the decorator demonstrably worked on a COPY: the payment_context it
+        # writes post-verify is absent from the caller's dict. Without this, the test
+        # would also pass if the pop became dead code (nothing would touch the dict).
+        assert "payment_context" not in caller_config["configurable"]
+        # Liveness: the payment flow actually executed (no swallowed early-out).
+        mock_payments.facilitator.verify_permissions.assert_called_once()
 
     def test_configurable_helpers_tolerate_missing_configurable(self):
         """``_get_configurable`` and the remove/store helpers no-op on a None or
         malformed config instead of raising (the unexercised guard branches)."""
+        from types import SimpleNamespace
+
         # None config / non-dict configurable resolve to None.
         assert decorator_module._get_configurable(None) is None
         assert (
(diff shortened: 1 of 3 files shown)


tests/unit/x402/test_langchain_decorator.py
```diff
@@ -149,6 +149,41 @@ def my_tool(topic: str, config: RunnableConfig = None) -> str:
...
+        """
+        my_tool = _make_protected_tool(mock_payments)  # wrapped with @tool
+        caller_config = {"configurable": {"payment_token": "tok-original"}}
+
+        my_tool.invoke({"topic": "x"}, config=caller_config)
+
+        assert caller_config["configurable"].get("payment_token") == "tok-original", (
+            "LangChain @tool no longer copies config per invocation — "
+            "_remove_from_configurable now mutates the caller's dict"
+        )
```
aaitor, line 171: **LOW — strengthen so the test can't pass vacuously.**

As written, `assert caller_config["configurable"].get("payment_token") == "tok-original"` is satisfied by *two* worlds: (a) the pop ran on a copy [intended], or (b) the pop never ran at all. The sibling `test_no_raw_token_survives_in_configurable` already pins that the pop runs, so this test *is* correct for its stated purpose — but a one-line positive assertion makes it prove the decorator operated on a **copy** directly (the `payment_context` it writes post-verify lands in the copy, never the caller's dict), plus a liveness check that the flow actually ran:

```suggestion
        my_tool.invoke({"topic": "x"}, config=caller_config)

        # The caller's dict is intact — payment_token survives for the next sibling.
        assert caller_config["configurable"].get("payment_token") == "tok-original", (
            "LangChain @tool no longer copies config per invocation — "
            "_remove_from_configurable now mutates the caller's dict"
        )
        # ...and the decorator demonstrably worked on a COPY: the payment_context it
        # writes post-verify is absent from the caller's dict. Without this, the test
        # would also pass if the pop became dead code (nothing would touch the dict).
        assert "payment_context" not in caller_config["configurable"]
        # Liveness: the payment flow actually executed (no swallowed early-out).
        mock_payments.facilitator.verify_permissions.assert_called_once()
```

tests/unit/x402/test_langchain_decorator.py
```diff
@@ -149,6 +149,41 @@ def my_tool(topic: str, config: RunnableConfig = None) -> str:
...
+
+    def test_configurable_helpers_tolerate_missing_configurable(self):
```
aaitor, line 179: **LOW — guard-branch coverage misses the attribute-style config path and the consumer guards.**

`_get_configurable` has a real second branch — `getattr(config, "configurable", None)` for attribute-style `RunnableConfig` objects — that this test never exercises (only dict configs are passed). And `_extract_payment_token` / the mutators have their own `is not None` guards that are only tested for `config=None`, not for a malformed `configurable`. Extending the test pins both:

```python
        from types import SimpleNamespace

        # Attribute-style config (the getattr branch), valid + malformed + missing.
        assert decorator_module._get_configurable(
            SimpleNamespace(configurable={"payment_token": "tok"})
        ) == {"payment_token": "tok"}
        assert decorator_module._get_configurable(SimpleNamespace(configurable="x")) is None
        assert decorator_module._get_configurable(SimpleNamespace()) is None

        # Consumer guards on a malformed (non-dict) configurable must also no-op.
        assert decorator_module._extract_payment_token({"configurable": "x"}) is None
        decorator_module._remove_from_configurable({"configurable": "x"}, "payment_token")
        decorator_module._store_in_configurable({"configurable": "x"}, "k", "v")
```

Optional, and worth noting separately: there's no async twin of the copy-pin test above — LangChain's `ainvoke` path copies config through different machinery than `invoke`, so it's a (small) distinct gap from the pop's correctness, which the shared-`_verify_payment` argument does cover.

aaitor (approved):
Clean follow-up that implements all three suggestions from the #219 round-2 review, correctly. I verified each:

- **Docstring (#219 MEDIUM):** the `PaymentContext.token` warning (`types.py:297-302`) no longer points integrators at `config["configurable"]["payment_token"]` — it now states the decorator *removes* that key before the tool body and only the FastAPI/Starlette path leaves the raw token in the field. Reads cleanly and matches the code.
- **DRY (#219 MEDIUM):** `_get_configurable` is the single resolver; the three helpers route through it. I walked every input shape (`None`, dict-with-dict-configurable, dict-with-non-dict-configurable, dict-with-no-configurable-key, attribute-style object with/without/non-dict `.configurable`, and non-dict/non-None config) — **behavior-equivalent to the three original inline blocks in all cases**, and it drops the dead `configurable = None` init.
- **Test (#219 LOW):** `test_tool_path_does_not_mutate_callers_config` pins the load-bearing "@tool copies the config per invocation" assumption; combined with the existing `test_no_raw_token_survives_in_configurable` (which proves the pop runs), it fails loudly if LangChain stops copying. Guard-branch test added for the new resolver. The omission of an async-twin pop test is reasonably justified (the pop lives in the shared `_verify_payment`).

No HIGH issues; the refactor is a faithful behavior-preserving change and the docstring is now correct. Two small test-strengthening nits below (neither gates merge). Approving.

**Verification I ran**
- `gh pr view/diff/checks` on `fix/langchain-decorator-followup-219` (base `main`, no chain). CI: lint + unit_integration green at `0d09e7b`; e2e pending (staging-dependent, expected per repo convention).
- Read all three changed files post-PR (`?ref=<head>`): confirmed `_get_configurable` equivalence, the reworded docstring, the `decorator_module` import alias (`test L16`), and that `_make_protected_tool` (`test L50`) genuinely `@tool`-wraps so the copy-pin test exercises the real `@tool` path.
- **Agent panel (parallel):** `python-reviewer` (refactor-equivalence — confirmed equivalent across all input shapes; one LOW: the attribute-style `getattr` branch of `_get_configurable` has no *direct* unit test, only indirect coverage), `tdd-guide` (test quality — flagged that the copy-pin test can't, by itself, distinguish "pop ran on a copy" from "pop never ran", and the guard-branch test covers only the `None`/dict cases). I down-rated tdd-guide's HIGH to a strengthening nit: the sibling `test_no_raw_token_survives_in_configurable` already pins that the pop runs, so the new test is correct for its stated purpose.

Both findings below are test polish on a PR that is *itself* net-improving coverage — fine to land as-is or fold in.

r-marques:
Thanks @aaitor — both nits folded in (d566716, test-only).

- **Copy-pin test non-vacuous:** `test_tool_path_does_not_mutate_callers_config` now also asserts `payment_context` is absent from the caller's dict (proving the decorator worked on a **copy**, so it can't pass if the pop became dead code) plus a `verify_permissions` liveness check.
- **Guard-branch coverage:** extended to exercise `_get_configurable`'s attribute-style `getattr` branch (valid / malformed / missing via `SimpleNamespace`) and the consumer guards (`_extract_payment_token` / `_remove_from_configurable` / `_store_in_configurable`) on a malformed non-dict `configurable`.
- **Async twin:** added `test_tool_path_does_not_mutate_callers_config_async` — `ainvoke` copies config through different machinery than `invoke`, so I agree it's a distinct path worth pinning (not covered by the shared-`_verify_payment` argument).

black + full unit suite green (564 passed).

PR #222, opened 2026-06-09
feat(langsmith): build_payment_app optional on langgraph-api >= 0.6.15

r-marques:
## What

Resolves the final two Sprint-3 follow-ups of epic nevermined-io/nvm-monorepo#1703: the empirical spike **nvm-monorepo#1762** and its gated code change **nvm-monorepo#1763**.

## Spike result (#1762)

**A plain Starlette `http.app` boots cleanly on `langgraph-api` 0.8.7.** Verified end-to-end against a real install (`langgraph dev` with a plain-Starlette `http.app`):

| Probe | Result |
|---|---|
| `GET /ok` | 200 |
| `GET /openapi.json` | 200 (spec generated) |
| ungated `POST /threads` | 200 (passes through) |
| gated `POST /threads/{id}/runs/wait` (no `payment_signature`) | **402** (PaymentMiddleware composed + enforced) |

The internal endpoint docstrings with YAML-unsafe colons (e.g. `/assistants/{assistant_id}` "Query params:") **still exist**, and the `SchemaGenerator` path is **still taken** for non-FastAPI apps — but they no longer crash startup. `langgraph_api/utils/__init__.py::SchemaGenerator.get_schema` broadened its catch around `parse_docstring` from `except AssertionError` to `except Exception` (the bad docstring degrades to a plain description + a logged warning).

**Bisected the fix to `langgraph-api 0.6.15`** (downloaded each wheel, inspected `get_schema`): `0.6.14` crashes, `0.6.15` boots. So the FastAPI wrapper is no longer load-bearing on `>= 0.6.15` — it's now a convenience.

## Changes

- **`build_payment_app` docstring** rewritten: "not required on `langgraph-api >= 0.6.15`", with the plain-Starlette form shown inline.
- **DeprecationWarning (#1763):** `build_payment_app` now detects the installed `langgraph-api` at runtime and, when `>= 0.6.15`, emits a `DeprecationWarning` pointing users at mounting `PaymentMiddleware` directly on a Starlette `http.app`. On `< 0.6.15` (wrapper still required) it's silent.
- **`_langgraph_api_version()`** — fully defensive: returns `None` (silent, no warning, app still built) if `langgraph-api` is absent or the version string is exotic. Construction never breaks.

## Tests

+7: warning fires at/above 0.6.15, silent below it and when undetected; version parser handles 3-component, pre-release suffix (`0.6.15rc1` → `(0, 6, 15)`), and absent package. `black --check` clean; full unit suite green (**571 passed**). Also verified end-to-end: the modified factory against a real `langgraph-api 0.8.7` fires the warning with the detected version.

## Notes

- No upstream filing needed — the "plain Starlette still crashes" branch of the spike did not occur.
- The tutorial (`langchain-langsmith-deployment-py`) keeps using `build_payment_app` (still valid); switching it to plain Starlette is optional cleanup, deferred.

> ℹ️ Linked issues live in `nvm-monorepo`, so GitHub's auto-close won't fire cross-repo on merge — they'll be closed manually.

### DIFF
--- a/payments_py/langsmith/middleware.py
+++ b/payments_py/langsmith/middleware.py
@@ -557,9 +557,13 @@ def build_payment_app(
             Middleware(PaymentMiddleware, payments=payments, routes={...}),
         ])
 
-    On ``langgraph-api >= 0.6.15`` this factory emits a ``DeprecationWarning``
-    pointing at the plain-Starlette form above. It remains useful on older
-    ``langgraph-api`` (where the FastAPI wrapper is still load-bearing). The
+    On ``langgraph-api >= 0.6.15`` this factory emits a ``UserWarning``
+    pointing at the plain-Starlette form above. (``UserWarning`` rather than
+    ``DeprecationWarning`` so the nudge is actually visible at app startup —
+    ``DeprecationWarning`` is suppressed by default outside ``__main__``/pytest;
+    and the factory is not being removed, only made optional.) It remains useful
+    on older ``langgraph-api`` (where the FastAPI wrapper is still load-bearing).
+    The
     middleware class itself (``PaymentMiddleware``) is a Starlette
     ``BaseHTTPMiddleware`` subclass and works on both Starlette and FastAPI —
     only the outer app wrapper ever mattered for the upstream bug.
@@ -597,8 +601,9 @@ def build_payment_app(
             f"langgraph-api >= {fix} (detected "
             f"{'.'.join(str(p) for p in detected)}). Mount PaymentMiddleware "
             f"directly on a plain Starlette http.app instead: "
-            f"app.add_middleware(PaymentMiddleware, payments=..., routes=...).",
-            DeprecationWarning,
+            f"Starlette(middleware=[Middleware(PaymentMiddleware, "
+            f"payments=..., routes=...)]).",
+            UserWarning,
             stacklevel=2,
         )
     app = FastAPI()
--- a/payments_py/langsmith/middleware.py
+++ b/payments_py/langsmith/middleware.py
@@ -30,13 +30,16 @@ more than payment gating.
 import asyncio
 import base64
 import contextlib
+import importlib.metadata
 import inspect
 import logging
 import posixpath
+import re
 import time
+import warnings
 from contextlib import contextmanager
 from dataclasses import dataclass
-from typing import Any, Awaitable, Callable, Dict, Iterator, Optional, Union
+from typing import Any, Awaitable, Callable, Dict, Iterator, Optional, Tuple, Union
 
 from fastapi import FastAPI
 from starlette.datastructures import MutableHeaders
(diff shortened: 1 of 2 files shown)


payments_py/langsmith/middleware.py
```diff
@@ -545,6 +589,18 @@ def build_payment_app(
...
+    detected = _langgraph_api_version()
+    if detected is not None and detected >= _LANGGRAPH_API_DOCSTRING_FIX:
+        fix = ".".join(str(p) for p in _LANGGRAPH_API_DOCSTRING_FIX)
+        warnings.warn(
+            f"build_payment_app's FastAPI wrapper is no longer required on "
+            f"langgraph-api >= {fix} (detected "
+            f"{'.'.join(str(p) for p in detected)}). Mount PaymentMiddleware "
+            f"directly on a plain Starlette http.app instead: "
+            f"app.add_middleware(PaymentMiddleware, payments=..., routes=...).",
+            DeprecationWarning,
```
aaitor: **MEDIUM — `DeprecationWarning` is silenced by default outside `__main__`/test runners, so production callers likely never see this nudge.**

Since Python 3.2 the default filter only surfaces `DeprecationWarning` in `__main__` and under pytest. `build_payment_app` runs at app-import/startup inside a module — exactly the context where `DeprecationWarning` is hidden. So a user on `langgraph-api >= 0.6.15` (the whole target audience of #1763) won't see the message in a normal deployment; only your own test suite does. If the goal is to actually steer users toward the plain-Starlette form, `FutureWarning` (shown by default everywhere) reaches them; `UserWarning` is the most honest category since `build_payment_app` isn't being *removed* (it stays valid on older langgraph-api).

```suggestion
        DeprecationWarning,  # NOTE: hidden by default outside __main__/pytest; use FutureWarning if you want production callers to actually see it
```

If low-noise was deliberate (rely on CI/pytest to surface it, don't spam prod logs), that's a defensible call — just flagging that the warning won't reach the audience the issue targets. Your decision.

payments_py/langsmith/middleware.py
```diff
@@ -545,6 +589,18 @@ def build_payment_app(
         # langgraph.json
         # { "http": { "app": "./nvm_app.py:app" } }
     """
+    detected = _langgraph_api_version()
+    if detected is not None and detected >= _LANGGRAPH_API_DOCSTRING_FIX:
```
aaitor, line 597: **LOW — the version comparison is correct, but the subtle part (short tuples) isn't pinned by a test, so a well-meaning refactor could silently break it.**

`detected >= (0, 6, 15)` relies on Python's tuple-prefix ordering, which happens to match semver against this non-zero-patch sentinel: `(0,7) >= (0,6,15)` → True, `(0,6) >= (0,6,15)` → False (0.6.0 < 0.6.15), `(1,) >= (0,6,15)` → True — all correct. **No change needed to the code.** The risk is that someone later "hardens" this (e.g. `zip_longest`-pads or adds a `len(detected) == 3` guard) and breaks the `"0.7"` case. Pin it + the untested parser fallback branches so that can't happen silently:

```python
    @pytest.mark.parametrize("detected,should_warn", [
        ((0, 6, 15), True), ((0, 8, 7), True),
        ((0, 7), True),     # 2-component, above fix
        ((0, 6), False),    # 2-component, below fix (0.6.0 < 0.6.15)
        ((1,), True), ((0,), False),
        ((0, 6, 14), False),
    ])
    def test_warning_gate_across_tuple_shapes(self, mock_payments, monkeypatch, recwarn, detected, should_warn):
        monkeypatch.setattr(ls_middleware, "_langgraph_api_version", lambda: detected)
        if should_warn:
            with pytest.warns(DeprecationWarning, match="no longer required"):
                build_payment_app(payments=mock_payments, routes=None)
        else:
            build_payment_app(payments=mock_payments, routes=None)
            assert not [w for w in recwarn.list if issubclass(w.category, DeprecationWarning)]

    def test_parses_non_digit_and_extra_components(self, monkeypatch):
        monkeypatch.setattr("importlib.metadata.version", lambda name: "0.6.rc1")   # else-0 branch
        assert ls_middleware._langgraph_api_version() == (0, 6, 0)
        monkeypatch.setattr("importlib.metadata.version", lambda name: "0.6.15.post1")  # [:3] slice
        assert ls_middleware._langgraph_api_version() == (0, 6, 15)
```

Two more cheap robustness tweaks: add one end-to-end test that patches only `importlib.metadata.version` and calls `build_payment_app` (the current warning tests stub `_langgraph_api_version` wholesale, so the real parse→compare→warn seam is never exercised together); and scope the two `recwarn` "no warning" assertions with `and "no longer required" in str(w.message)` so an unrelated `DeprecationWarning` from FastAPI/Starlette can't make them spuriously pass or fail.

payments_py/langsmith/middleware.py
```diff
@@ -545,6 +589,18 @@ def build_payment_app(
...
+        fix = ".".join(str(p) for p in _LANGGRAPH_API_DOCSTRING_FIX)
+        warnings.warn(
+            f"build_payment_app's FastAPI wrapper is no longer required on "
+            f"langgraph-api >= {fix} (detected "
+            f"{'.'.join(str(p) for p in detected)}). Mount PaymentMiddleware "
+            f"directly on a plain Starlette http.app instead: "
+            f"app.add_middleware(PaymentMiddleware, payments=..., routes=...).",
```
aaitor: **LOW — the warning text and the docstring example suggest two different Starlette idioms; align them.**

The docstring (L556-558) shows the constructor form `Starlette(middleware=[Middleware(PaymentMiddleware, ...)])`, while this warning string says `app.add_middleware(PaymentMiddleware, ...)`. Both are valid and behave identically here (a single middleware — ordering is moot), but a user following one path writes different code than one following the other. Match the warning to the docstring's constructor form for one consistent migration story:

```suggestion
            f"directly on a plain Starlette http.app instead: "
            f"Starlette(middleware=[Middleware(PaymentMiddleware, payments=..., routes=...)]).",
```

aaitor (approved):
Clean, well-scoped Sprint-3 wrap-up: a docstring rewrite plus a soft-deprecation nudge on `build_payment_app` when `langgraph-api >= 0.6.15`, backed by a fully defensive version parser that can never break app construction (returns `None` → silent). Backward-compatible (the factory still returns a FastAPI app in every case), no security surface, CI green. The version logic is correct — I verified the tuple comparison against the 3-tuple sentinel handles short/long/prerelease versions right.

No HIGH issues. Both review agents flagged the 2-component tuple comparison as HIGH, but that's a false positive: I traced it and the comparison is correct in every case (Python's prefix-comparison matches semver's "missing components default to 0" against a non-zero-patch sentinel — zero-padding would change nothing). The real takeaway is a test gap, not a code bug. Approving; one MEDIUM judgment call and two LOWs below, none gating.

**Verification I ran**
- `gh pr view/diff/checks` on `fix/build-payment-app-langgraph-0615-1762-1763` (base `main`, no chain). CI: 3/3 green at `12edcdf`. Read the full post-PR `middleware.py` (L503-606).
- Hand-checked `_langgraph_api_version` against PEP 440 oddities: leading-digit extraction handles `"0.6.15rc1"`→`(0,6,15)`, local `+build`, `.post1`/`.dev3` (sliced by `[:3]`), and components with no leading digit degrade to `0`; the only throw path (`PackageNotFoundError`) is caught → `None`. Construction never breaks. ✓
- Verified the `detected >= (0,6,15)` comparison is semantically correct for 1-, 2-, and 3-element tuples (the agents' "HIGH" zero-pad fix is a no-op against this sentinel).
- **Agent panel (parallel):** `python-reviewer` (version logic + warning semantics — surfaced the `DeprecationWarning` visibility point; its tuple "HIGH" I verified as a non-bug), `tdd-guide` (coverage — the comparison/parser edge branches aren't pinned by tests). Dropped: the zero-pad code change (no-op), the "different middleware execution order" framing (irrelevant for a single middleware).

r-marques:
Thanks @aaitor — all three folded in (ff95d13).

- **MEDIUM (warning invisible in prod):** switched `DeprecationWarning` → **`UserWarning`**. You're right that `DeprecationWarning` is suppressed by default outside `__main__`/pytest, so the #1763 audience would never see the nudge at app startup. `UserWarning` is visible by default and is the honest category (the factory stays valid on older langgraph-api — it's not being removed). Confirmed end-to-end against a real `langgraph-api 0.8.7` install under the default warning filter: the message now actually surfaces. (Picked `UserWarning` over `FutureWarning` since nothing's changing behaviour, just becoming optional.)
- **LOW (tuple-shape compare untested):** added a parametrized matrix pinning `(0,7)→warn`, `(0,6)→no`, `(1,)→warn`, `(0,)→no`, `(0,6,14)→no`, etc., plus parser tests for the non-digit (`0.6.rc1`→`(0,6,0)`) and 4th-component (`0.6.15.post1`→`(0,6,15)`) branches, and an end-to-end test that patches only `importlib.metadata.version` so the real parse→compare→warn seam runs together. The "no warning" assertions are now scoped to our message.
- **LOW (text mismatch):** the warning string now matches the docstring's constructor form — both show `Starlette(middleware=[Middleware(PaymentMiddleware, ...)])`.

`black` clean; full unit suite **579 passed**.

PR #225, opened 2026-06-11
feat(x402): require currency, add delegation planId, deprecate inline create-on-the-fly

### LINKED ISSUE #224: x402 SDK (Python): add delegation planId, require currency, deprecate inline create-on-the-fly token requests
Part of epic nevermined-io/nvm-monorepo#1549 (x402 hardening). Python SDK (`payments-py`) side of the backend decisions in nvm-monorepo#1534, nvm-monorepo#1677, and nvm-monorepo#1674. Parity with the TS SDK issue nevermined-io/payments#379.

**Changes:**
1. Add an optional `planId` to the delegation config / create-delegation payload (additive; use the repo's snake_case convention). Delegations are plan-agnostic by default; the planId opts into a plan-bound delegation (Visa still requires it). — nvm-monorepo#1534
2. Make `currency` **required** in the create-delegation payload — the backend now requires it (no more silent usd/usdc default). Coordinated breaking release. — nvm-monorepo#1677
3. The token-request call (parity with the TS `getX402AccessToken`): emit a runtime **deprecation warning** when called with an inline-create delegation config (spending-limits / no delegation id). Supported flow: create the delegation first, then request the token referencing the delegation id. — nvm-monorepo#1674
4. Update the SDK's **own reference docs in this repo** to the create-first pattern; retire inline-create / Auto-Selection examples. (Narrative/guide docs live in `nevermined-io/docs` — tracked separately.)

Coordinate the release with `payments` (TS). Do **not** manually bump the pyproject version — the release pipeline owns it. Decision detail: see the epic notes (`~/code/notes/x402-hardening-epic-1549.md`, 2026-06-10).

r-marques:
## Description

Python SDK (`payments-py`) side of the merged x402 hardening delegation contract
(`nevermined-io/nvm-monorepo#1674` / `#1677` / `#1534`), in parity with the TS SDK
(`nevermined-io/payments#379`). Mirrors the backend `CreateDelegationDto`.

**Changes (`payments_py/x402/types.py`, `token.py`):**

1. **`currency` is now REQUIRED** on `CreateDelegationPayload` and typed as a
   `DelegationCurrency = Literal["usd", "eur", "usdc", "eurc"]` enum. The backend no
   longer applies a silent `usd`/`usdc` default (`#1677`), so the SDK now surfaces the
   requirement at construction time. **Coordinated breaking change** — callers must pass
   `currency` explicitly.
2. **Add optional `plan_id` to `DelegationConfig`** (the token-request delegation config).
   Additive and plan-agnostic by default (`#1534`); set it to bind the delegation to a
   single plan. (`CreateDelegationPayload.plan_id` already existed.)
3. **Deprecation warning for inline create-on-the-fly.** `get_x402_access_token` now emits
   a `DeprecationWarning` when called with a `DelegationConfig` that has no `delegation_id`
   but carries an inline-create signal (`card_id` / `provider_payment_method_id` /
   `spending_limit_cents` / `duration_secs`). The supported flow is **create-first**:
   create the delegation via `payments.delegation.create_delegation(...)`, then request the
   token referencing only `DelegationConfig(delegation_id=...)`. The request still
   succeeds — the path is deprecated, not removed (parity with the backend, which keeps the
   inline path working and logs its own server-side deprecation warning, `#1674`).

**Docs:** updated the SDK's own hand-written `docs/api/` examples (07, 11, 12) to the
create-first pattern with required `currency`, plus the model/method docstrings (the
auto-generated `docs/reference/` is sourced from those docstrings). `mkdocs build` is green.
Central narrative docs are tracked separately in `nevermined-io/nvm-monorepo#1913`.

**Release:** no `pyproject.toml` version bump — the release pipeline owns it.

Related to Issue #224

Part of `nevermined-io/nvm-monorepo#1549`. Closes #224.

## Types of changes

- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] Tests Cover Changes
- [x] Documentation

#### Funny gif

currency required now

### DIFF
--- a/payments_py/x402/token.py
+++ b/payments_py/x402/token.py
@@ -108,8 +108,8 @@ class X402TokenAPI(BasePaymentsAPI):
         Passing spending limits / a payment method here instead (inline
         create-on-the-fly, i.e. a ``delegation_config`` with no
         ``delegation_id``) is **deprecated** (#1674): this method emits a
-        ``DeprecationWarning`` and the backend logs its own deprecation
-        warning. The inline path will be removed in a future release.
+        ``FutureWarning`` and the backend logs its own deprecation warning.
+        The inline path will be removed in a future release.
 
         Args:
             plan_id: The unique identifier of the payment plan
@@ -121,7 +121,9 @@ class X402TokenAPI(BasePaymentsAPI):
                 - accessToken: The X402 access token string
 
         Raises:
-            PaymentsError: If the request fails
+            PaymentsError: If the request fails, or (``code='validation'``) if
+                ``delegation_config.delegation_id`` is an empty string — pass a
+                valid delegation UUID or omit the field.
 
         Example:
             ```python
@@ -176,7 +178,23 @@ class X402TokenAPI(BasePaymentsAPI):
         # Add delegation config for both erc4337 and card-delegation schemes
         if token_options and token_options.delegation_config:
             delegation_config = token_options.delegation_config
+            # An empty-string delegation_id is neither a valid reuse (it's not a
+            # UUID) nor "absent": model_dump(exclude_none=True) keeps "" (empty
+            # is not None), so it would serialize `delegationId: ""` and 4xx at
+            # the backend, while _is_inline_create would (mis)read it as inline.
+            # Fail fast with a clear client-input error instead (symmetric with
+            # the TS SDK guard, payments#379).
+            if delegation_config.delegation_id == "":
+                raise PaymentsError.validation(
+                    "delegation_id must not be an empty string — pass a valid "
+                    "delegation UUID or omit the field."
+                )
             if _is_inline_create(delegation_config):
+                # FutureWarning (not DeprecationWarning): DeprecationWarning is
+                # filtered out by default outside __main__, so agents running
+                # under FastAPI / gunicorn / Celery / Docker workers would never
+                # see the nudge. FutureWarning is shown by default → true runtime
+                # parity with the TS SDK's console.warn.
                 warnings.warn(
                     "Passing spending limits / a payment method to "
                     "get_x402_access_token (inline delegation create-on-the-fly) "
@@ -184,7 +202,7 @@ class X402TokenAPI(BasePaymentsAPI):
                     "Create the delegation first with "
                     "payments.delegation.create_delegation(...) and pass only "
                     "DelegationConfig(delegation_id=...) here.",
-                    DeprecationWarning,
+                    FutureWarning,
                     stacklevel=2,
                 )
             body["delegationConfig"] = delegation_config.model_dump(
--- a/payments_py/x402/token.py
+++ b/payments_py/x402/token.py
@@ -178,13 +178,17 @@ class X402TokenAPI(BasePaymentsAPI):
         # Add delegation config for both erc4337 and card-delegation schemes
         if token_options and token_options.delegation_config:
             delegation_config = token_options.delegation_config
-            # An empty-string delegation_id is neither a valid reuse (it's not a
-            # UUID) nor "absent": model_dump(exclude_none=True) keeps "" (empty
-            # is not None), so it would serialize `delegationId: ""` and 4xx at
-            # the backend, while _is_inline_create would (mis)read it as inline.
-            # Fail fast with a clear client-input error instead (symmetric with
-            # the TS SDK guard, payments#379).
-            if delegation_config.delegation_id == "":
+            # An empty- or whitespace-only delegation_id is neither a valid
+            # reuse (it's not a UUID) nor "absent": model_dump(exclude_none=True)
+            # keeps it (it is not None), so it would serialize a blank
+            # `delegationId` and 4xx at the backend, while _is_inline_create
+            # would (mis)read it as inline. Fail fast with a clear client-input
+            # error instead. Strip first to match the TS SDK guard's
+            # `.trim() === ''` (payments#379) — exact cross-SDK symmetry.
+            if (
+                delegation_config.delegation_id is not None
+                and delegation_config.delegation_id.strip() == ""
+            ):
                 raise PaymentsError.validation(
                     "delegation_id must not be an empty string — pass a valid "
                     "delegation UUID or omit the field."
--- a/payments_py/x402/token.py
+++ b/payments_py/x402/token.py
@@ -7,6 +7,7 @@ Tokens are used to authorize payment verification and settlement.
 
 import base64
 import json
+import warnings
 import requests
 from typing import Dict, Any, Optional
 from payments_py.common.payments_error import PaymentsError
@@ -14,7 +15,31 @@ from payments_py.common.types import PaymentOptions
 from payments_py.api.base_payments import BasePaymentsAPI
 from payments_py.api.nvm_api import API_URL_CREATE_PERMISSION
 from payments_py.x402.schemes import get_default_network
-from payments_py.x402.types import X402TokenOptions
+from payments_py.x402.types import DelegationConfig, X402TokenOptions
+
+
+def _is_inline_create(delegation_config: DelegationConfig) -> bool:
+    """Whether a delegation config asks the backend to create a delegation
+    on the fly (the deprecated path) rather than reusing an existing one.
+
+    The supported flow is create-first: create the delegation via
+    ``POST /delegation/create`` and pass only ``delegation_id`` here. A config
+    that carries no ``delegation_id`` but does carry an inline-create signal
+    (a payment-method reference or spending limits) triggers the backend's
+    deprecated create-on-the-fly path, which logs its own deprecation warning
+    server-side (#1674) and will be removed in a future release.
+    """
+    if delegation_config.delegation_id:
+        return False
+    return any(
+        value is not None
+        for value in (
+            delegation_config.card_id,
+            delegation_config.provider_payment_method_id,
+            delegation_config.spending_limit_cents,
+            delegation_config.duration_secs,
+        )
+    )
 
 
 def decode_access_token(access_token: str) -> Optional[Dict[str, Any]]:
@@ -71,14 +96,20 @@ class X402TokenAPI(BasePaymentsAPI):
         token_options: Optional[X402TokenOptions] = None,
     ) -> Dict[str, Any]:
         """
-        Create a delegation and get an X402 access token for the given plan.
+        Get an X402 access token for the given plan against a delegation.
 
         This token allows the agent to verify and settle delegations on behalf
         of the subscriber.
 
-        For erc4337 scheme, you must pass ``token_options.delegation_config`` with either:
-        - ``delegation_id`` to reuse an existing delegation, or
-        - ``spending_limit_cents`` + ``duration_secs`` to auto-create a new one.
+        Supported flow (**create-first**): create the delegation once via
+        ``payments.delegation.create_delegation(...)`` and pass only its
+        ``delegation_id`` here via ``token_options.delegation_config``.
+
+        Passing spending limits / a payment method here instead (inline
+        create-on-the-fly, i.e. a ``delegation_config`` with no
+        ``delegation_id``) is **deprecated** (#1674): this method emits a
+        ``DeprecationWarning`` and the backend logs its own deprecation
+        warning. The inline path will be removed in a future release.
 
         Args:
             plan_id: The unique identifier of the payment plan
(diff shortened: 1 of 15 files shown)


r-marques (commented):
## 🤖 Automated PR review — 🟡 Mergeable with minor test-hardening (no blockers)

_Reviewed `origin/main...HEAD` (15 files) with the pr-review-toolkit panel: code-reviewer, type-design-analyzer, pr-test-analyzer. SDK delegation/x402 contract change._

**Headline:** Strong, correct mirror of the merged backend contract. **No blockers.** The deprecation predicate `_is_inline_create` is correct across every edge case probed (warns on `card_id`/`provider_payment_method_id`/spending-limits when no `delegation_id`; silent when `delegation_id` is set, even with leftover inline fields), the warning is non-fatal (request still proceeds; `stacklevel=2` attributes it to the caller), the `currency`-required breaking change is **complete** (every `CreateDelegationPayload(...)` site updated; a missed one would now hard-fail), and the `DelegationCurrency`/`DelegationProvider` Literals were verified to match the **merged** backend DTO (`#1677`) exactly — no drift. CLAUDE.md-clean (black, no pyproject bump, docstrings updated not the generated `docs/reference/`).

### 🔴 Blockers (0)
None.

### 🟡 Should fix (test-hardening — cheap, pins real contract edges)
- **Migration-footgun test: a real `delegation_id` + leftover inline fields must stay SILENT** — `tests/unit/x402/test_token.py`. The predicate uses a truthiness check (`if delegation_config.delegation_id:`), and the realistic case — a caller who switched to create-first but left `spending_limit_cents` in their config — should **not** be warned. Add a test: `DelegationConfig(delegation_id="deleg-123", spending_limit_cents=10000)` under `simplefilter("error")` → no warning. Without it, a refactor to `delegation_id is not None` would silently flip the contract untested. _(pr-test-analyzer S1)_
- **Pin the `{delegation_id, api_key_id}` no-warn path** — the api_key_id-scoped reuse is a documented usage pattern; only `{delegation_id}` alone is currently pinned. Add `DelegationConfig(delegation_id="deleg-123", api_key_id="key-1")` → silent. _(pr-test-analyzer S2)_

### 💡 Good to have
- **Assert `currency` reaches the wire** — add `assert body["delegationConfig"]["currency"] == "usd"` to `test_inline_create_via_payment_method_warns` (one line; currency is the headline change and this pins it survives serialization on the still-consumed inline path). _(pr-test-analyzer G2)_
- **Stale doc comment** — `docs/api/07-querying-an-agent.md:24` still reads `# Generate access token (basic — auto-creates delegation)`, the now-deprecated framing (the code itself is harmless — no `delegation_config`, so no warning). Reword to e.g. `# no delegation — credits/fiat plans`. _(code-reviewer + type-design)_
- Align the `match=` string in `test_inline_create_via_spending_limits_warns` to `"create_delegation"` (sibling tests use that; pins the actionable remediation text) _(G3)_; optionally a `DelegationConfig(currency="gbp")` ValidationError test _(G1)_.

### ✅ Strengths
- **Deprecation predicate correct across 10 probed cases** — exact inline-create signal set, `delegation_id` precedence, empty-string edge handled sensibly.
- **Breaking change is complete** — every real `CreateDelegationPayload` construction passes `currency`; the one negative test asserts the `ValidationError`.
- **Types verified against the *merged* backend** (`#1677` @ `f73c2060`): currency required + Literal, provider/currency value sets, and the Visa `plan_id` carve-out all match — no drift. The `DelegationCurrency` comment pins the SDK Literal to its backend source-of-truth.
- **Tests prove behavior, not happy paths** — `simplefilter("error")` for the negative no-warn cases genuinely pins silence; currency-required covered (missing → ValidationError, invalid → rejected, all 4 accepted); plan_id round-trips on both models.
- CLAUDE.md-clean: black, no version bump, generated `docs/reference/` untouched (docstrings are the lever), `mkdocs build` green.

_Verdict 🟡: no blockers — correct, complete, and well-tested. Land the two edge-case tests (real-delegation_id-with-leftover-fields + api_key_id-scoped reuse) and the doc-comment fix before flipping out of draft._

payments_py/x402/token.py
```diff
@@ -145,7 +175,19 @@ def get_x402_access_token(
...
         if token_options and token_options.delegation_config:
-            body["delegationConfig"] = token_options.delegation_config.model_dump(
+            delegation_config = token_options.delegation_config
+            if _is_inline_create(delegation_config):
+                warnings.warn(
+                    "Passing spending limits / a payment method to "
+                    "get_x402_access_token (inline delegation create-on-the-fly) "
+                    "is deprecated and will be removed in a future release. "
+                    "Create the delegation first with "
+                    "payments.delegation.create_delegation(...) and pass only "
```
aaitor, line 207: **MEDIUM — `DeprecationWarning` is silenced by default in production, so the runtime nudge won't reach most callers.**

CPython's default filters ignore `DeprecationWarning` everywhere except `__main__` (pytest re-enables it, which is exactly why the tests see it). Any agent that calls `get_x402_access_token` from a framework worker — FastAPI / gunicorn / Celery / a Docker entrypoint — runs outside `__main__`, so the warning is swallowed and the inline-create path goes silent for precisely the production users you want to migrate before removal. The TS sibling (`payments#380`) uses `console.warn`, which always prints — so the PR's "parity with the TS SDK" holds for the *predicate* but not for *runtime visibility* (TS loud, Python quiet).

This is a design call, not a bug — pick one:
- **Keep `DeprecationWarning`** if migration is driven by docs/changelog and you don't want to spam end users at runtime (the idiomatic-library choice). If so, soften the "parity" wording in the PR/commit.
- **Switch to `FutureWarning`** for true runtime parity — it is *not* filtered by default and is aimed at end-user code:

```python
warnings.warn(
    "Passing spending limits / a payment method to "
    "get_x402_access_token (inline delegation create-on-the-fly) "
    "is deprecated and will be removed in a future release. "
    "Create the delegation first with "
    "payments.delegation.create_delegation(...) and pass only "
    "DelegationConfig(delegation_id=...) here.",
    FutureWarning,
    stacklevel=2,
)
```
…and flip the three `pytest.warns(DeprecationWarning, ...)` assertions in `tests/unit/x402/test_token.py` to `FutureWarning`.

payments_py/x402/token.py
```diff
@@ -7,14 +7,39 @@
...
+    on the fly (the deprecated path) rather than reusing an existing one.
+
+    The supported flow is create-first: create the delegation via
+    ``POST /delegation/create`` and pass only ``delegation_id`` here. A config
+    that carries no ``delegation_id`` but does carry an inline-create signal
+    (a payment-method reference or spending limits) triggers the backend's
+    deprecated create-on-the-fly path, which logs its own deprecation warning
+    server-side (#1674) and will be removed in a future release.
+    """
+    if delegation_config.delegation_id:
```
aaitor, line 32: **LOW — Empty-string `delegation_id` is classified as "absent" (cross-SDK, non-blocking).**

`if delegation_config.delegation_id:` is truthy-based, so `delegation_id=""` falls through to the inline-create branch (warns), and `model_dump(by_alias=True, exclude_none=True)` still serializes `delegationId: ""` onto the wire (empty string isn't `None`), which the backend rejects. The TS sibling shares the same `!delegationId` behavior, so the two SDKs are consistent, and an empty id is ~equivalent to omission — hence `LOW`. If you decide to add an explicit empty-string guard, do it symmetrically with `payments#380`.

Note the `spending_limit_cents=0` edge just below is handled correctly — `value is not None` (not truthiness) keeps `0` as a valid inline signal. 👍

aaitor (approved):
Strong, complete Python-SDK mirror of the merged backend delegation contract — verified against the actual merged `CreateDelegationDto` on `nvm-monorepo@main`: `DelegationCurrency = Literal['usd','eur','usdc','eurc']` and `DelegationProvider` match the backend `@IsIn` sets exactly, `currency` is correctly required, and the Visa `plan_id` carve-out is accurate. The `_is_inline_create` predicate is correct across the edges I probed (including the subtle `spending_limit_cents=0` case, handled right via `is not None`), the warning is non-fatal, and every self-review should-fix was applied (migration-footgun + api_key_id-reuse no-warn tests, currency-on-the-wire assertion). **Approving** — no blockers. One `MEDIUM` design call on the warning class (a decision, not a defect) and one `LOW`, both inline.

**Verification I ran**
- `gh pr diff` + read every changed file in its post-PR state on `feat/x402-delegation-1549`; CI all green (lint, unit_integration, e2e).
- Cross-checked the contract against the **merged** backend DTO (`apps/api/src/delegation/dto/create-delegation.dto.ts` @ `main`) — currency/provider value sets and required-fields match; no drift.
- Cross-SDK predicate parity vs `payments#380`: both gate inline-create on the same 4 creation fields when `delegation_id` is absent — aligned.
- Parallel `python-reviewer` lens; dropped its over-rated/out-of-scope items after reading the code (the `currency`-alias "gap" is a runtime no-op since the wire key *is* `currency`; the `@field_validator`/`@classmethod` note is on pre-existing code outside this diff; positive-int validation is already enforced server-side by `@Min(1)`).

**Notes (non-blocking)**
- **Optional test hardening:** (1) an `api_key_id`-only config with **no** `delegation_id` should stay silent — currently only the `{delegation_id, api_key_id}` path is pinned; (2) a `spending_limit_cents=0` case to pin that `is not None` (not truthiness) keeps `0` as an inline signal — guards against a future `if value:` refactor.

payments_py/x402/types.py
```diff
@@ -338,7 +352,8 @@ class DelegationConfig(BaseModel):
     )
     spending_limit_cents: Optional[int] = Field(None, alias="spendingLimitCents")
     duration_secs: Optional[int] = Field(None, alias="durationSecs")
-    currency: Optional[str] = None
+    currency: Optional[DelegationCurrency] = None
```
eruizgar91, line 355: **Inline-create path no longer enforces the now-required `currency` — likely breaks against the real backend.**

This PR makes `currency` **required** on `CreateDelegationPayload` because the backend (#1677) dropped the silent `usd`/`usdc` default. But `DelegationConfig.currency` — the model used for the still-supported (deprecated) inline create-on-the-fly path through `get_x402_access_token` — stays `Optional[DelegationCurrency] = None`.

The asymmetry is self-documented: the docstring you added two lines below (types.py:339-340) says currency is *"Required by the backend for the deprecated inline-create path (no silent default)"*. Yet `_is_inline_create` triggers on `spending_limit_cents`/`duration_secs` alone and the request is still forwarded with no currency. `test_inline_create_via_spending_limits_warns` constructs exactly `DelegationConfig(spending_limit_cents=10000, duration_secs=604800)` with no currency and only passes because the backend is mocked.

Against the real backend an inline-create call that omits currency would be rejected (HTTP 400), so the deprecated-but-"still works" path is actually broken for callers who relied on the old default. Consider either keeping the path functional (require currency when `_is_inline_create` is true, raising a clear SDK-side error) or documenting that currency must be supplied on the inline `DelegationConfig` too.

payments_py/x402/types.py
```diff
@@ -314,18 +319,27 @@ class DelegationConfig(BaseModel):
...
+
+    To reuse an existing delegation supply ``delegation_id`` (preferred).
     To reuse an existing card (PaymentMethod entity) supply ``card_id``.
-    When creating a brand-new delegation provide ``provider_payment_method_id``,
-    ``spending_limit_cents``, and ``duration_secs``.
+    The deprecated inline path provides ``provider_payment_method_id``,
+    ``spending_limit_cents``, ``duration_secs``, and ``currency``.
 
     Attributes:
         card_id: PaymentMethod entity UUID -- preferred way to reference an enrolled card
```
eruizgar91, line 334: **Stale/contradictory docstring: `card_id` is described as "preferred" but now triggers a `DeprecationWarning`.**

This line still calls `card_id` the *"preferred way to reference an enrolled card"*, but with this PR `_is_inline_create` (token.py:34-42) treats a `card_id` with no `delegation_id` as the **deprecated** inline create-on-the-fly path — `test_inline_create_via_card_id_warns` asserts exactly that `DelegationConfig(card_id=...)` emits a `DeprecationWarning`.

So the docstring steers callers onto the path the code now warns against, and it contradicts the `delegation_id` line just below which (correctly) gets the new `(preferred)` marker. Please reword this to reflect that `delegation_id` (create-first) is the preferred path and `card_id` falls on the deprecated inline-create path.

PR #228, opened 2026-06-16
feat(mcp): align x402 transport with v2 in-band MCP spec (#203)

### LINKED ISSUE #203: Align MCP transport with x402 v2 spec (in-band signaling via tool-result + _meta)
## Summary

Our MCP transport for x402 in `payments_py/mcp/` doesn't follow the [x402 v2 MCP transport spec](https://github.com/coinbase/x402/blob/main/specs/transports-v2/mcp.md). We use HTTP-status / HTTP-header semantics where the spec calls for in-band MCP tool-result + JSON-RPC `_meta` fields. This is internal tech debt that surfaced while reviewing #181 — the contributor's discovery endpoint advertises our current (non-spec) wire shape; aligning it with the spec is our work, not theirs.

## Current shape (what we ship today)

- Payment required → HTTP `402` from `/mcp` (or equivalent HTTP-status signaling)
- Payment payload → HTTP `payment-signature` header
- Settlement → HTTP `payment-response` header
- Discovery → `/.well-known/x402-payment` (added in #181, marked `"stability": "experimental"`)

## Spec-defined shape (x402 v2 MCP transport)

From [`specs/transports-v2/mcp.md`](https://github.com/coinbase/x402/blob/main/specs/transports-v2/mcp.md):

- **Payment-required signaling** — MCP tool result with `isError: true` and `structuredContent` carrying the `PaymentRequired` JSON (`x402Version: 2`, `accepts: [...]`, `resource`, etc.). Both `structuredContent` (REQUIRED) and `content[0].text` (REQUIRED, JSON-stringified) must be populated. There is **no** HTTP 402 on `/mcp`.
- **Payment payload** — `_meta["x402/payment"]` field in the MCP request `params`. Plain JSON, not base64-encoded. There is **no** `payment-signature` HTTP header.
- **Settlement response** — `_meta["x402/payment-response"]` field in the MCP response `result`. Plain JSON, not base64-encoded. There is **no** `payment-response` HTTP header.
- **Settlement failure after tool execution** — server returns the same `isError: true` + `PaymentRequired` shape; the executed tool's content is suppressed to avoid delivering a paid result without payment landing.

The OAuth `401` challenge on `/mcp` and the payment-required signaling don't collide because they happen at different layers — OAuth at the HTTP layer (request rejected with 401), payment-required at the MCP tool-call layer (tool result with `isError: true`). The in-band shape inherently disambiguates without needing the "don't make /mcp return 402" coordination we currently document.

## Deltas to land

- [ ] Replace HTTP-402-on-`/mcp` with MCP tool-result `isError: true` + `structuredContent` carrying `PaymentRequired`. Populate both `structuredContent` (object) and `content[0].text` (JSON-stringified) per spec.
- [ ] Read payment payload from `_meta["x402/payment"]` on incoming tool-call requests instead of the HTTP `payment-signature` header.
- [ ] Emit settlement via `_meta["x402/payment-response"]` on successful tool-call responses instead of the HTTP `payment-response` header.
- [ ] On settlement failure post-execution, return `isError: true` + `PaymentRequired`; suppress the executed tool's content.
- [ ] Remove the CORS exposure of `payment-signature` / `payment-required` / `payment-response` from the MCP router (they belong to the HTTP transport, not the MCP one).
- [ ] Decide the migration story for `/.well-known/x402-payment`:
  - Option A — **deprecate and remove.** The MCP spec doesn't define discovery via a well-known endpoint; discovery is implicit-on-first-call via the tool-result error.
  - Option B — **keep, but rework the payload** to describe the in-band mechanism (document the `_meta` key names and the tool-result error shape) rather than HTTP header names. If we go this route, propose upstream as a discovery extension since the spec doesn't currently define this endpoint.
- [ ] Simplify the OAuth-401-vs-payment-required section of `docs/api/09-mcp-integration.md` — the spec's in-band mechanism inherently disambiguates from the OAuth 401 challenge, so the "don't make /mcp return 402" coordination framing isn't needed once we're spec-aligned.
- [ ] Add tests covering the spec-shaped tool-result error AND the `_meta` round-trip for payment + settlement (both `structuredContent` and `content[0].text` paths on the error side, both happy-path and post-execution-settlement-failure on the response side).

## Out of scope

- **HTTP transport.** The HTTP-status / HTTP-header semantics we ship today are spec-correct for the HTTP transport (see [`specs/transports-v2/http.md`](https://github.com/coinbase/x402/blob/main/specs/transports-v2/http.md)) — this issue only covers MCP.
- **Bazaar / `GET /discovery/resources` aggregator endpoint** (section 8 of the core spec) — separate feature if we want a marketplace endpoint.

## Migration considerations

- #181 ships `/.well-known/x402-payment` with a `"stability": "experimental"` marker specifically so that this migration doesn't become a breaking-change negotiation with external consumers. The marker exists so we can either remove the endpoint or rework its payload as part of this work.
- Any downstream that built against the experimental endpoint should be migrated by us (likely just our own demos / tutorials).
- The Python SDK (`payments-py`) MCP middleware is where most of this lives, but the TypeScript SDK (`@nevermined-io/payments`) and `tutorials/mcp-examples/` may need parallel changes — survey before merging the rewrite.

## References

- [x402 v2 core spec](https://github.com/coinbase/x402/blob/main/specs/x402-specification-v2.md)
- [x402 v2 MCP transport spec](https://github.com/coinbase/x402/blob/main/specs/transports-v2/mcp.md)
- [x402 v2 HTTP transport spec](https://github.com/coinbase/x402/blob/main/specs/transports-v2/http.md) (for contrast; out of scope here)
- Surfaced during review of #181
- Reference implementation: [`agents/x402-mcp`](https://github.com/cloudflare/agents/blob/main/packages/agents/src/mcp/x402.ts) (Cloudflare) — linked from the spec

r-marques:
## Summary

Aligns the MCP transport with the [x402 v2 MCP transport spec](https://github.com/coinbase/x402/blob/main/specs/transports-v2/mcp.md), replacing HTTP-status / header payment semantics with **in-band JSON-RPC signaling**. Closes #203.

## What changed
- **Payment-required** → MCP tool result with `isError: true`, populating **both** `structuredContent` (the `PaymentRequired` object, real `accepts` from the agent's plans) and `content[0].text` (JSON-stringified). No more HTTP 402 on `/mcp`.
- **Request payment** read from `_meta["x402/payment"]` (plain JSON). The `Authorization: Bearer` header is kept as a **deprecated fallback** for one release.
- **Settlement** emitted via `_meta["x402/payment-response"]` (same receipt the HTTP transport puts in the `payment-response` header). Nevermined observability moves under `_meta["nevermined/credits"]`.
- **Post-execution settlement failure** suppresses the tool's content and returns the payment error.
- **Discovery removed** (Option A): the `/.well-known/x402-payment` endpoint and `enableX402Discovery` toggle are gone (discovery is implicit-on-first-call per spec); the `payment-*` CORS headers are removed.
- Adds `encode_access_token` (inverse of `decode_access_token`) to carry the in-band payload into the facilitator token, and module-level loggers for settle / plan-lookup failures.

## Decisions
- **Option A discovery removal** — the endpoint shipped `"stability": "experimental"` (#181/#215) specifically to allow this migration without a breaking-change negotiation.
- **`onRedeemError: "ignore"`** no longer delivers paid content on a post-execution settlement failure — spec-correct ("never deliver paid content without settlement landing"). It now only distinguishes the surfaced error type.

## Testing
- `pytest -m "not slow"`: **666 passed, 0 failed** (89 slow staging-E2E deselected). Black clean.
- New tests cover the in-band tool-result shape (dual representation), the **real** `call_tool` dispatcher wiring (in-band read → re-encoded Bearer, `except → payment_required_result` conversion, header fallback), the happy-path receipt + post-execution suppression, the free/no-credit path, and a semantic codec round-trip.

## Reviewed
Multi-agent review panel (correctness, silent-failure, test-coverage) run against the diff; all blocking + should-fix findings addressed (stale-test updates, logging, dispatcher-wiring tests).

## Caveats / follow-ups
- **Live-facilitator round-trip not yet validated**: re-encoding a real in-band `_meta["x402/payment"]` token through `verify`/`settle` is proven at the codec level (object round-trip; the EIP-712 signature is inside `payload`, not the base64 envelope) but not against a live facilitator. Recommend an external/e2e check before merge.
- **Resources & prompts** intentionally keep raising JSON-RPC errors (the spec's tool-result shape is tool-only).
- **Streaming tools** get neither the spec receipt nor post-execution suppression (inherent to streaming).
- The **TypeScript SDK** (`@nevermined-io/payments`) and `tutorials/mcp-examples/` need the same parallel change (out of scope here).


aaitor (commented):
Solid, spec-aligned migration to the x402 v2 in-band MCP transport, and this Python SDK is the reference implementation - well-tested (666 passing), including a real-dispatcher test (TestDispatcherInBand) that the TS sibling lacks. No HIGH issue survived verification. The main should-fix is a cross-SDK divergence: a backend plans-lookup outage is reported to clients as a clean payment-required, where the TS sibling reports plans-unavailable. A few LOWs below. Leaving this as COMMENT (non-blocking - no HIGH).

**Verification I ran**
- Read the full diff plus every changed file post-PR on feat/issue-203-mcp-x402-inband; confirmed execute_tool passes result _meta through verbatim (no stale top-level success/txHash reader) and build_payment_required_for_plans pre-exists in x402/helpers.py.
- 5-agent panel (code-reviewer, security-reviewer, python-reviewer, tdd-guide, code-explorer); cross-checked each finding against the code. Dropped a duplicated-comment finding (the diff removes one of two dupes) and a test-attribute finding (self._credits is correct, L38).
- CI: lint OK, unit_integration OK, e2e pending at review time.

**Out-of-scope / parity notes (not gating)**
- Streaming observability is silent: _RedeemOnCloseAsyncIterator (L437-478) computes _credits_result then discards it, so Python streams emit no _meta observability and a streaming settlement failure is invisible to the consumer (no parity with the TS final _meta chunk).
- Public-package breaking changes (tool payment-required no longer raises -32003; _meta keys moved under nevermined/credits; the well-known x402-payment endpoint plus enableX402Discovery plus payment-* CORS removed; onRedeemError ignore no longer returns content). Documented in the PR body, but the commit is feat(mcp) with no BREAKING CHANGE footer - confirm the version bump/changelog reflect the break.
- Deprecation log spam: server_manager logs the header-fallback deprecation on every tool call lacking an in-band payload, vs the TS once-per-process flag.
- onRedeemError propagate (raises -32002 Misconfiguration) has no test in either SDK.

payments_py/mcp/core/auth.py
```diff
@@ -23,6 +27,80 @@ def __init__(self, payments: Any) -> None:
...
+
+        payment_required = build_payment_required_for_plans(
+            plan_ids or [""],
+            endpoint=endpoint,
+            agent_id=agent_id,
+            http_verb="POST",
+            environment=getattr(self._payments, "environment_name", None),
+        )
+        pr_dict = payment_required.model_dump(by_alias=True)
+        pr_dict["error"] = "payment required"
```
aaitor: **MEDIUM - A backend plans-lookup outage is reported to the client as a clean payment-required.**

When get_agent_plans raises (network / 5xx / bad agent_id), the except at L78 logs it (good) but L101 still sets error to payment-required and accepts falls back to an empty plan id. A client cannot tell you-have-not-paid from our-backend-is-down, and may push the user to checkout during an outage. The TS sibling (payments#384, auth.ts) sets error to plans-unavailable in this case - the two SDKs are meant to mirror each other.

Track the failure and branch the error string (set plans-unavailable when the except fired, payment-required otherwise), and add a unit test asserting the plans-unavailable value when get_agent_plans raises (mirror the TS test buildPaymentRequiredError plans-lookup failure).

payments_py/mcp/core/paywall.py
```diff
@@ -304,6 +357,11 @@ async def _redeem(
                     "success": settle_success,
                     "txHash": settle_result.transaction if settle_success else None,
                     "creditsRedeemed": credits_burned,
+                    "settlement": (
```
aaitor, line 364: **LOW - settlement receipt is built even when settle_result.success is False.**

L360-364 calls model_dump() regardless of settle_success. It is currently latent - wrapped() raises SettlementFailedError on not-success (L240) before the receipt is read (L257) - but a future reorder would emit a spec settlement receipt for a failed settle. Guard the model_dump on settle_success (same for the fallback branch at L403).

payments_py/mcp/utils/meta.py
```diff
@@ -0,0 +1,80 @@
...
+    Args:
+        mcp_server: The low-level MCP ``Server`` instance (exposes
+            ``request_context``).
+
+    Returns:
+        The decoded PaymentPayload dict, or ``None``.
+    """
+    try:
+        meta = mcp_server.request_context.meta
+    except LookupError:
```
aaitor, line 52: **LOW - except LookupError is broader than the contextvar-miss it targets.**

LookupError is the base of KeyError/IndexError; an unrelated KeyError raised while accessing request_context would be swallowed and silently return None, bypassing the in-band read. Scope the try to just the contextvar access (wrap only mcp_server.request_context, then read .meta outside the try).

r-marques:
Thanks for the thorough review, @aaitor — all addressed in bbae326.

- **Plans-lookup outage reported as a clean payment-required (MEDIUM)** — `_build_payment_required_error` now tracks whether `get_agent_plans` raised and sets `error: "plans unavailable"` in that case (instead of `"payment required"`), matching the TS sibling. Added a test asserting the value and that the failure is logged.
- **Settlement receipt built on failure (LOW)** — `model_dump()` is now gated on `settle_success` in both the primary and fallback branches; no spec receipt is emitted for a failed settle.
- **Broad `except LookupError` (LOW)** — narrowed to wrap only the `request_context` access; `.meta` is read outside, so an unrelated `KeyError`/`IndexError` isn't swallowed.
- **Deprecation log spam** — the header-fallback notice now fires at most once per process (module-level flag), matching the TS behavior.
- **`onRedeemError: "propagate"` untested** — added a test asserting it raises Misconfiguration (`-32002`).

Deferred (with a tracked comment on `_RedeemOnCloseAsyncIterator`):
- **Streaming `_meta` observability parity** — the dispatch `str()`s non-dict/list results, so streamed tool results have no `_meta` channel to surface settlement observability through; matching the TS final-`_meta`-chunk behavior would require restructuring how the dispatch handles streamed results. Out of proportion to a non-gating LOW, so left as a documented follow-up.

On the **breaking-change / version** note: the branch carries a *minor* bump (`1.13.0`) but these changes are breaking — leaving the `BREAKING CHANGE:` footer / major-bump decision to the maintainers.

payments_py/mcp/core/server_manager.py
```diff
@@ -603,8 +611,37 @@ async def execute_tool(
...
+            # x402 v2 MCP transport: prefer the in-band payment payload from
+            # params._meta["x402/payment"]. Re-encode it into the access token
+            # string the verify/settle path expects and present it via the same
+            # extra/headers shape the auth flow already reads, so the in-band
+            # payload takes precedence over (and the existing Authorization
+            # header remains a deprecated fallback when absent).
+            payment_payload = read_payment_payload(self._mcp_server)
+            if payment_payload is not None:
+                token = encode_access_token(payment_payload)
+                extra = {
```
aaitor: **MEDIUM — In-band path drops all non-auth request headers from the user tool handler (silent regression + TS divergence).**

When an in-band `_meta["x402/payment"]` payload is present, `extra` is *replaced* wholesale here (L623-625) with an auth-only dict. That same `extra` is then forwarded to the user tool handler: `protected_handler(arguments, extra)` (L642) -> paywall `wrapped` -> `execute_tool(args, extra, ctx)` -> `tool_handler(args, extra, ctx)`. So any Python tool handler that reads request headers other than `authorization` (tenant/tracing/custom headers from `_get_request_extra()`, L612) silently receives an almost-empty headers dict — and *only* on the new in-band transport, i.e. exactly the path this PR steers clients toward. The old code passed `_get_request_extra()` straight through, so this is a new regression. The TS sibling deliberately avoids it: it builds a separate `authExtra` for auth and forwards the **raw** `extra` to the handler (`paywall.ts` L118-125, L166: "The RAW `extra` (with `_meta`) is still forwarded to the user handler below").

Minimal fix — override only the Authorization header, keep the rest of the request context:
```python
token = encode_access_token(payment_payload)
# Override only the Authorization header with the in-band token; keep the rest
# of the request context for the user handler (mirrors the TS raw-extra forward).
request_info = dict(extra.get("requestInfo") or {})
request_info["headers"] = {
    **(request_info.get("headers") or {}),
    "authorization": f"Bearer {token}",
}
extra = {**extra, "requestInfo": request_info}
```
(The fuller TS mirror — a separate `auth_extra` for the auth call only — is also fine but needs a second extra threaded through `wrapped`.) Please add a dispatcher test asserting a non-auth header in the request context survives to the handler when an in-band payload is present.

payments_py/mcp/core/server_manager.py
```diff
@@ -603,8 +611,37 @@ async def execute_tool(
...
+                    "requestInfo": {"headers": {"authorization": f"Bearer {token}"}}
+                }
+            elif self._log and not _x402_header_fallback_warned["done"]:
```
aaitor: **MEDIUM — Deprecation nudge never fires without an `onLog` callback (TS warns unconditionally).**

The once-per-process header-fallback warning is gated on `self._log`, so a server that didn't pass an `onLog` callback — the common case — never sees it, even while on the deprecated header path. The TS sibling calls `console.warn` unconditionally (`paywall.ts` L126-132). The whole point is to nudge migration before the header path is removed, so emit via the stdlib logger regardless of `onLog`:
```python
elif not _x402_header_fallback_warned["done"]:
    _x402_header_fallback_warned["done"] = True
    logging.getLogger(__name__).warning(
        "x402: no _meta['x402/payment'] on tool call; falling back to the "
        "Authorization header (deprecated under the x402 v2 MCP transport). "
        "Shown once per process."
    )
```
(Keep `self._log` as an additional sink if you want the callback to receive it too.)

payments_py/mcp/core/paywall.py
```diff
@@ -223,28 +232,72 @@ async def wrapped(*all_args: Any) -> Any:
...
-                }
-            )
+            # Spec-defined settlement receipt (x402 v2 MCP transport). Absent for
+            # free/no-credit calls where no settlement occurred.
+            settlement = credits_result.get("settlement")
+            if settlement is not None:
+                result["_meta"][X402_PAYMENT_RESPONSE_META_KEY] = settlement
+
+            # Nevermined-namespaced observability (NOT part of the x402 spec).
+            nvm_meta = {
```
aaitor, line 262: **MEDIUM — `nevermined/credits` omits `remainingBalance` (TS includes it).**

The TS sibling puts `remainingBalance` in the `nevermined/credits` block (`paywall.ts` L239); this `nvm_meta` (L262-268) doesn't, so a consumer reading `_meta["nevermined/credits"]["remainingBalance"]` across both SDKs gets it in TS but not Python. (It *is* present in the `x402/payment-response` spec receipt via `model_dump`, so the data isn't lost — but the namespaced block diverges.) Surface it from `_redeem` and add it here:
```python
nvm_meta = {
    "txHash": credits_result.get("txHash"),
    "creditsRedeemed": credits_result.get("creditsRedeemed", "0"),
    "remainingBalance": credits_result.get("remainingBalance"),
    "planId": auth_result.get("plan_id"),
    "subscriberAddress": auth_result.get("subscriber_address"),
    "success": credits_result.get("success", False),
}
```
and add `"remainingBalance": settle_result.remaining_balance if settle_success else None` to the dicts `_redeem` returns (both the primary and fallback branches).

payments_py/mcp/core/paywall.py
```diff
@@ -368,7 +435,17 @@ async def _maybe_await(self, maybe_awaitable: Any) -> Any:
 
 
 class _RedeemOnCloseAsyncIterator:
-    """Wrap an async-iterable to ensure a callback runs on completion or early close."""
+    """Wrap an async-iterable to ensure a callback runs on completion or early close.
+
+    Settlement runs on completion and the result is stored in ``_credits_result``,
```
aaitor, line 447: **MEDIUM (test debt) — streaming redeem-on-close has no test.**

`_RedeemOnCloseAsyncIterator` is what guarantees credits are settled for *streaming* tools, via two paths: natural completion (`__anext__` -> `StopAsyncIteration`, L475) and early close (`aclose`, L478-488). Neither is exercised by the suite, so a regression that stopped calling `on_finally` would silently stop settling streaming tools with no failing test. Add:
```python
@pytest.mark.asyncio
async def test_redeem_runs_on_stream_completion():
    calls = []
    async def src():
        yield 1; yield 2
    async def on_finally():
        calls.append(1); return {"success": True, "creditsRedeemed": "5"}
    it = _RedeemOnCloseAsyncIterator(src(), on_finally)
    assert [x async for x in it] == [1, 2]
    assert calls == [1] and it._credits_result["success"] is True

@pytest.mark.asyncio
async def test_redeem_runs_once_on_early_close():
    calls = []
    async def src():
        yield 1; yield 2; yield 3
    async def on_finally():
        calls.append(1); return {"success": True}
    it = _RedeemOnCloseAsyncIterator(src(), on_finally)
    async for _ in it:
        break
    await it.aclose()
    assert calls == [1]   # idempotent: not redeemed twice
```

payments_py/mcp/core/paywall.py
```diff
@@ -341,8 +400,16 @@ async def _redeem(
...
                         "creditsRedeemed": credits_burned,
+                        "settlement": (
+                            settle_result.model_dump(by_alias=True)
+                            if settle_success and hasattr(settle_result, "model_dump")
+                            else None
+                        ),
                     }
                 except Exception as fallback_error:
+                    logger.error(
+                        "x402 settle failed (fallback endpoint): %s", fallback_error
```
aaitor, line 418: **MEDIUM (test debt) — the settle fallback-endpoint retry is untested in Python.**

The TS PR added a red->green regression test for exactly this path after fixing a swapped `fallbackEndpoint`/`httpVerb` arg order (`x402_inband.test.ts`: "settle fallback uses httpUrl, not the verb"). The Python `_redeem` has the same primary->fallback retry (L375-413) but no test drives it, so the same kind of transposition bug here would be invisible. Add a test where `settle_permissions` raises on the first call and succeeds on the second, asserting the 2nd call's `payment_required.resource.url == <http fallback>` (and `!= "POST"`).

aaitor (changes requested):
Re-review of the in-band x402 migration after `bbae326`. The spec-shaped in-band signaling, the plans-unavailable distinction, the post-execution suppression, and the dispatcher wiring all look correct and are well covered (666 passing). I confirmed `SettleResponse` carries camelCase aliases, so the `x402/payment-response` spec receipt is key-compatible with the TS sibling — the snake_case worry does not apply.

Requesting changes on one item the first pass missed: the dispatcher now **replaces** `extra` wholesale on the in-band path, so user tool handlers silently lose every non-`authorization` request header — but only on the new transport this PR steers clients toward. The TS sibling deliberately preserves the raw extra; Python should match. The remaining items (deprecation nudge gated on `onLog`, `remainingBalance` parity, two untested load-bearing paths) are should-fix MEDIUMs.

**Verification I ran**
- Read the full diff + every changed file post-PR on `feat/issue-203-mcp-x402-inband`; traced the dispatcher -> paywall -> user-handler `extra` flow and compared old vs new `server_manager.py` to confirm the header-drop is a *new* regression (the old code passed `_get_request_extra()` straight through to the handler).
- Confirmed `SettleResponse` (`payments_py/x402/types.py` L250-279) aliases `creditsRedeemed`/`remainingBalance`/`errorReason`/`orderTx`, so `model_dump(by_alias=True)` matches the TS receipt keys.
- 4-agent panel (security, python idioms, cross-SDK parity, test coverage); cross-checked every finding against the code. Dropped two over-rated HIGHs: the `decodeAccessToken` URL-safe claim (the TS L62 already swaps `-`/`_`) and the `model_dump` snake_case claim (aliases produce camelCase).
- CI: 3/3 green at review time.

**Non-gating LOWs**
- `test_plan_lookup_error_still_valid_and_logged` asserts `error == "plans unavailable"` but not the `accepts` shape — add `assert pr["accepts"] == [{...,"planId":""}]` (len 1, empty id) so an empty-`accepts` regression is caught.
- `_redeem` propagate message is `"Failed to settle credits"` vs the TS `"Failed to redeem credits"` — align the string.
- The spec receipt via `model_dump(by_alias=True)` includes null-valued keys (`payer`/`orderTx`/`errorReason`) the TS raw-JSON receipt omits; consider `exclude_none=True` for a tighter match.
- `read_payment_payload` decodes/re-encodes a client-supplied object with no size bound — a cheap `len(json.dumps(value)) <= 64K` guard before returning is reasonable defense-in-depth (same applies to the TS sibling).

**Parity note** — TS (#384) is the reference for the three parity items (raw-extra forwarding, unconditional deprecation warn, `remainingBalance`); only Python needs aligning.

r-marques:
Thanks for the re-review, @aaitor — addressed in 3dd0c06.

**Blocker — in-band path dropped non-auth headers:** fixed. The dispatcher now overrides only the `Authorization` header and preserves the rest of the request context for the user handler (your minimal-fix shape), instead of replacing `extra` wholesale. Added `test_inband_preserves_non_auth_request_headers` asserting a non-auth header survives to the handler alongside the injected Bearer token.

**MEDIUMs:**
- Deprecation nudge now emits via `logging.getLogger(__name__).warning` regardless of `onLog` (once per process; still forwards to `onLog` if present).
- `nevermined/credits` now includes `remainingBalance`, surfaced from `settle_result.remaining_balance` in both the primary and fallback `_redeem` branches.
- Added the two streaming redeem-on-close tests (natural completion + idempotent early-close) and the settle primary→fallback retry test (asserts the fallback builds `resource.url == httpUrl`, not `"POST"`).

**LOWs:**
- `test_plan_lookup_error_*` now also asserts `accepts == [{planId: ""}]` (one entry, empty id).
- `_redeem` propagate message aligned to `"Failed to redeem credits"`.
- Spec receipt now uses `model_dump(by_alias=True, exclude_none=True)`, so null `payer`/`orderTx`/`errorReason` are omitted (tighter match with the TS receipt).
- `read_payment_payload` rejects in-band payloads > 64KB (defense-in-depth); the same guard was added to the TS sibling.

672 passing, black clean. (Version is still a minor bump — the `BREAKING CHANGE:` footer / major-bump call remains with the maintainers.) Ready for re-review.

aaitor (approved):
Re-review of `bbae326..3dd0c06` ("address aaitor 2nd review"). Every finding from the CHANGES_REQUESTED pass is resolved with a correct fix and a real regression test, and CI is green (3/3). Clearing the block — approving.

**Resolved (verified line-by-line against the delta, not just the PR comment):**
- **[blocker] In-band `extra` overwrite** → now merges the in-band Bearer into the existing `requestInfo.headers` instead of replacing `extra` wholesale (`server_manager.py`), so the user handler keeps tenant/tracing/custom headers. `test_inband_preserves_non_auth_request_headers` sets `x-tenant-id`/`x-trace-id` and asserts both survive alongside the injected Bearer. Edge cases (empty extra, missing `requestInfo`/`headers`, stale Authorization override) all handled.
- **Deprecation nudge** → now emits via `logging.getLogger(__name__).warning` regardless of `onLog` (still forwards to `onLog` if set).
- **`remainingBalance` parity** → added to `nevermined/credits` and surfaced from `settle_result.remaining_balance` in both `_redeem` branches.
- **Streaming redeem-on-close** → `TestStreamingRedeemOnClose` covers natural completion + idempotent early-close.
- **Settle fallback retry** → `test_fallback_uses_http_url_not_verb` drives primary→fallback and asserts the fallback `resource.url == httpUrl` (≠ `"POST"`), guarding the swapped-args class of bug.
- LOWs: `accepts` shape now asserted on the plans-unavailable path; propagate message aligned to `"Failed to redeem credits"`; spec receipt uses `exclude_none=True` (test asserts `orderTx`/`errorReason` dropped); `read_payment_payload` rejects >64KB payloads (tested).

**Verification I ran**
- `gh api compare bbae326...3dd0c06` — read every source + test hunk; traced the `extra`-merge edge cases and the `remainingBalance` threading through both `_redeem` branches.
- Confirmed the existing-test edits (`_FakeSettlement.model_dump` now models `by_alias`/`exclude_none`; mock settle results gain `remaining_balance`) **strengthen** assertions rather than weaken them.
- CI: e2e / lint / unit_integration all green; 672 passing per the author.

Non-gating, for the release runner: the branch still carries a *minor* bump while the change is breaking (tool payment-required no longer raises `-32003`; `_meta` keys moved under `nevermined/credits`; the well-known x402-payment endpoint + `enableX402Discovery` removed; `onRedeemError: ignore` no longer returns content). Confirm the `BREAKING CHANGE:` footer / version bump before release — as you both already flagged, that call is the maintainers'.

PR #235, opened 2026-06-17
feat(mcp): make agentId optional, require planId

r-marques:
## Summary
Make `agentId` **optional** in the MCP server and require `planId` instead — plan-centric, matching the direction that only the pricing plan is necessary.

## Why it's safe (empirically verified)
On staging, `verifyPermissions`/`settlePermissions` succeed with a `PaymentRequired` built with **no agentId** and a no-agentId token (`isValid:true`, settle success, credits charged). The facilitator resolves everything from `planId` + the token/delegation; `agentId` is informational only (response showed `agentId:""`, `agentName:"Unknown Agent"`).

## Changes
- Server config (`McpServerConfig`, paywall config, `configure`/`start`/`with_paywall`/`attach`): **`planId` required, `agentId` optional**.
- Guard now requires an effective `planId` (per-tool option ?? server config); the four `agentId is required` throws → `planId is required`.
- First-call payment-required `accepts` built from the configured `planId`; `getAgentPlans(agentId)` only when an agentId is present.
- `agentId` threaded as optional into `buildPaymentRequired`/auth (already optional in the builder). Verify/settle work plan-only.

## ⚠️ Maintainer decision — OAuth `client_id`
`agentId` also served as the OAuth `client_id`. With it optional, this PR omits `client_id` from the protected-resource metadata when absent and generates an id for dynamic registration. The in-band/bearer flow is unaffected, but confirm no full-OAuth-dance client relies on a stable `client_id == agentId`.

## Tests
`poetry run pytest -m "not slow"` 674 passed (incl. a new "works with planId, no agentId" test + a missing-planId guard test); black clean.

## Release
Should ship in the same new version as the in-band transport (nevermined-io/payments-py#228), which is merged to `main` but not yet released.


payments_py/mcp/core/server_manager.py
```diff
@@ -847,7 +852,8 @@ def _log_startup_message(self, info: ServerInfo, config: McpServerConfig) -> Non
   Tools: {tools_list}
   Resources: {resources_list}
   Prompts: {prompts_list}
-  Agent ID: {config['agentId']}""")
+  Plan ID: {config['planId']}{f"""
+  Agent ID: {config['agentId']}""" if config.get('agentId') else ''}""")
```
aaitor: **HIGH — Nested same-quote f-string is a `SyntaxError` on Python 3.10/3.11; this is why CI is red.**

Lines 855-856 nest an `f"""…"""` *inside* another `f"""…"""` replacement field. That's legal only on Python ≥3.12 (PEP 701). The repo supports Python 3.10+ and CI runs **3.10**, so this raises `SyntaxError: f-string: expecting '}'` at import time — collection of **all 7 MCP test files fails** and the `unit_integration` job exits 2. The PR body's "674 passed" came from a local 3.12+ interpreter; it doesn't hold on the supported floor.

One-click fix (concatenate instead of nesting — valid on 3.10; run `black` after):

```suggestion
  Plan ID: {config['planId']}""" + (
            f"\n  Agent ID: {config['agentId']}" if config.get("agentId") else ""
        ))
```

Cleaner alternative — hoist the agent line out of the f-string entirely:

```python
        agent_line = f"\n  Agent ID: {config['agentId']}" if config.get("agentId") else ""
        self._log(f"""MCP Server Started!
  ...
  Plan ID: {config['planId']}{agent_line}""")
```

Add a Python 3.10 entry to the test matrix so a 3.12-only syntax can't pass review again.

payments_py/mcp/payments_mcp.py
```diff
@@ -149,6 +149,7 @@ def __init__(
         self,
         payments: Any,
         name: str = "payments-mcp-server",
+        plan_id: Optional[str] = None,
         agent_id: Optional[str] = None,
         version: str = "1.0.0",
         description: Optional[str] = None,
```
aaitor, line 156: **MEDIUM — Inserting `plan_id` as the 3rd positional parameter silently breaks positional callers.**

The constructor is now `PaymentsMCP(payments, name, plan_id, agent_id, version, description)`. A caller that passed the agent DID positionally — `PaymentsMCP(payments, "my-server", "did:nv:abc")` — now binds that DID to `plan_id` and leaves `agent_id=None`. Nothing raises (both are `Optional[str]`), so the server starts and charges against a bogus plan id. Make the payment params keyword-only so the positional slot can't be mis-bound:

```suggestion
        self,
        payments: Any,
        name: str = "payments-mcp-server",
        *,
        plan_id: Optional[str] = None,
        agent_id: Optional[str] = None,
        version: str = "1.0.0",
        description: Optional[str] = None,
```

aaitor (changes requested):
The plan-centric direction is sound and the implementation mirrors the TS sibling (nevermined-io/payments#391) cleanly — but this currently **does not build on the supported Python floor**: a nested same-quote f-string in `server_manager.py` is a `SyntaxError` on 3.10/3.11, so `unit_integration` is red and all 7 MCP test files fail to collect (the "674 passed" was a local 3.12+ run). Fix that, make the new constructor params keyword-only, and label the breaking change; the rest is LOW polish.

**Verification I ran**
- `gh pr diff 235` + read post-PR `auth.py`, `paywall.py`, `server_manager.py`, `payments_mcp.py`, `oauth_metadata.py`, `client_registration.py` at `feat/mcp-agentid-optional`.
- Reproduced the failure from the CI log: `SyntaxError: f-string: expecting '}'` ×7, `unit_integration` exit 2. Confirmed lines 855-856 nest `f"""…"""` inside `f"""…"""` (PEP 701, Python ≥3.12 only; CI runs 3.10).
- Cross-checked the TS sibling payments#391 (same change) — its CI is green, so this is Python-only.
- Verified the security claim: `agent_id` flows into `build_payment_required_for_plans(… agent_id=agent_id or None …)` as informational `extra` only; verify/settle resolve from planId + token. Parallel security review found no authorization bypass introduced by dropping agentId.
- Confirmed the Python `paywall.protect` planId guard lives inside `wrapped` (line 93) — same deferred-to-first-call behaviour as before, not a regression.
- Panel: python-reviewer, security-reviewer, ponytail/leanness. Leanness verdict: **lean already** — no extraction warranted (unlike the TS side's 3×-duplicated `configure`).

**MEDIUM (release hygiene) — label the breaking change.** This flips the required field: `agentId` required→optional, `planId` optional→required. Every existing caller configured with only `agentId` now hits `ValueError: plan_id is required` / `Server misconfiguration: missing planId`. Add a `BREAKING CHANGE:` footer and a changelog/README note, and make sure the shared release version with #228 is bumped accordingly.

**LOW (optional; pre-existing, cheap while you're in these files)**
- `payments_py/mcp/types/paywall_types.py:12` — `AuthResult` declares camelCase (`agentId`, `logicalUrl`, `httpUrl`) but `auth.py` writes snake_case keys (`agent_id`, `logical_url`, `http_url`, lines 232-235 / 254-257). Harmless at runtime (`total=False`, consumers use `.get("agent_id")`) but the TypedDict is misleading; rename to snake_case.
- `auth.py:103,179` — `agent_id or None` is a no-op now that the param is `Optional[str]` and callers (`paywall.py:166`, `index.py:161`) already normalize `""`→`None`. Drop it.
- `oauth_metadata.py:331` — `config["agentId"]` subscript inside an `if config.get("agentId"):` guard; assign to a local for consistency with the surrounding `.get()` style.

r-marques:
Addressed the review — **CI is now fully green (`unit_integration`, `lint`, `e2e` all pass)**.

**The blocker (py3.10 f-string `SyntaxError`)** was already fixed in `eb7afb0`: the nested triple-quoted f-string in `server_manager.py` was extracted to a plain `agent_line` variable, validated on real 3.10 via `uv run --python 3.10 … ast.parse`. That's why `unit_integration` is green now.

**The remaining red check was `e2e`,** same root cause as the TS sibling (#391): the e2e fixture (`tests/e2e/mcp_tests/test_mcp_oauth_server.py`) started the server without `planId`. Fixed by adding `"planId": "plan-123"` (the id the existing `get_agent_plans`/decode mocks already use) to both `configure()` and the `start()` options.

**LOW items:**
- **AuthResult camelCase→snake_case:** done — and it surfaced a latent bug. `credits_context.py` read `auth_result.get("logicalUrl")` (camelCase), but `auth.py` only ever wrote `logical_url`, so the callable-credits `ctx.request.logicalUrl`/`toolName` were **always empty**. Fixed the source lookup to snake_case while keeping the public output key `"logicalUrl"` camelCase (the credits-callable contract). Net effect: those fields are now populated as intended — a minor behaviour change worth noting.
- **`agent_id or None` no-ops:** dropped at both sites (callers already normalize `""`→`None`).
- **`oauth_metadata.py` subscript:** assigned to a local for `.get()`-style consistency.

**Breaking-change footer:** holding on the maintainer's call (shared release with #228).

Commit `7d3de16`.

payments_py/mcp/payments_mcp.py
```diff
@@ -149,6 +149,7 @@ def __init__(
         self,
         payments: Any,
         name: str = "payments-mcp-server",
+        plan_id: Optional[str] = None,
```
aaitor, line 153: **MEDIUM (carryover, still open) — `plan_id` as the 3rd positional parameter can silently mis-bind.**

This is the one item from the prior review that's unaddressed. The signature is `PaymentsMCP(payments, name, plan_id, agent_id, …)`, so a caller that passed the agent DID positionally — `PaymentsMCP(payments, "my-server", "did:nv:abc")` — now binds that DID to `plan_id` and leaves `agent_id=None`. Nothing raises (both `Optional[str]`); `start()` then accepts the DID as a plan id and charges against it. Make the payment params keyword-only:

```python
    def __init__(
        self,
        payments: Any,
        name: str = "payments-mcp-server",
        *,
        plan_id: Optional[str] = None,
        agent_id: Optional[str] = None,
        version: str = "1.0.0",
        description: Optional[str] = None,
    ) -> None:
```

Non-blocking (no positional callers in the docs/tests), but cheap insurance for a public constructor.

aaitor (approved):
Re-review at `6f89d4cc`. The Python-3.10 f-string blocker is fixed and `unit_integration` is now green, so I'm clearing my earlier **CHANGES_REQUESTED** and approving. The author addressed almost everything from the prior round.

**Fixed since my last review (`d735877`)**
- ✅ **HIGH** — `server_manager.py:847-862`: the nested triple-quoted f-string is gone; `agent_line` is hoisted to a plain variable (valid on 3.10), rendered output unchanged. CI confirms.
- ✅ **LOW** — `paywall_types.py`: `AuthResult` fields renamed to snake_case (`agent_id`/`logical_url`/`http_url`), now matching the keys `auth.py` actually writes and `paywall.py`/`credits_context.py` read.
- ✅ **LOW** — `auth.py`: redundant `agent_id or None` removed (callers normalize `""`→`None` upstream).
- ✅ **LOW** — `oauth_metadata.py:330-332`: `config["agentId"]` subscript replaced with a guarded local.

**Verification I ran**
- Read post-fix `server_manager.py`, `auth.py`, `paywall_types.py`, `payments_mcp.py`, `oauth_metadata.py` at `feat/mcp-agentid-optional`; diffed `d735877…6f89d4cc` (filtering out the unrelated `main`-merge files: `a2a/*`, `api/*`, user-scoped-listing).
- CI: `unit_integration` ✅, `lint` ✅, `e2e` pending.
- Focused python-reviewer pass on the delta. It raised the skipped-e2e fixture mock (`verify_permissions` returns `{"isValid": True}` while `auth.py` reads `result.is_valid`) — I traced it: **pre-existing, not in this PR's diff, and `TestMcpEndpointWithAuth` is `@pytest.mark.skip` for an unrelated SSE-transport reason**, so it's out of scope here (worth a separate cleanup ticket whenever those tests are unskipped).

**Still open**
- The constructor positional-`plan_id` MEDIUM (inline) — non-blocking.
- Breaking-change labeling: `agentId` required→optional / `planId` optional→required still ships without a `BREAKING CHANGE:` footer. The `CHANGELOG.md` in the diff is the auto-generated v1.14.0 train (in-band #228 etc.), not a note for this change. Make sure the release notes call out the required-field flip.

r-marques:
Folded in the carryover MEDIUM from your re-review — `d3cbef8`:
- **`PaymentsMCP.__init__`:** inserted `*` so `plan_id`/`agent_id`/`version`/`description` are keyword-only. A DID passed in the 3rd positional slot can no longer silently mis-bind to `plan_id`. Confirmed every existing call site (src/tests/docs) already uses keywords.

20 constructor tests pass, `black` clean, py3.10 parse OK. Breaking-change footer still held for the shared release with #228.

PR #236, opened 2026-06-17
feat(a2a): align A2A with x402 v2 in-band transport + canonical agent-card discovery

aaitor:
## Why

The A2A integration only interoperated with Nevermined's **own** SDK. The agent card was served at the pre-spec `/.well-known/agent.json`, but stock A2A clients (and `a2a-sdk` 0.3+) discover it at `/.well-known/agent-card.json`, so third-party A2A clients 404'd on Nevermined Python agents. And payment used a home-grown HTTP-header paywall rather than the standard, so a generic A2A agent couldn't pay a Nevermined agent. A2A's whole value is cross-vendor interoperability, so this defeated the purpose.

This is the A2A analog of the MCP in-band migration the team already shipped (#228), applying the same approach to the A2A transport. Notably, the in-band utilities (`X402A2AUtils`) already existed in `payments_py/x402/a2a.py` but were **latent** — this PR wires them into the live server/handler.

### 1. Canonical agent-card discovery (interop fix)
- Serve the card at the canonical `/.well-known/agent-card.json`, keeping `/.well-known/agent.json` as a backward-compat **alias**. (The client synthesizes a minimal card from the base URL, so no client change is needed.)

### 2. In-band x402 v2 A2A transport (standards payment flow)
Per the [Coinbase x402 v2 A2A transport spec](https://github.com/coinbase/x402/blob/main/specs/transports-v2/a2a.md) + the official [a2a-x402 extension](https://github.com/google-agentic-commerce/a2a-x402):
- The agent card now declares the official extension URI (`.../a2a-x402/blob/main/spec/v0.2`) alongside `urn:nevermined:payment` (kept one release). `x402/a2a.py` spec references bumped v0.1 → v0.2.
- **Payment required** → server returns an `input-required` Task with `x402.payment.status: payment-required` + `x402.payment.required` metadata (no HTTP 402).
- **Payment submitted** → the payload is read from the message's `x402.payment.payload` metadata, re-encoded to the base64 token the facilitator verify/settle path already expects.
- **Settlement** is stamped in band (`payment-completed` + `x402.payment.receipts`). **A settlement failure forces `failed`/`payment-failed` and drops paid content — status message, artifacts AND history — so a paid result is never delivered without settlement landing.**
- The legacy `payment-signature` header flow is kept as a **deprecated fallback (one release)**.

New module `payments_py/a2a/inband.py` (in-band payload extraction + token re-encode) wires the latent `X402A2AUtils` into the server.

## Backward compatibility
- Legacy `/.well-known/agent.json` still served (alias); legacy `payment-signature` header flow still works (deprecated).
- The shared `x402/a2a.py` change is covered by the existing MCP suite (no regression).

## Tests
- New coverage: first-contact → `input-required`; in-band payload → verify/execute/settle/receipts; verify/settle failure with **content + artifact** suppression; legacy-header regression.
- `black` clean; **85 a2a / 691 full suite** (`-m "not slow"`) green.

## Deferred (out of scope)
- **A2A 1.0** — `a2a-sdk` 1.x is GA but `@a2a-js/sdk` 1.0 is still alpha; defer to keep TS/Python parity.
- **AP2 embedded flow** — the in-band structures are AP2-nestable; build when a counterparty exists.
- **Agent-card JWS signing** — production hardening, later.

## Pre-release note
The in-band token round-trip is verified at the codec level + against the mocked facilitator. A live-facilitator E2E pass on staging is worth running before release, same caveat as the MCP PRs.

### DIFF
--- a/docs/api/10-a2a-integration.md
+++ b/docs/api/10-a2a-integration.md
@@ -64,7 +64,7 @@ agent_card = build_payment_agent_card(
 
 ### Agent Card Structure
 
-The agent card includes a payment extension:
+The agent card declares two extensions: the Nevermined payment extension (pricing metadata) and the official a2a-x402 extension (`https://github.com/google-agentic-commerce/a2a-x402/blob/main/spec/v0.2`), which signals support for the standards-compliant in-band x402 v2 flow (see [In-Band x402 v2 Payments](#in-band-x402-v2-payments-standards-flow)). Both ship for one release; `urn:nevermined:payment` is dropped once clients use v0.2 only.
 
 ```json
 {
@@ -82,6 +82,9 @@ The agent card includes a payment extension:
           "planIds": ["plan-basic", "plan-premium"],
           "costDescription": "1-5 credits depending on review depth"
         }
+      },
+      {
+        "uri": "https://github.com/google-agentic-commerce/a2a-x402/blob/main/spec/v0.2"
       }
     ]
   }
@@ -185,7 +188,7 @@ asyncio.run(result.server.serve())
 | `port` | `int` | No | Server port (default: 8080) |
 | `task_store` | `TaskStore` | No | Task storage implementation |
 | `base_path` | `str` | No | Base URL path (default: "/") |
-| `expose_agent_card` | `bool` | No | Expose /.well-known/agent.json |
+| `expose_agent_card` | `bool` | No | Expose /.well-known/agent-card.json (+ legacy /.well-known/agent.json alias) |
 | `hooks` | `dict` | No | Request lifecycle hooks |
 | `async_execution` | `bool` | No | Enable async task execution |
 
@@ -207,6 +210,36 @@ payment-required: eyJ4NDAy... (base64-encoded X402PaymentRequired)
 {"error": {"code": -32001, "message": "Missing bearer token."}}
 ```
 
+## In-Band x402 v2 Payments (Standards Flow)
+
+Payment is signalled **in band** following the [Coinbase x402 v2 A2A transport spec](https://github.com/coinbase/x402/blob/main/specs/transports-v2/a2a.md) and the official [a2a-x402 extension](https://github.com/google-agentic-commerce/a2a-x402). The `X402A2AUtils` helpers (see [x402 README](../../payments_py/x402/README.md)) now power this live server flow — `PaymentsA2AServer` stamps and reads the metadata automatically, so the executor is unchanged. The handshake rides on A2A task/message metadata, correlated by `taskId` — no HTTP 402 is exchanged.
+
+Lifecycle:
+
+1. **Payment required** — a payment-gated `message/send` arrives with no payment. The server returns a Task with `status.state = "input-required"` and `status.message.metadata`:
+
+   ```json
+   {
+     "x402.payment.status": "payment-required",
+     "x402.payment.required": { "...": "X402PaymentRequired (accepts[] of plans)" }
+   }
+   ```
+
+2. **Payment submitted** — the client replies with a follow-up `message/send` whose `message.metadata` carries the payload, correlated via `message.taskId`:
+
+   ```json
+   {
+     "x402.payment.status": "payment-submitted",
+     "x402.payment.payload": { "...": "PaymentPayload" }
+   }
+   ```
+
+3. **Completed / failed** — on success the final Task carries `x402.payment.status: "payment-completed"` plus `x402.payment.receipts`. On failure it is `failed` / `payment-failed` with the error under `x402.payment.error`, and paid content is suppressed.
+
+Metadata keys: `x402.payment.status`, `x402.payment.required`, `x402.payment.payload`, `x402.payment.receipts`, `x402.payment.error`.
+
+> **Note:** The legacy `payment-signature` HTTP header flow still works but is now a **deprecated fallback, kept for one release**. New integrations should use the in-band metadata flow above.
+
 ## Client Usage
 
 ### Discovering Plans from the Agent Card
@@ -217,7 +250,8 @@ Consumers can fetch the agent card to discover available plans:
 import httpx
 
 async with httpx.AsyncClient() as client:
-    resp = await client.get("http://agent-url/.well-known/agent.json")
+    # Canonical path (A2A 0.3+); /.well-known/agent.json still works as a legacy alias
+    resp = await client.get("http://agent-url/.well-known/agent-card.json")
     card = resp.json()
 
 extensions = card["capabilities"]["extensions"]
--- a/payments_py/a2a/agent_card.py
+++ b/payments_py/a2a/agent_card.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 from typing import Any, Dict, List
 
 from payments_py.a2a.types import PaymentAgentCardMetadata
+from payments_py.x402.a2a import A2A_X402_EXTENSION_URI
 
 # Canonical A2A agent-card discovery path (A2A >= 0.3, per RFC 8615). Served by
 # Nevermined A2A agents and used as the default when fetching a remote card.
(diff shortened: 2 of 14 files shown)


r-marques (commented):
## 🤖 Automated PR review — 🔴 Blocked

_Reviewed `origin/main...HEAD` (14 files, +1206/−73) with the pr-review-toolkit panel: code-reviewer, silent-failure-hunter, pr-test-analyzer, type-design-analyzer. Scoped to the diff; `.env*` skipped. Suite runs green (43 tests) per the test-analyzer._

This is a well-built change — the in-band x402 v2 flow is conservatively designed, the core "never deliver unpaid content" rule is implemented and (mostly) tested, body-restore is correct on every `call_next` path, and there's no new cross-tenant oracle or issuer/audience leak in the curated failure path. One blocker keeps it from being merge-ready: the settlement-failure content suppression is incomplete.

### 🔴 Blockers (1)

- **Settlement-failure suppression leaks paid content via `task.status.message.parts`** — `payments_py/a2a/payments_request_handler.py:759-779` (`_apply_inband_settlement`, failure branch) + `payments_py/x402/a2a.py:441-448` (`record_payment_failure`). _(code-reviewer, confidence 88; corroborated by pr-test-analyzer)_
  On settlement failure the code forces `state=failed`, sets `artifacts=None` and `history=[]`, then calls `record_payment_failure`. But `record_payment_failure` only **synthesizes** a status message when one is *absent* (`if not ... task.status.message:`). On a task that already executed to completion, `task.status.message.parts` carries the agent's deliverable text, and the failure branch never clears it — only `artifacts`/`history` are dropped. The success-path test asserts the deliverable lives exactly there (`task["status"]["message"]["parts"][0]["text"]`, `test_inband_x402.py:191-195`), so for any agent that returns its result as status-message text (a common A2A pattern), that paid content survives a failed settlement — defeating the PR's headline guarantee.
  **Why the existing test misses it:** `test_inband_settlement_failure_suppresses_content` asserts `"PAID_SECRET_RESULT" not in response.text`, but `PAID_SECRET_RESULT` is emitted as an *artifact* (which is cleared); the surviving status-message text is the generic `"Request completed successfully!"`, which the assertion never checks. The whole-body check gives false confidence here.
  **Fix:** in the failure branch, also neutralize the status message before re-stamping — e.g. set `task.status.message = None` (so `record_payment_failure` synthesizes the neutral "Payment failed." message) or replace its parts. Then extend the test so the status-message parts are asserted to no longer contain the agent's deliverable.

### 🟡 Should fix (6)

- **Verify exception swallowed with no `Logger.error` before transforming** — `payments_py/a2a/server.py:329-348`. The new in-band branch catches *every* exception from `validate_request`, stringifies it into `validation={"verify_error": str(exc)}`, and returns `200 OK` with a `payment-failed` task — with nothing logged server-side. This violates the repo's mandatory "always `Logger.error()` the original error before re-throwing/transforming" rule (CLAUDE.md), and now hides facilitator outages / programming errors behind a "your payment failed" response with no trace. Log before transforming; ideally split `except PaymentsError` (expected) from `except Exception` (unexpected, `exc_info=True`). _(silent-failure-hunter, HIGH)_
- **Raw `str(exc)` forwarded into client-visible task metadata** — `payments_py/a2a/server.py:337` → `x402.payment.error`. The catch is `except Exception`, not `except PaymentsError`, so an unexpected exception's message (or a backend-wrapped transport error containing a URL / connection detail) is forwarded verbatim across the trust boundary to the A2A caller. Forward a curated reason for `PaymentsError`; use a fixed generic reason for anything else and log the detail server-side (pairs with the finding above). _(silent-failure-hunter, HIGH)_
- **`_coerce_settle_response` duck-typed fallback defaults `success=True`** — `payments_py/a2a/payments_request_handler.py:471` (`getattr(settle_result, "success", True)`). A settle result you can't read `success` from is coerced to a *successful* settlement → `record_payment_success` → paid content delivered, the exact opposite of the safe default on a money path. **Latent, not live:** the real `facilitator.settle_permissions` always returns a `SettleResponse` or raises (`facilitator_api.py:182-194`), so this branch only fires for test doubles today — but the polarity is wrong and a future partial-object return would silently turn unpaid calls into "paid." One-line fix: default to `False`. _(type-design-analyzer + silent-failure-hunter, independent)_
- **`_settle_receipt_by_task` can grow unbounded when the receipt is captured but never stamped** — store at `payments_request_handler.py:650-667`, only `.pop()` at `:767`. `_apply_inband_settlement` is reached only when `isinstance(result, Task)` **and** `http_ctx.inband`; if `result` is a `Message`, or an exception fires between finalization and the stamp (e.g. `_validate_task_id_match` at `:407`), the entry is orphaned — the `finally` block cleans up the HTTP ctx but not this map. Slow but genuinely unbounded on a long-lived server. Fix: `self._settle_receipt_by_task.pop(task_id, None)` in the `finally` of `on_message_send`. _(code-reviewer + pr-test-analyzer)_
- **Settlement-failure unit test asserts content dropped only vacuously** — `tests/unit/a2a/test_inband.py:188-205`. The fixture builds the task with `history=[]` already, so the assertion that history is dropped passes even if the `task.history = []` line (`payments_request_handler.py:778`) were deleted. Give the fixture a non-empty `history=[<message>]` (and a status message with parts) and assert both are cleared — this is the regression guard for the blocker above. _(pr-test-analyzer, rating 6)_
- **Settlement failure logged at `warning`, not `error`** — `payments_py/a2a/payments_request_handler.py:654-659`. A failed credit settlement is a money-movement failure (work done, charge didn't land); it's the event you most want on an error dashboard. Promote to `logger.error` (the `exc_info=True` is already correct here). _(silent-failure-hunter, M3)_

### 💡 Good to have (6)

- **Redundant scheme-default swallow** — `payments_request_handler.py:501-506`. `resolve_scheme` already swallows fetch failures internally and returns `"nvm:erc4337"` with a debug log, so this outer `try/except` only masks *unexpected* errors (silently defaulting the scheme) with no log. Drop it, or `logger.warning(..., exc_info=True)`. _(silent-failure-hunter, M2)_
- **Streaming in-band settlement failure is not suppressed** — `on_message_send_stream` (`payments_request_handler.py:805-847`) settles inline and swallows settle failures with a `logger.warning`, never calling `_apply_inband_settlement`; events are already yielded. The `server.py:277-280` comment explicitly scopes the in-band lifecycle to non-streaming `message/send` and this mirrors the legacy header behaviour, so it's arguably out of scope — but the "paid content suppressed on failure" guarantee does **not** hold for the streaming transport. Worth a docs caveat or a follow-up issue. _(code-reviewer + silent-failure-hunter, ~confidence 60)_
- **`verify_error` is a stringly-typed key duplicated across producer/consumer** — `server.py:402` vs `payments_request_handler.py:345`, with no shared constant; a typo silently disables the entire in-band failure path (fails open to execution). It's also honored only when `inband=True`, so a `verify_error` set with `inband=False` is dropped. Promote to a typed `verify_error: Optional[str]` field on `HttpRequestContext` (cheapest high-value cleanup). _(type-design-analyzer, MEDIUM)_
- **`HttpRequestContext` models `(inband, bearer_token, validation)` as a loose tuple** admitting illegal combinations; the load-bearing "force blocking" safety rule and four handler branches rely on convention across three construction sites, with `inband` defaulting to `False` if a site forgets it. A `__post_init__` assertion now (or a discriminated union as a follow-up) would make illegal states unrepresentable. _(type-design-analyzer)_
- **`resolve_token_for_message` returns an unnamed `(Optional[str], bool)`** and is duplicated by the inline middleware logic (`server.py:323-324`) that production actually uses — the tested helper isn't the source of truth. A `NamedTuple` + calling the helper from the middleware would fix both. _(type-design-analyzer, LOW)_
- **Test coverage additions** _(pr-test-analyzer)_: dict-shape input to `_coerce_settle_response` (the third shape, currently untested — `:682-683`); a *non-raising* `settle(success=False)` integration case (second failure mode); `extract_inband_token` encode-failure→`None` (`inband.py:99-104`); and `validate_request` deriving plan from `accepted.planId` with a plan-less card (`:211-219`, the new v2 branch, never reached by current tests).

### ✅ Strengths

- Core "never deliver unpaid content" is implemented on the non-streaming path (forces `blocking=True` for in-band paid requests, drops artifacts/history on settle failure) and well-tested for the **artifact** vector — `_ArtifactDummyExecutor` makes the suppression assertion non-tautological.
- `extract_inband_token` fails **closed** and logs with `exc_info=True` before returning `None` — a malformed payload degrades to "unpaid", the safe direction.
- Body is restored (`request._body = body`) on every `call_next` path; the two early returns that skip `call_next` correctly don't need it.
- `encode_access_token` round-trip for the in-band payload is byte-safe (EIP-712 sig lives inside `payload`), with a round-trip test locking it in.
- Idempotent agent-card extension build (`has_official` guard) + dual-route discovery (canonical `agent-card.json` + legacy `agent.json`), both tested.
- Precise call-count assertions on the mock facilitator prove *when* verify/settle run across all four states; explicit legacy-header non-regression test guards the transport boundary.
- `SettleResponse` itself is a clean pydantic type with sensible aliases/defaults.

---
_Comment-only review; no approve/request-changes. Verdict lives in the text above._

aaitor:
Thanks for the review — addressed in ec9c729:

- **Status-message leak** — on in-band settlement failure, `_apply_inband_settlement` cleared `artifacts` + `history` but left the agent's paid reply in `status.message.parts` (`record_payment_failure` only adds metadata). It now replaces the status message outright, so paid content can't surface there.
- **`_coerce_settle_response` fail-open** — a duck-typed result with no `success` attribute now defaults to `False` (never claim a settlement landed unless the result says so).
- Strengthened the settlement-failure test to assert the **status-message** content is suppressed (not just artifacts), and replaced `ponytail:` markers with `TODO(a2a):`.

r-marques (approved):

PR #239, opened 2026-06-18
docs(mcp): plan-centric MCP server config + fix a2a link-rot

r-marques:
## Summary
Source-doc fixes in `docs/api/` so the next release's auto-synced docs PR is correct and passes link validation. The docs site's `api-reference/python/**` is **generated from this repo**, so these belong here, not hand-edited in the docs repo.

## Changes
- **`docs/api/09-mcp-integration.md`** — plan-centric refinements (the agentId-optional examples + options table were already current on `main`; this fills the remaining gaps):
  - Handler Options table: `planId` reworded as a per-handler override on top of a required server-level `planId`; `onRedeemError` clarified (on post-execution settlement failure, tool content is always suppressed — a paid result is never delivered without settlement).
  - In-band request section: added a session-auth note — the MCP session is an OAuth-protected resource, so the client must send an `Authorization: Bearer <access_token>` header to establish the session (`initialize` returns `401` without it); the payment is the separate in-band `_meta["x402/payment"]` channel.
- **`docs/api/10-a2a-integration.md`** — fixed link-rot: the relative `../../payments_py/x402/README.md` link resolves in the repo but 404s once synced to the docs site. This is exactly why the v1.15.0 release docs PR (nevermined-io/docs#234) fails the Mintlify **link-rot** check. Replaced with an absolute GitHub URL (valid in both mkdocs and the site).

## Why now
After this merges + a patch release, the release bot's docs sync PR will carry the corrected MCP api-reference **and** pass link-rot — no hand-editing of the generated `api-reference/**` in the docs repo.

### DIFF
--- a/docs/api/09-mcp-integration.md
+++ b/docs/api/09-mcp-integration.md
@@ -322,15 +322,15 @@ asyncio.run(main())
 | Option | Type | Description |
 |--------|------|-------------|
 | `credits` | `int` or `callable` | Credits to consume per call |
-| `planId` | `str` | Optional override for the plan ID (otherwise inferred from token) |
+| `planId` | `str` | Per-handler plan ID override. A server-level `planId` (set via `configure`/`start`) is required; set this only to charge a different plan for this handler. |
 | `maxAmount` | `int` | Max credits to verify during authentication (default: `1`) |
-| `onRedeemError` | `str` | `"ignore"` (default) or `"propagate"` to raise on redemption failure |
+| `onRedeemError` | `str` | On post-execution settlement failure: `"ignore"` (default) returns the in-band payment error; `"propagate"` raises a JSON-RPC error. Tool content is always suppressed either way (a paid result is never delivered without settlement). |
 
 ## In-band x402 signaling (`_meta`)
 
 The MCP transport follows the [x402 v2 MCP transport specification](https://github.com/coinbase/x402/blob/main/specs/transports-v2/mcp.md): payments are signalled **in band** through the MCP tool-call machinery, not via HTTP status codes or headers.
 
-**Request — payment payload.** The client sends the x402 `PaymentPayload` as plain JSON in the tool-call request params under `_meta["x402/payment"]`:
+**Request — payment payload.** The client sends the x402 `PaymentPayload` as plain JSON in the tool-call request params under `_meta["x402/payment"]`. This is the **payment** channel — separate from session auth: the MCP session is an OAuth-protected resource, so the client must also send an `Authorization: Bearer <access_token>` header when opening the transport to establish the session (`initialize` returns `401` without it).
 
 ```json
 {
--- a/docs/api/10-a2a-integration.md
+++ b/docs/api/10-a2a-integration.md
@@ -212,7 +212,7 @@ payment-required: eyJ4NDAy... (base64-encoded X402PaymentRequired)
 
 ## In-Band x402 v2 Payments (Standards Flow)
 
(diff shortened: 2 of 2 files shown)


eruizgar91 (commented):
## Review — two new claims in `09-mcp-integration.md` contradict the Python SDK

The a2a link-rot fix and the new 401 / session-auth note are correct (verified against `payments_py/mcp/http/oauth_metadata.py:248-255`). Two of the new MCP claims, however, don't match the code in `payments_py/mcp/` — and since this PR exists to make the auto-synced api-reference *correct*, they'd ship new inaccuracies.

### 1. `planId` — "required server-level" is not how this SDK works
> "A server-level `planId` (set via `configure`/`start`) is required; set this only to charge a different plan for this handler."

Neither `configure` nor `start` accepts a `planId`:
- `PaymentsMCP.start()` config (`payments_mcp.py:557-566`) carries `agentId`, `serverName`, `version`, `description`, `baseUrl`, `onLog` — no `planId`.
- The paywall misconfiguration guard (`core/paywall.py:75`) checks only `agentId`.
- `plan_id` is resolved per-request in `core/auth.py:48-71`: handler override → token's `accepted.planId` → agent's plans fallback.

The handler `planId` is an override on top of the **token-resolved** plan, not on top of a required server-level setting. Suggest anchoring the wording to "the plan inferred from the access token" (the old wording was closer).

### 2. "Tool content is always suppressed either way" is false on the default path
> `"ignore"` (default) returns the in-band payment error … Tool content is always suppressed either way (a paid result is never delivered without settlement).

On the default `ignore` path the content is **delivered**, not suppressed: `core/paywall.py:226-244` takes the handler `result`, adds `_meta` (`success: False`, `errorReason`) on settlement failure, and `return result`. There is no suppression layer downstream. Only `propagate` (`paywall.py:348-353`) withholds content, by raising a JSON-RPC error.

- `ignore` (default): content delivered + `_meta.success=False` + `errorReason`.
- `propagate`: content not returned (error raised).

This is the more important one — it states a payment-safety guarantee the code doesn't make.

### Caveat
If these docs are meant to mirror the TS SDK rather than describe `payments_py`, that's a deliberate divergence — but per `CLAUDE.md` `docs/api/` documents the Python SDK, so as written they're inaccurate for this repo. Holding approval on these two; the link fix + 401 note are good as-is.

r-marques:
Thanks for the careful review @eruiz — and you're right on the a2a link-rot fix and the 401/session-auth note.

On the two MCP claims, I re-checked against current `origin/main` (payments-py `1a02a6c`) and I believe both reflect a **pre-agentId-optional revision** — the line numbers you cited are ~16 lines off from what's there now, which fits a stale local checkout (easy trap to hit). Current main:

**1. `planId` is a required server-level config.**
- `payments_py/mcp/payments_mcp.py:559-566` — `start()` raises `ValueError("plan_id is required …")` if unset and builds `config = { …, "planId": self.plan_id, "agentId": self.agent_id, … }`; the constructor takes a keyword-only `plan_id` (docstring: "required at start").
- The misconfiguration guards check **planId**, not agentId: `core/server_manager.py:372-375` (`# planId is required; agentId is optional` → `if not config.get("planId"): raise ValueError("planId is required")`) and `core/paywall.py:91-97` (`effective_plan_id = options.get("planId") or self.config.get("planId")` → raises `"Server misconfiguration: missing planId"`).
- The per-request resolution in `core/auth.py` is a fallback layer **on top of** the required server-level planId (its own comment: *"planId is the required, configured source of truth for the accepts list"*). So the doc wording — "a server-level `planId` is required; the handler option is a per-handler override" — matches.

**2. Tool content is suppressed on the default `ignore` path.**
- `core/paywall.py:225-229` **unconditionally** `raise SettlementFailedError(...)` on settlement failure (comment: *"do NOT return the tool's content … a paid result is never delivered without payment landing"*). `onRedeemError` only controls whether the inner `_redeem` raises (`propagate`) or returns `{success: False}` (`ignore`, lines 430-442) — the outer guard suppresses either way.
- This mirrors the TS behavior you (correctly) confirmed on #398 (`paywall.ts:219-228`): both SDKs have converged to always-suppress. (One real caveat not in this doc claim: the streaming path can't retract already-yielded content — tracked separately.)

Could you re-check against current main? If you're seeing a different revision that'd explain the gap — happy to adjust if I've misread anything. The a2a fix + 401 note stand as-is.

eruizgar91 (approved):

eruizgar91 (approved):

PR #245, opened 2026-06-18
ci(docs): catch docs link-rot before release (staged link-check gate)

### LINKED ISSUE #241: Catch docs link-rot before release: staged link-check CI gate (payments-py)
Part of epic nevermined-io/nvm-monorepo#2004. Sibling (TS side): nevermined-io/payments#401. **Scope: `payments-py` only.**

## Problem
The docs site `nevermined-io/docs` generates `docs/api-reference/python/**` from this repo's `docs/api/**` via the release bot. A link that is **valid in this repo** (mkdocs / repo-relative path) can be **dead on the site** once synced — e.g. `[x402 README](../../payments_py/x402/README.md)`. Today this is only caught **after release**, when the bot's docs sync PR fails Mintlify link-rot (it bit nevermined-io/docs#234 for v1.15.0). By then the SDK is published and the docs PR is stuck.

## Goal
Fail fast: catch link-rot at PR/release time **in this repo**, before a release ships docs that break on the site.

## Design (key constraints)
1. **Replicate the SITE's resolution, not the repo's.** A plain `mkdocs build` here would PASS and miss it — the link resolves in-repo. The job must first apply the bot's sync transform (`docs/api/*.md` → the `api-reference/python/...` layout, frontmatter, renames) into a staged temp tree, then check links against that staged tree.
2. **PR-time, with a release-time backstop.** Trigger on changes to `docs/api/**` so a bad link never merges; also run in the release pipeline as a safety net.
3. **Reuse the real checker.** Run the same `mintlify broken-links` the docs repo uses (pinned to the same version) against the staged tree — don't reimplement (a hand-rolled checker drifts from Mintlify).
4. **Hard-gate on internal links only.** Mintlify also pings external URLs (network-flaky); gating a release on that causes spurious failures. Hard-gate internal/escaping-link errors; treat external-URL liveness as non-blocking/periodic.
5. **Cheap deterministic lint as the first gate.** Reject any relative link in a synced doc that escapes the synced tree (`](../../…)`, links pointing at repo source like `payments_py/...`). Catches the exact class that broke #234, runs in ms, no network, never flakes — ship it even before the full staged-Mintlify check.
6. **Pair with an authoring rule (prevention).** Synced docs use absolute GitHub URLs or site-relative links, never repo-relative paths. Add to CONTRIBUTING.

## Scope (payments-py)
- [ ] Escaping-link lint on `docs/api/**` (reject `](../../…)` / links to repo source) — PR CI hard gate.
- [ ] Staged sync transform (`docs/api/*.md` → site `api-reference/python/` layout) + `mintlify broken-links` (internal-only) — PR CI + release-pipeline backstop.
- [ ] CONTRIBUTING note: synced docs (`docs/api/**`) use absolute GitHub URLs / site-relative links, never repo-relative.

`docs/api/**` is currently clean (`grep -E '\]\(\.\.?/'` → none; the offender was fixed in #239), so this is **preventive** — it stops the class from recurring.

## Context
- Incident: nevermined-io/docs#234 (v1.15.0 sync) failed Mintlify link-rot on `api-reference/python/a2a-module.mdx` → `../../payments_py/x402/README.md`.
- Source fix: nevermined-io/payments-py#239 (absolute GitHub URL).
- Related packaging gap: nevermined-io/payments-py#240.

r-marques:
## Description

Add a CI gate that catches docs **link-rot before release** — the class of link
that resolves inside this repo but is **dead once `docs/api/**` is synced to the
docs site**. Today this only fails *after* release, when the bot's docs-sync PR
trips Mintlify link-rot (it bit `nevermined-io/docs#234` for v1.15.0:
`a2a-module.mdx` → `../../payments_py/x402/README.md`). The tree is currently
clean, so this is **preventive** — it stops the class from recurring.

Three layers:

1. **`scripts/lint_doc_links.py`** — deterministic, network-free escaping-link
   lint (**hard PR gate**). Flags any relative link in `docs/api/**` that the
   converter won't rewrite (escapes the synced tree, or basename not a chapter
   in `convert_to_mintlify.LINK_MAPPING`). Runs in ms, never flakes. Driven off
   `LINK_MAPPING`, so it can never drift from the transform it guards.
2. **`scripts/check_synced_doc_links.sh`** — converts via the real
   `convert_to_mintlify.py`, drops the pages into a clone of the docs site, and
   runs the same `mintlify broken-links` the docs repo uses (**internal links
   only** — no `--check-external`; external liveness is network-flaky). The
   Python pages legitimately link cross-section via site-relative `/docs/...`
   paths, so a python-only mini-site false-positives on the clean tree; this
   stages against the **whole** site. Blocking PR backstop (source-scoped to
   the staged python pages, so pre-existing site breakage elsewhere is ignored).
3. **`.github/workflows/docs-link-check.yml`** — lint (hard gate) +
   staged-Mintlify (blocking, source-scoped to the python pages) on
   `docs/api/**` changes; and
   **`publish-mintlify-docs.yml`** runs `mintlify broken-links` on the staged
   site before opening the docs PR (release-time backstop).

Plus **`CONTRIBUTING.md`** with the authoring rule: synced docs use chapter
links / site-relative paths / absolute GitHub URLs, never repo-relative.

Mintlify is pinned to `4.2.629` for reproducibility (the docs repo installs
floating latest; bump when that materially changes).

Verified: lint + staged check both pass on the current clean tree and both fail
(exit 1) when the `#234` link is injected.

Part of epic nevermined-io/nvm-monorepo#2004.

Closes #241.

## Types of changes

CI-only / docs-tooling change — no SDK source touched.

## Checklist:

- [x] Follows the code style of this project. (`black` clean; pre-commit passes)
- [x] Tests Cover Changes — both scripts verified against clean + injected-bad trees
- [x] Documentation — CONTRIBUTING.md authoring rule added

#### Funny gif

dead link

### DIFF
--- a/scripts/check_synced_doc_links.sh
+++ b/scripts/check_synced_doc_links.sh
@@ -31,7 +31,7 @@
 # not fail this gate. (See the scoped-parse step at the end of this script.)
 #
 # Env knobs (all optional):
-#   DOCS_REPO       default nevermined-io/docs
+#   DOCS_REPO       default nevermined-io/docs_mintlify
 #   DOCS_REF        default main
 #   MINTLIFY_VERSION default 4.2.629 (pin; the docs repo tracks latest)
 #   DOCS_CHECKOUT   pre-cloned docs repo to reuse instead of cloning
@@ -121,6 +121,10 @@ cd "$DOCS_DIR"
 REPORT="$WORK_DIR/broken-links.txt"
 set +e
 "${MINTLIFY[@]}" broken-links 2>&1 | sed 's/\x1b\[[0-9;]*[A-Za-z]//g' | tr -d '\r' > "$REPORT"
+# The pipe's exit status is tr's (always 0), so grab mintlify's REAL exit code
+# from PIPESTATUS — used below to fail closed if mintlify crashed without
+# producing a parseable report.
+mintlify_rc="${PIPESTATUS[0]}"
 set -e
 
 cat "$REPORT"
@@ -130,7 +134,7 @@ echo ""
 # Output shape (one source block per file with broken links):
 #   docs/api-reference/python/a2a-module.mdx
 #    ⎿  ../../payments_py/x402/README.md
-SCOPE_PREFIX="$SCOPE_PREFIX" python3 - "$REPORT" <<'PY'
+SCOPE_PREFIX="$SCOPE_PREFIX" MINTLIFY_RC="$mintlify_rc" python3 - "$REPORT" <<'PY'
 import os, re, sys
 
 prefix = os.environ["SCOPE_PREFIX"]
@@ -158,6 +162,20 @@ with open(sys.argv[1], encoding="utf-8") as fh:
             total += 1
             if current.startswith(prefix):
                 scoped.append((current, b.group(1)))
+
+# Fail-closed guard: a non-zero mintlify exit with no broken-links header means
+# mintlify did not actually run (npx install failure, exec error, OOM) rather
+# than "ran and found links". Without this, an empty report → reported=None →
+# total==0 → the script would print "✓ no broken links" and exit 0 (fail-OPEN).
+# This is the sibling of the missing-docs.json / empty-dir / format-drift guards.
+rc = int(os.environ["MINTLIFY_RC"])
+if rc != 0 and reported is None:
+    print(
+        f"✗ mintlify broken-links produced no parseable report (exit {rc}) — "
+        "it likely failed to run. Failing closed.",
+        file=sys.stderr,
+    )
+    sys.exit(2)
 
 # False-green guard: the checker reported broken links but our parser attributed
 # none to any source — the output format drifted (e.g. a mintlify bump changed
--- a/scripts/check_synced_doc_links.sh
+++ b/scripts/check_synced_doc_links.sh
@@ -23,11 +23,19 @@
 # by default and only pings external URLs with --check-external (NOT passed —
 # external liveness is network-flaky and must not block a release).
 #
+# Scoped to OUR breakage: the whole site may carry pre-existing broken links we
+# don't own. We replace ALL python pages with freshly-converted ones, so any
+# broken link whose SOURCE file is under docs/api-reference/python/ is breakage
+# introduced by these staged pages. We parse the checker output and fail ONLY on
+# python-sourced broken links — pre-existing breakage elsewhere on the site does
+# not fail this gate. (See the scoped-parse step at the end of this script.)
+#
 # Env knobs (all optional):
 #   DOCS_REPO       default nevermined-io/docs
 #   DOCS_REF        default main
 #   MINTLIFY_VERSION default 4.2.629 (pin; the docs repo tracks latest)
 #   DOCS_CHECKOUT   pre-cloned docs repo to reuse instead of cloning
+#   SCOPE_PREFIX    site path whose broken links we own (default the python tree)
 set -euo pipefail
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
@@ -39,6 +47,8 @@ DOCS_REF="${DOCS_REF:-main}"
 # Pin Mintlify for reproducibility. The docs repo itself installs floating
 # latest (npm i -g mintlify); bump this when that materially changes.
 MINTLIFY_VERSION="${MINTLIFY_VERSION:-4.2.629}"
+# Broken links whose source file starts with this site path are ours to fix.
+SCOPE_PREFIX="${SCOPE_PREFIX:-docs/api-reference/python/}"
 
 if [ ! -d "$SOURCE_DIR" ]; then
   echo "Error: source docs not found at $SOURCE_DIR" >&2
@@ -92,5 +102,72 @@ fi
 echo "Running mintlify@$MINTLIFY_VERSION broken-links (internal links only) on the staged site …"
 echo ""
 cd "$DOCS_DIR"
-# Exits non-zero when it finds broken internal links.
-"${MINTLIFY[@]}" broken-links
+
+# Capture the full report (strip ANSI/spinner CRs), then scope to our pages.
+# Do not let mintlify's own non-zero exit abort the script — we decide pass/fail
+# from the SCOPED parse, since pre-existing site breakage must not fail us.
+REPORT="$WORK_DIR/broken-links.txt"
+set +e
+"${MINTLIFY[@]}" broken-links 2>&1 | sed 's/\x1b\[[0-9;]*[A-Za-z]//g' | tr -d '\r' > "$REPORT"
+set -e
+
+cat "$REPORT"
+echo ""
+
+# Parse the report and fail ONLY on broken links sourced from our staged pages.
+# Output shape (one source block per file with broken links):
+#   docs/api-reference/python/a2a-module.mdx
+#    ⎿  ../../payments_py/x402/README.md
+SCOPE_PREFIX="$SCOPE_PREFIX" python3 - "$REPORT" <<'PY'
+import os, re, sys
+
+prefix = os.environ["SCOPE_PREFIX"]
+source_re = re.compile(r"^(\S+\.mdx)\s*$")
+broken_re = re.compile(r"^\s*⎿\s+(.+?)\s*$")
+# Header the checker prints, e.g. "found 2 broken links in 2 files".
+header_re = re.compile(r"found\s+(\d+)\s+broken\s+links?\b")
+
+reported = None  # broken-link count from the checker's own summary line
+total = 0  # broken links our parser attributed to ANY source file
+scoped, current = [], None
+with open(sys.argv[1], encoding="utf-8") as fh:
+    for line in fh:
+        line = line.rstrip("\n")
+        h = header_re.search(line)
+        if h:
+            reported = int(h.group(1))
+            continue
+        m = source_re.match(line)
+        if m:
+            current = m.group(1)
+            continue
+        b = broken_re.match(line)
+        if b and current:
+            total += 1
+            if current.startswith(prefix):
+                scoped.append((current, b.group(1)))
+
+# False-green guard: the checker reported broken links but our parser attributed
+# none to any source — the output format drifted (e.g. a mintlify bump changed
+# the glyph/layout). Fail loudly rather than pass silently on an unparsed report.
+if reported and reported > 0 and total == 0:
+    print(
+        "✗ mintlify reported broken links but this script parsed none — the "
+        "broken-links output format has likely changed. Update the parser in "
+        "scripts/check_synced_doc_links.sh (the source/⎿ line patterns).",
+        file=sys.stderr,
+    )
+    sys.exit(2)
+
+if scoped:
+    print(f"✗ {len(scoped)} broken internal link(s) introduced by the staged "
+          f"pages ({prefix}):")
+    for src, target in scoped:
+        print(f"  {src} -> {target}")
+    print("\nThese links resolve in-repo but are dead on the docs site. Use a "
+          "chapter link the converter rewrites, a site-relative /docs/... path, "
+          "or an absolute https://github.com/... URL. See CONTRIBUTING.md.")
+    sys.exit(1)
+
+print(f"✓ No broken internal links sourced from the staged pages ({prefix}).")
+PY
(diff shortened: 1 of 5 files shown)


r-marques (commented):
## 🤖 Automated PR review — 🟡 Mergeable (minor fail-closed hardening + one required admin follow-up)

_Reviewed `origin/main...HEAD` @ `c9cad38` (5 files) with a 2-agent panel (code-reviewer + silent-failure-hunter), scoped to the diff, honoring the repo CLAUDE.md._

Well-engineered. `is_violation()` is correct across every canonical case (allows chapter links / `/docs/...` / `#anchors` / absolute URLs; rejects `../` escaping + repo-source paths), and the `LINK_MAPPING`-driven lint cannot drift from the converter it guards. The `c9cad38` rework is the right design: the staged Mintlify check is **source-scoped** — it parses the report and fails only on broken links whose source is under `docs/api-reference/python/`, so pre-existing breakage elsewhere on the site no longer couples to this gate — and it now runs the **same scoped script** at PR-time and as the release backstop. Empirically verified by the author 4 ways (clean→0; injected escaping link→1; pre-existing non-python break→ignored; format-drift→2). No correctness blockers.

### ⚠️ Operational follow-up (required before the gate enforces — repo admin)
- **`lint-links` is not a required status check.** `main` branch protection requires only `e2e` / `lint` / `unit_integration`; the new `lint-links` / `staged-mintlify-check` jobs aren't in that set, so a red `lint-links` won't block a merge — the gate is advisory until added to required checks. This is a repo-settings change (can't land in the diff). _(silent-failure-hunter, 0.85)_

### 🟡 Should fix (cheap, fail-closed hardening on the gate)
- **`mintlify broken-links` no-ops to success (exit 0) when no `docs.json`/`mint.json` is present** (verified). The new false-green guard only catches the "header says N>0 but parser attributed 0" case — a *missing config entirely* gives no broken-links header at all, so it slips through green. The `TARGET_DIR` guard checks `api-reference/python` exists but not the site config. Add a symmetric assertion that `docs.json` (or `mint.json`) exists at `$DOCS_DIR` before running, so a docs-site layout change fails closed. _(silent-failure-hunter, 0.9)_
- **The deterministic lint misses reference-style (`[t][ref]` + `[ref]: ../escape.md`) and HTML (`<a href>`) links** — an escaping link in those forms passes silently. None exist in `docs/api/**` today. Extend `LINK_RE` to also scan reference-style link definitions (the realistic future form), or document the inline-only scope in a comment. _(silent-failure-hunter 0.8 / code-reviewer 0.9)_

### 💡 Good to have
- **Lint vacuously passes on an empty `docs/api/`** (dir present, no `.md` → exit 0). A `total_files == 0` guard makes it fail closed. _(silent-failure-hunter, 0.9)_
- **`docs` vs `docs_mintlify` naming.** The script defaults `DOCS_REPO=nevermined-io/docs`; the release workflow uses `nevermined-io/docs_mintlify`. Both resolve to the same SHA via a GitHub alias today — standardize on the canonical `docs_mintlify` to avoid silent breakage if the alias goes away. _(code-reviewer, 0.95)_
- **Lint `glob("*.md")` is non-recursive** while the trigger watches `docs/api/**` (and the converter is flat-only) — agree the three or note the flat-only scope. Moot today. _(silent-failure-hunter, 0.7)_
- _Adjacent, pre-existing:_ `publish-mintlify-docs.yml` still checks `-lt 12` / "12 files" while there are 13 chapters — harmless (13 ≥ 12); optional one-line bump since the PR touches that file.

### ✅ Strengths
- The source-scoping + **false-green guard** (exit 2 on output-format drift) is exactly the right defense for a checker whose output you parse — it fails loud rather than passing silently on an unparsed report.
- One shared scoped script for PR + release → no PR/release divergence.
- `set +e` around the mintlify call with pass/fail decided by the scoped parse; `set -euo pipefail` elsewhere; `cp *.mdx` with zero matches fails closed under `set -e`; EXIT trap doesn't mask the exit code (all verified).
- Mintlify pinned (`4.2.629`) with a bump comment; reuses a matching global CLI before falling back to pinned `npx`.
- Thorough rationale in comments + `CONTRIBUTING.md` (the `nevermined-io/docs#234` incident). `pyproject.toml` untouched; protects the generated-from-SDK pipeline rather than editing generated files.

scripts/check_synced_doc_links.sh
```diff
@@ -0,0 +1,185 @@
...
+echo "Running mintlify@$MINTLIFY_VERSION broken-links (internal links only) on the staged site …"
+echo ""
+cd "$DOCS_DIR"
+
+# Capture the full report (strip ANSI/spinner CRs), then scope to our pages.
+# Do not let mintlify's own non-zero exit abort the script — we decide pass/fail
+# from the SCOPED parse, since pre-existing site breakage must not fail us.
+REPORT="$WORK_DIR/broken-links.txt"
+set +e
+"${MINTLIFY[@]}" broken-links 2>&1 | sed 's/\x1b\[[0-9;]*[A-Za-z]//g' | tr -d '\r' > "$REPORT"
```
aaitor, line 123: **MEDIUM — the staged check passes green if `mintlify` itself fails to run (fail-open hole).**

`mintlify`'s exit code is discarded here — the pipe's status is `tr`'s (always 0) — and pass/fail is decided solely from the scoped parse. That's correct for *"mintlify ran and found links we don't own."* But if `mintlify` **crashes** (npx install failure, exec error, OOM), the report is empty: `reported` stays `None`, `total == 0`, `scoped == []`, so the script prints `✓ No broken internal links` and exits **0**. The false-green guard at L165 only fires when a `found N>0` header was printed — a crash prints no header, so it slips through green.

This is the same fail-closed class you already hardened three other ways (missing `docs.json` L88, empty `docs/api/` in the lint, format-drift L165) — this sibling case is the one left open. It bites hardest at **release time** (`publish-mintlify-docs.yml`), where this is a hard step: a flaky `npx mintlify` install would let the release proceed with the backstop silently never having run.

Capture mintlify's real exit and fail closed when it produced no parseable report:

```bash
REPORT="$WORK_DIR/broken-links.txt"
set +e
"${MINTLIFY[@]}" broken-links 2>&1 | sed 's/\x1b\[[0-9;]*[A-Za-z]//g' | tr -d '\r' > "$REPORT"
mintlify_rc="${PIPESTATUS[0]}"
set -e
```

Then pass it into the heredoc (add `MINTLIFY_RC="$mintlify_rc"` to the `python3 - "$REPORT"` env, next to `SCOPE_PREFIX`) and add a crash guard beside the existing false-green guard:

```python
rc = int(os.environ["MINTLIFY_RC"])
# Fail closed: a non-zero exit with no broken-links header means mintlify did
# not run (install/exec failure), not "found links" — never pass green on that.
if rc != 0 and reported is None:
    print(f"✗ mintlify broken-links produced no parseable report (exit {rc}) — "
          "it likely failed to run. Failing closed.", file=sys.stderr)
    sys.exit(2)
```

scripts/check_synced_doc_links.sh
```diff
@@ -0,0 +1,185 @@
...
+#
+# Scoped to OUR breakage: the whole site may carry pre-existing broken links we
+# don't own. We replace ALL python pages with freshly-converted ones, so any
+# broken link whose SOURCE file is under docs/api-reference/python/ is breakage
+# introduced by these staged pages. We parse the checker output and fail ONLY on
+# python-sourced broken links — pre-existing breakage elsewhere on the site does
+# not fail this gate. (See the scoped-parse step at the end of this script.)
+#
+# Env knobs (all optional):
+#   DOCS_REPO       default nevermined-io/docs
```
aaitor: **LOW — stale env-knob comment.** This line documents `DOCS_REPO default nevermined-io/docs`, but the actual default (L48) is now `nevermined-io/docs_mintlify`. Sync it so the canonical repo is documented consistently:

```suggestion
#   DOCS_REPO       default nevermined-io/docs_mintlify
```

aaitor (approved):
APPROVE on the diff's merits — a well-built, internally-consistent gate. The deterministic `lint_doc_links.py` (the real always-on guard) is correct across every canonical case I traced (allows chapter links / `./x.md#a` / `/docs/...` / `#anchor` / absolute URLs; flags `../` escapes, repo-source paths, and chapter links not yet in `LINK_MAPPING`), and the prior-review findings (empty-dir vacuous pass, ref-style/HTML links, `docs.json`-missing, `docs`→`docs_mintlify` default) are all genuinely fixed at `6c02663`. The release backstop in `publish-mintlify-docs.yml` is wired correctly: `docs_mintlify` is checked out at L64 → `DOCS_CHECKOUT` resolves, and the script runs against a `cp -r` copy so it never disturbs the tree `create-pull-request` publishes.

One **MEDIUM** worth closing *before* `staged-mintlify-check` is promoted to a required check: the staged check fails **open** if `mintlify` itself crashes (empty report → `reported=None` → the L165 false-green guard doesn't fire → exit 0). It's the one sibling of the fail-closed class you hardened everywhere else. Inline fix below (PIPESTATUS + a crash guard). Not a merge blocker — the deterministic lint is a separate, network-free job and still catches the `#234` class regardless.

**Verification I ran**
- Read all 5 changed files + `convert_to_mintlify.py` (the `LINK_MAPPING` the lint imports) at `6c02663`; hand-traced `is_violation()` over ~10 canonical/edge destinations — correct on each.
- Confirmed the release backstop's `DOCS_CHECKOUT` path matches the L64 `docs_mintlify` checkout, and the script copies before mutating (no interference with PR creation).
- CI: `lint`, `unit_integration`, `lint-links` (5s), `staged-mintlify-check` (1m11s) all green. `e2e` is **red**, but Build (step 7) passed and only the staging-backed test-execution step (step 8) failed, on a diff touching **zero** SDK code → unrelated staging flakiness (CLAUDE.md's E2E-vs-staging gotcha), not this PR. `e2e` is a required check, so it needs a re-run to actually merge.
- Over-engineering / leanness pass (`/ponytail-review`): **lean already, ship.** Net removable ≈ 0 lines. The lint's ref-style/HTML coverage is cheap belt-and-suspenders on a republish gate (real escaping classes, not speculative abstraction); the 3-layer split covers *distinct* link classes (lint = escaping/unmapped relative; staged = site-relative `/docs/...` liveness the lint waves through). Nothing worth deleting.

**Notes for the author / ops (out of scope for the diff)**
- The PR **description** still says `staged-Mintlify (continue-on-error)`, but the YAML has no `continue-on-error` and `CONTRIBUTING.md` says *"Both gates block."* The code is the consistent one — update the description so it doesn't read as a non-blocking job.
- `staged-mintlify-check` is network-dependent (clone + `npx mintlify`). Defensible as a blocking gate (pinned + internal-only + scoped), but once it's a **required** check a transient npm/clone flake will block merges with only a re-run as recourse. The MEDIUM fix at least makes a *crash* fail loudly instead of silently green.
- Pre-existing, optional since you're already editing `publish-mintlify-docs.yml`: the `-lt 12` guard (L49) and the hard-coded "all 12 documentation files" list (L172–189) are stale at 13 chapters — harmless (13 ≥ 12), but a one-line bump to 13 + adding `langsmith-deployment-module.mdx` to the list stops them drifting further.
- Admin follow-up (already in your CONTRIBUTING admin note): neither `lint-links` nor `staged-mintlify-check` is in `main`'s required checks yet, so the gate is advisory until added.

r-marques:
Addressed @aaitor's post-approval findings in `fa2db2c`:

- **MEDIUM (mintlify-crash fail-open)** — fixed exactly as suggested: capture `mintlify`'s real exit via `PIPESTATUS[0]`, pass `MINTLIFY_RC` into the parser, and fail **closed** (`sys.exit(2)`) when `rc != 0 and reported is None`. Verified across 6 fixtures incl. crash-empty at `rc=1` **and** `rc=127` → both now exit 2 (were 0).
- **LOW (stale `DOCS_REPO` comment)** — synced the env-knob comment (L34) to `nevermined-io/docs_mintlify`.
- **PR description** — removed the inaccurate `staged-Mintlify (continue-on-error)` / "non-blocking" wording; the staged check is blocking + source-scoped, matching the YAML and CONTRIBUTING.
- **Your out-of-scope note** — bumped the stale `-lt 12`→`13` guard, "all 12"→"13", and added `langsmith-deployment-module.mdx` to the file list (confirmed 13 chapters via the converter's `LINK_MAPPING`).

CI: `lint` / `lint-links` / `staged-mintlify-check` (exercises the crash path) / `unit_integration` all green; `e2e` pending — the staging-backed suite on a diff touching zero SDK code (the unrelated flakiness you flagged). Thanks for the careful review.

PR #257, opened 2026-07-31
docs: retire docs.nevermined.app references; follow the docs-site flatten

eruizgar91:
Started as the `payments-py` slice of the `docs.nevermined.app` sweep (nevermined-io/nevermined.ai-website#233). The first push exposed a second, unrelated breakage from the same epic, so this PR carries both.

## 1. Retire `docs.nevermined.app` (the original ask)

`docs.nevermined.app` 301s to `nevermined.ai/docs/*`. The redirect works, so nothing is broken, but these references live in surfaces users **copy** (docstrings, synced docs) rather than click. All five occurrences now point at the final URLs.

Three of the five were dead links, not just stale hosts:

```
https://docs.nevermined.app/api-reference/versioning
  → 301 https://nevermined.ai/docs/api-reference/versioning
  → 404
```

A host-only rewrite would have produced a working URL to a 404. The live page is `https://nevermined.ai/docs/development-guide/api-versioning` (verified 200).

| File | Change |
| --- | --- |
| `docs/index.md` | host swap |
| `payments_py/x402/README.md` | host swap |
| `docs/api/02-initializing-the-library.md` | retarget dead versioning link |
| `payments_py/common/api_version.py` | retarget dead versioning link |
| `CLAUDE.md` | retarget dead versioning link |

`docs/api/02-initializing-the-library.md` is synced to the docs site, and is the source of the last remaining `docs.nevermined.app` occurrence in the published `nevermined.ai/docs/llms-full.txt` — this drops that count from 1 to 0 on the next release.

## 2. Follow the docs-site flatten (found by CI on this branch)

nevermined-io/docs#257 ("flatten `docs/` to the repo root for base-path hosting", also #214) moved the site's pages out of the top-level `docs/` directory on 2026-07-29. This repo still assumed the old layout in two ways. The `staged-mintlify-check` gate had not run since 2026-07-24, so this PR is where it surfaced.

**Target path.** `scripts/check_synced_doc_links.sh` hard-coded `docs/api-reference/python` and hard-failed:

```
Error: nevermined-io/docs_mintlify has no docs/api-reference/python — site layout changed?
```

`publish-mintlify-docs.yml` had the same stale path but `mkdir -p`'d it, so **the release would not have failed** — it would have recreated the dead directory and published the `.mdx` files into it, then opened a docs PR adding pages nothing links to. Both now detect the tree (repo root first, then under `docs/`), and the workflow no longer creates it: a missing tree means the layout moved again and must fail loudly. `SCOPE_PREFIX` follows the detected path, since mintlify reports broken-link sources relative to the project root.

**Link prefix.** With the site no longer nested, internal links dropped their `/docs` prefix — `docs.json` navigates to `api-reference/python/*`, and the site now contains zero `](/docs/...)` links. Every link we emitted was therefore dead: 32 broken links sourced from the staged pages. Rewritten in `convert_to_mintlify.py` `LINK_MAPPING` (13), `docs/api/12-langchain-integration.md` (3), `tests/unit/test_convert_to_mintlify.py` (4 pinned assertions), plus the two docs describing the convention.

## Verification

```
$ grep -rn 'docs\.nevermined\.app' --exclude-dir=.git .   # no matches
$ grep -rn '](/docs/' --exclude-dir=.git .                # no matches
$ python3 scripts/lint_doc_links.py                       # ✓ no escaping links
$ bash scripts/check_synced_doc_links.sh                  # success no broken links found
$ poetry run black --check .                              # 164 files unchanged
$ poetry run pytest -m "not slow"                         # green (see note)
$ poetry run mkdocs build                                 # OK
```

`check_synced_doc_links.sh` was run locally against the real docs site, and passes in CI (`staged-mintlify-check`).

Refs nevermined-io/nevermined.ai-website#233, nevermined-io/docs#257

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,7 +20,7 @@ numbered chapters (`NN-<slug>.md`). On release, `publish-mintlify-docs.yml`
 converts them to Mintlify MDX via
 [`scripts/convert_to_mintlify.py`](./scripts/convert_to_mintlify.py) and opens a
 PR against the docs site (`nevermined-io/docs`), publishing them at
-`docs/api-reference/python/<slug>`.
+`api-reference/python/<slug>`.
 
 When you change a public API, update the matching chapter in `docs/api/` (see
 [`CLAUDE.md`](./CLAUDE.md) for which files map to which APIs) and rebuild with
@@ -38,7 +38,7 @@ In `docs/api/**`, use only:
 
 - ✅ **Chapter links** to other `docs/api/` pages, by filename — `[x402](11-x402.md)`
   or `[x402](./11-x402.md#section)`. The converter rewrites these to the site URL
-  (`/docs/api-reference/python/x402-module`). The set of rewritable chapters is
+  (`/api-reference/python/x402-module`). The set of rewritable chapters is
   `LINK_MAPPING` in [`scripts/convert_to_mintlify.py`](./scripts/convert_to_mintlify.py);
   add a chapter there when you add a page.
 - ✅ **Site-relative links** to other docs-site pages — `[LangChain](/integrate/add-to-your-agent/langchain)`.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -41,7 +41,7 @@ In `docs/api/**`, use only:
   (`/docs/api-reference/python/x402-module`). The set of rewritable chapters is
   `LINK_MAPPING` in [`scripts/convert_to_mintlify.py`](./scripts/convert_to_mintlify.py);
   add a chapter there when you add a page.
-- ✅ **Site-relative links** to other docs-site pages — `[LangChain](/docs/integrate/add-to-your-agent/langchain)`.
+- ✅ **Site-relative links** to other docs-site pages — `[LangChain](/integrate/add-to-your-agent/langchain)`.
 - ✅ **In-page anchors** — `[Credits](#credits-semantics)`.
 - ✅ **Absolute GitHub URLs** for anything in the repo (source, tests, READMEs,
   other directories) — `[x402 README](https://github.com/nevermined-io/payments-py/blob/main/payments_py/x402/README.md)`.
--- a/scripts/check_synced_doc_links.sh
+++ b/scripts/check_synced_doc_links.sh
@@ -35,7 +35,8 @@
 #   DOCS_REF        default main
 #   MINTLIFY_VERSION default 4.2.629 (pin; the docs repo tracks latest)
 #   DOCS_CHECKOUT   pre-cloned docs repo to reuse instead of cloning
-#   SCOPE_PREFIX    site path whose broken links we own (default the python tree)
+#   SCOPE_PREFIX    site path whose broken links we own (default: the python
+#                   tree, auto-detected from the checkout's layout)
 set -euo pipefail
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
@@ -50,8 +51,9 @@ DOCS_REF="${DOCS_REF:-main}"
 # Pin Mintlify for reproducibility. The docs repo itself installs floating
 # latest (npm i -g mintlify); bump this when that materially changes.
 MINTLIFY_VERSION="${MINTLIFY_VERSION:-4.2.629}"
-# Broken links whose source file starts with this site path are ours to fix.
-SCOPE_PREFIX="${SCOPE_PREFIX:-docs/api-reference/python/}"
+# SCOPE_PREFIX (broken links whose source file starts with this site path are
+# ours to fix) defaults to the python tree, resolved after the checkout — see
+# PYTHON_TREE below.
 
 if [ ! -d "$SOURCE_DIR" ]; then
   echo "Error: source docs not found at $SOURCE_DIR" >&2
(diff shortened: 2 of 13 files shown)


CONTRIBUTING.md
```diff
@@ -41,7 +41,7 @@ In `docs/api/**`, use only:
   (`/docs/api-reference/python/x402-module`). The set of rewritable chapters is
```
aaitor: **MEDIUM — the convention doc still shows the retired `/docs/` prefix here, and now contradicts L44.**

This PR drops the `/docs` prefix everywhere links are *emitted*, but this line — in the file that *defines* the convention — still gives the converter's rewrite target as `/docs/api-reference/python/x402-module`. After this PR the converter emits `/api-reference/python/x402-module` (see the updated `LINK_MAPPING` in `convert_to_mintlify.py`), and L44 two lines down was correctly updated to `/integrate/...`. So the same list now documents two conventions, and a contributor copying the L41 form re-introduces exactly the dead-link class this PR removes.

```suggestion
  (`/api-reference/python/x402-module`). The set of rewritable chapters is
```

scripts/lint_doc_links.py
```diff
@@ -23,7 +23,7 @@
 It intentionally ALLOWS:
   - convertible chapter links: ``](11-x402.md)``, ``](./11-x402.md#anchor)``
     (basename in LINK_MAPPING, no path separator) — rewritten to the site URL.
-  - site-relative paths: ``](/docs/integrate/...)`` — resolve on the site.
+  - site-relative paths: ``](/integrate/...)`` — resolve on the site.
```
aaitor, line 26: **MEDIUM — two more `/docs/...` references in this file were missed (both outside the diff).**

You fixed the `ALLOWS` example on this line, but the retired prefix survives twice more:
- **L18** (docstring): ``a site-relative ``/docs/...`` path`` — stale example.
- **L149** (failure help text): `a site-relative path (/docs/…), or` — this is printed to a contributor *at the moment CI rejects their link*, so it now instructs the dead prefix. And because this lint allows any `/`-leading target (`NON_RELATIVE_PREFIXES`, L63/L74), a `/docs/…` link passes this gate and only fails later at `check_synced_doc_links.sh` — the wrong guidance costs a full CI round-trip.

Fixes:
- L18: `` ``/docs/...`` `` → `` ``/…`` `` (or drop the example)
- L149: `a site-relative path (/docs/…), or` → `a site-relative path (/…), or`

The PR's `grep '](/docs/'` check can't catch these — they're prose, not link syntax. `git grep -n '/docs/' scripts/ CONTRIBUTING.md` surfaces the full set.

scripts/check_synced_doc_links.sh
```diff
@@ -90,11 +92,27 @@ if [ ! -f "$DOCS_DIR/docs.json" ] && [ ! -f "$DOCS_DIR/mint.json" ]; then
...
+done
+if [ -z "$PYTHON_TREE" ]; then
+  echo "Error: $DOCS_REPO has no api-reference/python (looked at the repo root" \
+       "and under docs/) — site layout changed?" >&2
   exit 1
 fi
+TARGET_DIR="$DOCS_DIR/$PYTHON_TREE"
+# Mintlify reports broken-link sources as paths relative to the project root,
+# so the scope prefix tracks whichever layout we just found.
+SCOPE_PREFIX="${SCOPE_PREFIX:-$PYTHON_TREE/}"
```
aaitor, line 114: **MEDIUM — the same stale `/docs/` guidance lives in this script too (both outside the diff).**

- **L16** (header comment): example `/docs/integrate/add-to-your-agent/langchain` — the convention this PR establishes is now `/integrate/...` (and I confirmed `nevermined.ai/docs/integrate/...` → 200 while `nevermined.ai/integrate/...` → 404, so the `/docs/` form as a live URL is fine but as the *site-relative link* contributors write it is now dead).
- **L216** (failure help text): `a site-relative /docs/... path,` — printed when this gate rejects a link, telling the contributor to use the retired prefix.

Fixes:
- L16: `/docs/integrate/add-to-your-agent/langchain` → `/integrate/add-to-your-agent/langchain`
- L216: `a site-relative /docs/... path,` → `a site-relative /... path,`

aaitor (approved):
Tidy, well-motivated cleanup — the URL retargets and the docs-flatten follow-through are all correct, and I verified the load-bearing claims: the old `docs.nevermined.app/api-reference/versioning` redirect really does land on a **404**, the new `nevermined.ai/docs/development-guide/api-versioning` is **200**, and `docs.nevermined.app` / `](/docs/` are genuinely zero on the branch. The layout-detection in both scripts is sound and fails loud.

One non-blocking gap: the PR retired the `/docs` prefix in the links it *emits*, but left it in the prose/comments/help-text that *tell contributors the convention* — 5 stale `/docs/...` references across `CONTRIBUTING.md` and the two scripts, two of them in **failure help-text** that would steer a contributor straight back into the dead prefix. None break the build (the mintlify gate still catches an actual `/docs/…` link, one CI cycle later), so this approves — worth a quick sweep before merge.

**Verification I ran**
- `gh pr checks` → 5/5 green; base is `main` (not a chain), so no parent branch to cross-check.
- URL chain: `docs.nevermined.app/api-reference/versioning` → 301 → `nevermined.ai/docs/api-reference/versioning` → **404** (confirms the retarget was necessary, not cosmetic); `nevermined.ai/docs/development-guide/api-versioning` → **200**; `nevermined.ai/docs` → 200 (308 → `/docs/getting-started/overview`); the `/integrate/...` langchain target → **200** and its `/docs/integrate/...` variant → **404** (confirms the flatten).
- `git grep '](/docs/'` and `git grep 'docs\.nevermined\.app'` on the head branch → **zero** (the link-syntax claims hold exactly as stated).
- `scripts/lint_doc_links.py` allows any `/`-leading target (`NON_RELATIVE_PREFIXES`, L63/L74), so dropping `/docs` does not break that gate; `convert_next_steps_to_cards` derives card `href`s from the now-`/docs`-free `LINK_MAPPING`, so no hardcoded prefix leaks into the generated cards. Test assertions match the new mapping and CI is green.
- Leanness pass (`/ponytail-review`): net removable ≈ 0. The two-candidate detection loops are ~8 lines each but buy real transitional resilience (`DOCS_REF` can point at a pre-flatten ref) and fail loud when neither exists — lean already, ship.
- Broader `git grep '/docs/'` on head surfaced the 5 stale convention references the narrow link-grep missed (inline comments below); the false positives (`docs/api` source dir, the `../../docs/api/...` GitHub-viewed README link) are correct and left alone.

Nothing here gates the release; the three inline comments are consistency fixes to keep the retired prefix from creeping back via contributor-facing guidance.

CONTRIBUTING.md
eruizgar91: Fixed in f910b7a — `/docs/api-reference/python/x402-module` → `/api-reference/python/x402-module`, so the two bullets agree again.

Also swept the same file's L23 (`docs/api-reference/python/<slug>` → `api-reference/python/<slug>`), which describes where the release publishes the pages.

scripts/lint_doc_links.py
eruizgar91: Fixed in f910b7a — both. L18 docstring → ``/...``, and the L149 help text → `(/…)`.

Good catch on the cost: the lint allows any `/`-leading target, so its own error message was sending contributors to a link that passes here and only dies at `check_synced_doc_links.sh` a cycle later.

Also swept the module docstring L5, which named the pre-flatten publish path (`docs/api-reference/python/<slug>.mdx` → `api-reference/python/<slug>.mdx`).

scripts/check_synced_doc_links.sh
eruizgar91: Fixed in f910b7a — L16 → `/integrate/add-to-your-agent/langchain`, L216 help text → `a site-relative /... path`.

Your distinction is the right one: `nevermined.ai/docs/integrate/...` is a live URL, but as the site-relative form a contributor writes in an mdx it is dead — the site now has zero `](/docs/...)` links.

Also swept four more comments in this script that named the pre-flatten site path (L8, L20, L28, and the L153 sample of mintlify output), plus the two in `docs-link-check.yml`. The legacy `docs/api-reference/python` candidate in the detection loop is deliberately untouched — it is what keeps a pre-flatten `DOCS_REF` working.

Full sweep: `git grep -n "/docs/\|docs/api-reference" -- scripts/ CONTRIBUTING.md .github/` is now clean apart from the legacy candidate and the repo's own `docs/api` source dir. Re-ran the gate against the live site: `success no broken links found`. All 5 checks green.

PR #262, opened 2026-08-20
feat(mpp): payments-py parity for the MPP seller + buyer surface

eruizgar91:
Closes nevermined-io/nvm-monorepo#2976.

Python parity for the MPP surface that shipped in `@nevermined-io/payments`
(nevermined-io/payments#417 seller + #418 buyer, released in `v1.11.0`). The
backend is already live and protocol-agnostic, so this is SDK-only: no API
change, no version gate, no new backend surface. `LOCKED_API_VERSION` is
untouched.

## What ships

- **`payments_py/mpp/`** — a dependency-free MPP wire codec plus `payments.mpp`
  for challenge issuance, credential verification and settlement.
- **`payments.mpp.fetch()`** — pays an MPP-challenged endpoint with the
  delegation the buyer already uses for x402.
- **An additive, default-off `mpp` option on `RouteConfig`**, handled by the
  FastAPI middleware. With MPP off the x402 path is unchanged and its tests pass
  unedited.
- **Optional `mpp={"bind_body": True}`** challenge-to-body binding.
- **`docs/api/15-mpp.md`** — seller + buyer guide; `mkdocs build` is green.

## Notes for review

- **No `mppx` dependency, and no new runtime dependency at all.** The challenge
  id is an HMAC over `canonicalize(request)` and `opaque`; the codec carries both
  base64url strings through verbatim, so the HMAC stays valid. The fixtures are
  the same `mppx@0.6.31`-modelled literals the TS suite pins — **copied, not
  re-derived**, so the two SDKs cannot drift apart silently.
- **The settled design decisions from the TS review are carried over as-is**:
  credits sealed at mint time, `settled` reading `receipt.status`
  asymmetrically, a retryable code deciding alone, `credits_presented` as an
  upper bound, and `mpp_spend_of` reporting only when a credential was actually
  presented. Each one is documented at the symbol that implements it.
- **Namespace is `payments.mpp.*`**, symmetric with `payments.x402`. Epic
  nevermined-io/nvm-monorepo#2024 lists that as an MVP non-goal; the TS PR
  deviated deliberately and the deviation was accepted on merge. Python follows
  TS, not the epic text.
- **`get_x402_access_token`'s signature and behaviour are unchanged.** It now
  builds its body through the shared `build_x402_token_request_body`, and its 299
  tests pass unedited. One user-visible string did change on that path, flagged
  here because it is not otherwise reviewable: the inline-create deprecation
  `FutureWarning` no longer names `get_x402_access_token`, because the same
  warning is now reachable from the MPP mint and naming one caller would send an
  MPP buyer grepping for a symbol they never called. The existing tests match on
  `create_delegation`, which is still in the text.

## Three places this is NOT a transliteration

Framework differences, each documented at the module that carries it:

1. **Transport is `requests`, not `fetch`.** A single-read body is a
   generator/iterator/file object rather than a `ReadableStream`, and it is
   refused at the point a retry would reuse it — never before the first request.
   The `burns=True` classification is mapped onto `requests`' exception tree: a
   `ReadTimeout` or a connection torn down after the request was written is an
   UNKNOWN settlement outcome, while a `ConnectTimeout` or a refused connection
   is a definite failure — nothing was written, so nothing can have burned.
2. **The 64 KiB error-body cap bounds what is PARSED**, not what was received,
   unless the caller passed `stream=True`: `requests` buffers the body before the
   SDK sees it. Called out at the constant rather than left to be discovered.
3. **The seller middleware is FastAPI/Starlette.** Starlette hands the middleware
   the raw bytes, so `bind_body` needs no `captureRawBody` equivalent — it reads
   them and re-arms `receive` for the handler (there is a test that the handler
   still sees the buyer's body). And `BaseHTTPMiddleware` buffers the handler's
   response, so there is no headers-already-flushed case and therefore no
   detached-settle branch and no delivery probe.

One caveat that is WIDER in Python than in Node and is documented in the guide
under a warning: the single-use and in-flight guards are process-local, and a
multi-worker `uvicorn`/`gunicorn` deployment is already several processes. A
hard guarantee needs a shared store this package does not provide.

## Verification

- **776 unit tests pass** (211 of them new, under `tests/unit/mpp/`). The
  `tests/unit/langsmith/test_spans.py` failures on my machine predate this branch
  — they fail identically at `origin/main`.
- `black --check .` clean, `mkdocs build` clean.
- **All 9 e2e tests pass against `staging_sandbox`** (run
  [32345565150](https://github.com/nevermined-io/payments-py/actions/runs/32345565150)):
  challenge mint, credential mint, verify, settle, the idempotent re-settle, and
  the **full buyer↔seller loop** — a uvicorn-served FastAPI app behind the
  middleware, paid over a real socket by `payments.mpp.fetch`. MPP is deployed on
  staging now; it was not when the TS PR shipped
  (nevermined-io/nvm-monorepo#2645), so that gap is closed for Python.
- **The e2e caught a real defect on its first run**, which is fixed here: the
  first settle exceeded the SDK's generic 30s read default and surfaced as
  `MppSettlementOutcomeUnknownError` while the idempotent retry succeeded at
  once. The classification was right — the burn HAD committed — but inheriting
  that default makes every slow-but-successful burn look unknown, so the seller
  never attaches a receipt and the buyer reads `paid=False` on a request that was
  paid for. Burning calls now get 90s to read; the connect deadline is untouched,
  since a connection that never opens has burned nothing.

### One unrelated CI fix rides along

The e2e job was red before this branch for a reason that is not MPP:
`tests/e2e/test_a2a_e2e.py` builds its own `Payments` instances instead of using
the conftest fixtures, and was the only e2e file doing that **without**
`organization_id`. Everything it publishes therefore went into the builder's
PERSONAL workspace, whose tier caps publishing at 10 plans / 20 agents.

That cap counts LISTED resources, and neither a plan nor an agent can be deleted
— both are on-chain registry entries, so the only release is an unlist. The
personal workspace sits at **64 listed plans against a cap of 10**, so the slots
earlier runs took are never coming back and every run 403s with
`BCK.ENTITLEMENTS.0001` before the suite does anything. The error names "the
active workspace" but not which one, which is why it reads like an auth or
backend fault.

Pinning the same org conftest already pins for every other file fixes it. The
whole e2e job is now green: **73 passed, 25 skipped, 0 failed, 0 errors**
([run 32346722817](https://github.com/nevermined-io/payments-py/actions/runs/32346722817)).

Two things this does NOT fix, worth a follow-up rather than more scope here:
the personal workspace stays dirty (64 plans / 77 agents listed), so a future
test that forgets the org pin hits the same wall; and the org accumulates too
(≥94 listed plans), with no slot-reclaim step like the monorepo's
`tier-slot-reclaim.ts` (nevermined-io/nvm-monorepo#2745).


r-marques:
:eyes: Reviewing

payments_py/x402/fastapi/mpp_flow.py
```diff
@@ -0,0 +1,603 @@
...
+            )
+            return served
+        except Exception as settle_error:  # noqa: BLE001
+            # A DEFINITE failure: the resource has been served and the seller has
+            # not been paid — the most expensive thing that can happen to them.
+            # on_payment_error is documented as the handler for payment failures,
+            # so notify through it; this is the one MPP failure site that runs
+            # after the response, so the hook cannot answer the request and is
+            # purely a signal.
+            payment_context.credits_to_settle = 0
```
r-marques, line 525: **🟡 SHOULD FIX — The definite settle failure is the one outcome `on_after_settle` never sees**

This branch is, in your own words two lines up, "the most expensive thing that can happen to them" — served, and definitely not paid. It notifies `on_payment_error` and returns. It does **not** call `_run_after_settle_hook`.

The branch immediately above it does:

```python
except MppSettlementOutcomeUnknownError as settle_error:   # :484
    ...
    await self._run_after_settle_hook(
        self.credits_to_charge,
        MppSettlementOutcomeUnknown(reason=settle_error.message),
    )
except Exception as settle_error:                          # :502  ← this one
    payment_context.credits_to_settle = 0
    await self._notify_payment_error(settle_error)
    return served                                          # no after-settle hook
```

The reasoning given for the unknown branch — "skipping `on_after_settle` would make a real burn vanish from the seller's own accounting" — argues the general case: `on_after_settle` is where a seller writes its usage/revenue record. So a seller reconciling from that hook gets a row for *settled* and a row for *unknown*, and **nothing at all** for the definite failure. The absence is the only signal, and it is indistinguishable from "this request was never an MPP request".

Concrete scenario: the backend answers `BCK.MPP.0003` on settle (a 4xx, so correctly classified definite). Buyer got a 2xx body. Seller's ledger, built on `on_after_settle`, has no entry — so their delivered-but-unpaid count reads zero, and the only trace is a `logger.error` and an `on_payment_error` callback whose documented job (per `_notify_payment_error`'s own docstring) is answering the request, which it cannot do here.

Calling the hook with `credits=0` and a typed outcome would make the trichotomy — settled / unknown / definitely-not — three positive signals instead of two plus an absence. `payment_context.credits_to_settle` is already set to 0 on the line this comment is anchored to, so the value is in hand.

<sub>🤖 pr-review-toolkit · reviewer (own verification, corroborated by type-design-analyzer) · high</sub>

payments_py/x402/fastapi/mpp_support.py
```diff
@@ -0,0 +1,222 @@
...
+    option: Optional[Union[bool, Dict[str, Any]]],
+) -> ResolvedMppOption:
+    """Normalize the per-route ``mpp`` option. ``True`` is shorthand for
+    ``{"bind_body": False}``."""
+    if option is None or option is False:
+        return ResolvedMppOption(enabled=False, bind_body=False)
+    if option is True:
+        return ResolvedMppOption(enabled=True, bind_body=False)
+    return ResolvedMppOption(
+        enabled=True, bind_body=option.get("bind_body", False) is True
```
r-marques: **🟡 SHOULD FIX — A misspelled `bind_body` key silently ships an unbound challenge**

`option.get("bind_body", False) is True` against `RouteConfig.mpp: Optional[Union[bool, Dict[str, Any]]]` means `{"bindBody": True}`, `{"bind-body": True}` and `{"bind_body": "true"}` all resolve to `bind_body=False`. No error, no warning, no log.

What that silently turns off is a security control, not a nicety. `mpp_flow.py:236-243` documents it precisely: with no digest bound, "the backend skips the comparison when the challenge carries no digest, so **unbound** means the BUYER decides whether `bind_body` applies" — mint against an empty request, then attach any body to the paid retry.

The inconsistency is what makes this worth fixing rather than documenting: the **outer** config already rejects unknown keys. `middleware.py:242` does `RouteConfig(**value)`, so a route dict with `{"plan_ids": ...}` raises `TypeError` at startup. The nested `mpp` dict is the one place that accepts anything and keeps going.

Coverage is one-sided too — `tests/unit/mpp/test_fastapi_seller.py:404` exercises `{"bind_body": True}` and `:442` exercises the off case, but nothing pins that an unrecognized key is refused, so the typo path is untested in both directions.

An `MppRouteOptions` dataclass constructed the same way as `RouteConfig` inherits the `TypeError` for free; failing that, raising on any key outside `{"bind_body"}` is a few lines.

<sub>🤖 pr-review-toolkit · type-design-analyzer (verified independently) · high</sub>

payments_py/x402/fastapi/mpp_flow.py
```diff
@@ -0,0 +1,603 @@
...
+                    agent_id=self.route_config.agent_id,
+                    resource=self.resource,
+                    http_verb=self.http_verb,
+                    digest=self.body_digest,
+                    description=self.route_config.description,
+                )
+            )
+            challenge = issued["challenge"]
+        except Exception as challenge_error:  # noqa: BLE001
+            await self._notify_payment_error(challenge_error)
```
r-marques: **🟡 SHOULD FIX — This is the one `on_payment_error` site that drops a hook-returned response**

`_notify_payment_error` returns `Optional[Response]`, and its docstring states the policy outright: *"A hook that returns a `Response` has answered the request explicitly … and it wins."*

Here the return value is discarded and a fixed 500 is sent regardless:

```python
except Exception as challenge_error:
    await self._notify_payment_error(challenge_error)   # ← return value dropped
    logger.error("MPP challenge issuance failed: %s", challenge_error)
    return JSONResponse(status_code=500, ...)
```

Compare `_resolve_credits` (`:223`), which is the same shape and honours it: `return taken_over or JSONResponse(...)`. The credential-rejection sites at `:298`, `:335`, `:364` and `:417` all honour it too. Two other sites drop it deliberately and say why (`:383`, the hook's own bug; `:510`, after the response is committed) — this one gives no such reason.

Effect: a seller who wires `on_payment_error` to return a branded 503 or to shed load gets that honoured when the credits callable throws, and silently ignored when challenge issuance throws — the two adjacent server-side failures on the same route. Both are reachable via the same `BCK.MPP.0002` "MPP not configured" condition the docstring cites.

Either `return taken_over or JSONResponse(...)`, or add a sentence saying why issuance is exempt. No test covers a hook returning a `Response` on this path today, so the current behaviour is unpinned either way.

<sub>🤖 pr-review-toolkit · reviewer (own verification) · high</sub>

r-marques (commented):

**The money paths are right, and I traced them rather than reading the rationale.** The buyer counts a credential as presented *before* the request carrying it is issued (`fetch.py:596-599`) and funnels every exit — return and raise alike — through one boundary that attaches `MppSpendReport` (`fetch.py:641-642`), so no path leaves a caller unable to tell whether money moved. The seller marks a credential spent *before* settling a delivered 2xx (`mpp_flow.py:466`), releases the in-flight claim in a `finally` (`mpp_flow.py:425-430`), and skips settlement on any non-2xx. The `burns=True` classification (`mpp_api.py:305-397`) puts ReadTimeout, a post-write socket death, 5xx/408 and an unreadable 2xx on the UNKNOWN side while leaving ConnectTimeout and 4xx definite — the asymmetry the accounting depends on.

Three findings below, none blocking. The largest is that the seller's *definite* settle failure — the branch the code itself calls "the most expensive thing that can happen to them" — is the one settlement outcome that never reaches `on_after_settle`.

**What I reviewed**
- Diff range `origin/main...HEAD` — **28 files**, +6034/−92, head **`349dd5b7`**.
- **Panel: partial, and I am saying so rather than implying otherwise.** Five agents were spawned; only `type-design-analyzer` returned a report. `code-reviewer`, `silent-failure-hunter`, `pr-test-analyzer` and `comment-analyzer` each finished without delivering findings despite repeated requests. **I covered the general, failure-path and convention ground myself** by reading `fetch.py`, `codec.py`, `mpp_api.py`, `errors.py`, `mpp_flow.py` and `mpp_support.py` end to end; the test-coverage and comment-rot dimensions got a lighter pass than a full panel would give, and a second human eye on those two is worth having.
- **Every type-design claim I repeat below, I re-verified in the code** — I did not relay any of them on trust.
- Checked the repo's own release rules: `pyproject.toml` untouched, `LOCKED_API_VERSION` untouched, and `black --check` reports all 10 changed modules clean.
- **Verified the "four edits, not two" docs claim**: chapter 15 is registered in `mkdocs.yml` nav, in `FILE_MAPPING` *and* in `LINK_MAPPING` (`scripts/convert_to_mintlify.py:93-97`, `:115`) — so it will not ship dead to the docs site, and the `14-onboarding-customers.md` gap you flagged is genuinely pre-existing.
- **Checked one thing that looked wrong and was not**: pinning `organization_id` on the *subscriber* in `test_a2a_e2e.py` reads as over-broad against a comment that justifies only the publisher — but `tests/e2e/conftest.py:114-116` pins the same org on the subscriber fixture, so the change matches the established convention exactly.
- Confirmed the 90s settle override is well-formed: `DEFAULT_HTTP_TIMEOUT` is a `(connect, read)` tuple (`base_payments.py:35`), so `options["timeout"][0]` is a real index, not a latent `TypeError`.

**Verdict:** 🟡 Mergeable with nits — 0 blockers, 3 should-fix, 3 nits, 2 praise.

### 🧭 Inline comments (8)

- 🟡 **SHOULD FIX** — The definite settle failure is the one outcome `on_after_settle` never sees — `payments_py/x402/fastapi/mpp_flow.py:509`
- 🟡 **SHOULD FIX** — A misspelled `bind_body` key silently ships an unbound challenge — `payments_py/x402/fastapi/mpp_support.py:60`
- 🟡 **SHOULD FIX** — This is the one `on_payment_error` site that drops a hook-returned response — `payments_py/x402/fastapi/mpp_flow.py:163`
- 💡 **NIT** — `PaymentContext.mpp` keys exist only in a source comment, and the obvious guess KeyErrors — `payments_py/x402/types.py:320`
- 💡 **NIT** — `MPP_SPEND_OUTCOME_UNKNOWN_CODE` is public API with no consumer and no test — `payments_py/mpp/errors.py:144`
- 💡 **NIT** — Function-level import of a module this file already imports at the top — `tests/unit/mpp/test_fetch.py:609`
- ✅ **NICE** — Counting the credential before the request is what makes the accounting honest — `payments_py/mpp/fetch.py:608`
- ✅ **NICE** — Chapter 15 is registered in all three places — I checked, and the "four edits" claim holds — `scripts/convert_to_mintlify.py:115`

payments_py/x402/types.py
```diff
@@ -313,6 +314,10 @@ class PaymentContext:
     verified: bool
     agent_request_id: Optional[str] = None
     agent_request: Optional[Any] = None
+    # MPP framing details, present only when the route was paid over MPP.
+    # The credential is the opaque ``Payment …`` value the buyer presented;
+    # resource and http_verb are what the challenge was bound to.
+    mpp: Optional[Dict[str, str]] = None
```
r-marques: **💡 NIT — `PaymentContext.mpp` keys exist only in a source comment, and the obvious guess KeyErrors**

`Optional[Dict[str, str]]` carries no key names. It is populated at `mpp_flow.py:443-447` with `credential`, `resource` and `http_verb`, and `docs/api/15-mpp.md:49-50` tells sellers to read `context.mpp` from their handler — without naming a single key.

The snake/camel mix in this PR makes that more than a documentation gap: `RedeemMppParams` serializes the same concept as `httpVerb` on the wire (`mpp_api.py:299`), so `context.mpp["httpVerb"]` is the natural thing for a seller to write, and it raises `KeyError` at runtime inside their handler.

A small frozen dataclass (or a `TypedDict`) makes the three keys discoverable from the type and turns the mistake into an attribute error at author time.

<sub>🤖 pr-review-toolkit · type-design-analyzer (verified independently) · medium</sub>

payments_py/mpp/errors.py
```diff
@@ -0,0 +1,255 @@
...
+        # make ``isinstance`` false for a genuinely-MPP error, and the check
+        # would degrade silently to the "nothing happened" path — which the
+        # integration guide tells sellers to rely on. Same across a process or
+        # serialization boundary.
+        super().__init__(message, MPP_SETTLEMENT_OUTCOME_UNKNOWN_CODE)
+
+
+#: Stable ``code`` on :class:`MppSpendOutcomeUnknownError`. Not a backend code,
+#: for the same reason as :data:`MPP_SETTLEMENT_OUTCOME_UNKNOWN_CODE`.
+MPP_SPEND_OUTCOME_UNKNOWN_CODE = "spend_outcome_unknown"
```
r-marques, line 144: **💡 NIT — `MPP_SPEND_OUTCOME_UNKNOWN_CODE` is public API with no consumer and no test**

Grepping the tree, this constant appears only in its own definition, in the `MppSpendOutcomeUnknownError` constructor that sets it, and in the two export lists (`errors.py:250`, `mpp/__init__.py:62`). Nothing reads it, no test pins it, no doc mentions it.

Its sibling `MPP_SETTLEMENT_OUTCOME_UNKNOWN_CODE` is in better shape — `tests/unit/mpp/test_mpp_api.py:261` asserts an error carries it and `:388` asserts it never collides with the `BCK.MPP.` namespace. That second assertion is the one that gives the constant its purpose, and the spend-side twin has no equivalent.

Either pin it the same way (the namespace-collision assertion is the valuable half), or keep it module-private until a consumer exists — once exported, withdrawing it costs a major bump.

<sub>🤖 pr-review-toolkit · type-design-analyzer (verified independently) · medium</sub>

tests/unit/mpp/test_fetch.py
```diff
@@ -0,0 +1,779 @@
...
+        assert mpp_spend_of(excinfo.value) is None
+
+    def test_reports_no_spend_on_a_first_turn_cap_refusal(self, transport, minter):
+        transport([challenged(credits="10")])
+        with pytest.raises(PaymentsError) as excinfo:
+            mpp_fetch(minter, "GET", DEFAULT_URL, options(max_credits="1"))
+        assert mpp_spend_of(excinfo.value) is None
+
+    def test_reports_no_spend_when_the_mint_itself_fails(self, transport):
+        from .conftest import FakeMinter
```
r-marques: **💡 NIT — Function-level import of a module this file already imports at the top**

`from .conftest import FakeMinter` sits inside the test body, while line 26 of this same file already does `from .conftest import DEFAULT_URL, make_response` at module level — so there is no circular-import pressure to work around, which is the exemption the repo's `CLAUDE.md` allows (*"All imports must be at the top of the file — do not use inline imports inside functions unless absolutely necessary (e.g. to avoid circular dependencies)"*).

Adding `FakeMinter` to the existing line-26 import removes the special case. Purely a convention nit; nothing misbehaves.

<sub>🤖 pr-review-toolkit · reviewer (own verification) · high</sub>

payments_py/mpp/fetch.py
```diff
@@ -0,0 +1,687 @@
...
+            headers["Authorization"] = (
+                f"{credential}, {existing_auth}" if existing_auth else credential
+            )
+            retry_kwargs = {**request_kwargs, "headers": headers}
+
+            # Counted BEFORE the request, not after: the credential is on the
+            # wire as soon as the request is issued, so a failure here
+            # (disconnect, DNS, a read timeout) must not report zero credentials
+            # presented. The seller may already have verified and burned it.
+            credentials_presented += 1
```
r-marques, line 653: **✅ NICE — Counting the credential before the request is what makes the accounting honest**

Incrementing `credentials_presented` and `credits_presented_total` *before* `requests.request` — rather than after it returns — is the detail that makes every downstream guarantee hold. A disconnect, a DNS failure or a read timeout on this line leaves the credential already on the wire and possibly already burned; counted afterwards, all three would report zero presented and send a caller straight into a blind retry.

It composes with the single `except BaseException` boundary at `:641-642`, so the property is structural rather than per-branch: there is no exit from `mpp_fetch` — return or raise — that can be silent about money. I checked each exit individually (the two `MppFetchResult` constructions at `:664` and `:679`, the terminal `to_mpp_error` raise, and the wrapper path) and they all carry the same accounting.

The conservative direction is right too: `credits_presented` is documented as an upper bound and deliberately not lowered for a credential a single-use seller refused, because the buyer cannot see which one the remote honoured.

<sub>🤖 pr-review-toolkit · reviewer (own verification) · high</sub>

scripts/convert_to_mintlify.py
```diff
@@ -107,6 +112,7 @@
     "11-x402.md": "/api-reference/python/x402-module",
     "12-langchain-integration.md": "/api-reference/python/langchain-module",
     "13-langsmith-deployment.md": "/api-reference/python/langsmith-deployment-module",
+    "15-mpp.md": "/api-reference/python/mpp-module",
```
r-marques, line 115: **✅ NICE — Chapter 15 is registered in all three places — I checked, and the "four edits" claim holds**

The failure mode you describe is real and silent: a chapter present in `FILE_MAPPING` but absent from `LINK_MAPPING` (or from the `mkdocs.yml` nav) still builds locally and still ships dead to the docs site, which is exactly what PR #202 had to fix retroactively for chapter 13.

Verified all three registrations for chapter 15: `mkdocs.yml` nav (`15. MPP Protocol`), `FILE_MAPPING` (`:93-97`, → `mpp-module.mdx`) and `LINK_MAPPING` (this line). So the new chapter is not going to repeat #202.

Also confirms the corollary you flagged and deliberately left alone: `14-onboarding-customers.md` is in the nav but in neither mapping, so that chapter genuinely never reaches Mintlify. Leaving it out of a PR about MPP is the right call — its slug and description are editorial — but it is a live bug worth its own issue rather than a note in this description, since nothing in the repo will surface it again.

<sub>🤖 pr-review-toolkit · reviewer (own verification) · high</sub>

r-marques (approved):
🟡 **Approved with comments** — no blockers; 3 should-fix, 3 nit(s) left inline.

Inline review: https://github.com/nevermined-io/payments-py/pull/262#pullrequestreview-4980743961

payments_py/mpp/fetch.py
```diff
@@ -0,0 +1,687 @@
...
+def _is_non_replayable_body(request_kwargs: Dict[str, Any]) -> bool:
+    """Whether the request body can only be sent once.
+
+    ``requests`` accepts a generator, an iterator or a file-like object as
+    ``data=``; all three are consumed by the first send, so replaying them on
+    the retry would silently transmit an empty (or partial) body rather than
+    raise. Every other shape — ``str``, ``bytes``, a dict/list of form fields,
+    ``json=``, ``files=`` — can be sent again, so a retry is safe.
+    """
+    body = request_kwargs.get("data")
```
r-marques: **🔴 BLOCKER — `files=` is not replayable, so the PAID retry sends an empty file**

The docstring above lists `files=` among the shapes that "can be sent again, so a retry is safe", and the implementation follows that claim — it inspects only `request_kwargs.get("data")`. Both are wrong about `files=`, and the consequence lands on the request that costs credits.

`requests` encodes `files=` through `RequestEncodingMixin._encode_files`, which calls `fp.read()` on each handle. The second encode gets an exhausted handle. Reproduced against this branch:

```python
f = io.BytesIO(b"PAYLOAD-PAYLOAD-PAYLOAD")
files = {"f": ("a.txt", f)}
RequestEncodingMixin._encode_files(files, {})   # 161 bytes, payload present
RequestEncodingMixin._encode_files(files, {})   # 138 bytes, payload GONE

_is_non_replayable_body({"files": files})       # False  ← guard misses it
_is_non_replayable_body({"data": io.BytesIO()}) # True   ← guard catches this one
```

Failure scenario, end to end: a buyer calls `payments.mpp.fetch("POST", url, options, files={"doc": open(path, "rb")})` against a challenging endpoint. Turn 1 sends the file and comes back 402. The guard at this line says the body is replayable, so the helper mints a credential, **charges the plan**, and issues the retry — with an empty multipart part. The seller verifies the credential, burns the credits and serves a response computed from nothing. `MppFetchResult.paid` is `True`. Nobody sees an error.

That is precisely the outcome `_is_non_replayable_body` exists to prevent — its own docstring says an unreplayable body "would silently transmit an empty (or partial) body rather than raise". The `data=` case is handled; `files=` slips through because the docstring asserted it was safe.

Fix is small: inspect `files=` as well, treating any value whose handle is a stream (`hasattr(v, "read")`, including the 2-/3-tuple forms `("name", fp)` / `("name", fp, ctype)`) as non-replayable, and correct the docstring so it no longer licenses the narrower check. A `bytes`/`str` file payload stays replayable and should keep working.

Worth a test alongside `test_refuses_a_single_read_body_once_a_retry_is_required` (`tests/unit/mpp/test_fetch.py:470`) — that one pins the `data=` path only, which is why this went unnoticed.

<sub>🤖 pr-review-toolkit · comment-analyzer (independently reproduced by the reviewer) · high</sub>

payments_py/mpp/codec.py
```diff
@@ -0,0 +1,470 @@
...
+    Returns the ``challenge.id`` a token68 credential decodes to, or ``None``
+    when the credential cannot be decoded into one — a structured
+    ``key="value"`` scheme (that shape is a challenge, never a credential),
+    undecodable base64url JSON, or JSON without a non-empty string
+    ``challenge.id``. The backend rejects all of those anyway; returning ``None``
+    lets the caller refuse them without a round-trip rather than fall back to a
+    key it cannot trust.
+    """
+    token68 = _PAYMENT_PREFIX.sub("", credential).strip()
+    if not token68 or _AUTH_PARAM_START.match(token68):
```
r-marques: **🟡 SHOULD FIX — A padded base64url credential is misread as a structured challenge and refused forever**

`_AUTH_PARAM_START` is `^{token}\s*=` (`codec.py:23`). Base64url padding is `=`, so a padded token68 matches it, this guard classifies the credential as a structured `key=value` challenge, and the function returns `None`.

Reproduced on this branch with an otherwise valid credential:

```
unpadded (ends 'X0'): challenge id -> c1
PADDED   (ends '0='): challenge id -> None
```

RFC 7235 defines token68 as explicitly permitting trailing `=`, so a padded credential is well-formed on the wire. Downstream, `mpp_flow.py:296-304` treats `None` as "no decodable challenge id" and refuses with `BCK.MPP.0003` plus a fresh challenge — which the buyer answers by minting again, padding again, and being refused again.

This falsifies the docstring directly above the guard: it says `None` means "a structured `key=\"value\"` scheme … undecodable base64url JSON, or JSON without a non-empty string `challenge.id`", and that "the backend rejects all of those anyway". A padded credential is in none of those three classes, and the backend would decode it fine — the whole justification for refusing without a round-trip is that the backend would refuse too, which here it would not.

First-party buyers are unaffected because `build_credential_header` (`codec.py:411`) strips padding, which is also why no test catches it (`tests/unit/mpp/test_codec.py:408-430` builds its inputs the same way). A third-party MPP buyer that pads is simply unable to buy from a payments-py seller.

Stripping `=` before the structured-vs-token68 test, or requiring the matched key to contain a non-`=` character, closes it.

<sub>🤖 pr-review-toolkit · comment-analyzer (independently reproduced by the reviewer) · high</sub>

docs/api/15-mpp.md
```diff
@@ -0,0 +1,259 @@
...
+  with a 402 carrying `code: "BCK.MPP.0003"` and a *fresh* challenge, so the
+  buyer can still make progress by paying again.
+- **No concurrent double-spend within the process.** A second request presenting
+  a credential already in flight gets `409 Conflict`. Verification burns
+  nothing and settlement is idempotent, so without this guard N concurrent
+  requests would each be served for a single burn.
+- **Settlement only on a 2xx.** A handler that fails or refuses is never
+  settled, and the credential stays unspent.
+
+!!! danger "Both guards are process-local"
```
r-marques, line 91: **🟡 SHOULD FIX — Both `!!! danger` blocks ship dead to the docs site — including the process-local guard warning**

`scripts/convert_to_mintlify.py` has patterns for `warning`, `note`, `tip` and `info` only (`:131-168`). There is no `danger` case, and this PR introduces the only two `!!! danger` blocks in the whole `docs/` tree — this one and `:237`.

So neither is converted. The literal `!!! danger "…"` line survives into the `.mdx`, and the 4-space-indented body underneath renders as an **indented code block** rather than a callout. `!!! warning "Experimental"` at `:16` converts correctly, so the same page will show one proper callout and two blocks of broken markup.

The content that breaks is the worst possible choice: this block is the process-local single-use caveat — the one telling operators that a multi-worker uvicorn deployment does not enforce single-use — and `:237` is "Settlement failures are not all the same", the burn-may-have-committed warning. Those are the two paragraphs a seller most needs to read, and they are the two that arrive malformed.

This is the same class of defect you caught and fixed for the x402 cross-reference, and the same one PR #202 fixed retroactively for chapter 13 — a docs problem invisible in `mkdocs build` and only visible after publish.

Either add a `danger` → `<Warning>` (Mintlify has no `<Danger>`; `<Warning>` is the strongest callout) pattern to the converter, or downgrade both to `!!! warning`, which is already known to convert.

<sub>🤖 pr-review-toolkit · comment-analyzer (independently reproduced by the reviewer) · high</sub>

payments_py/payments.py
```diff
@@ -94,6 +95,9 @@ def _initialize_api(self, options: PaymentOptions) -> None:
         self.observability = ObservabilityAPI.get_instance(options)
         self.organizations = OrganizationsAPI.get_instance(options)
         self.facilitator = FacilitatorAPI.get_instance(options)
+        # MPP: a second payment framing over the same plans/credits/
```
r-marques, line 98: **💡 NIT — `payments.mpp` has no `docs/reference/` page, so its docstrings render nowhere**

`docs/reference/payments.md` renders each peer API through mkdocstrings — `FacilitatorAPI` (`:31`), `X402TokenAPI` (`:38`), plus the plans/agents APIs — and `MppAPI` was not added; the Reference section of `mkdocs.yml` is untouched too.

Per this repo's `CLAUDE.md`, `docs/reference/` is the generated-from-docstrings half of the docs, and the docstrings in this PR are unusually rich — `MppAPI`, `MppFetchOptions`, `MppFetchResult` and the error hierarchy all carry the reasoning that makes the money semantics usable. None of it reaches a reader today.

One line (`::: payments_py.mpp.mpp_api.MppAPI`, alongside the existing entries) plus a nav entry fixes it.

docs/api/15-mpp.md
```diff
@@ -0,0 +1,259 @@
...
+ownership of the response, because a 402 without a fresh challenge leaves the
+buyer unable to make progress. A hook that returns a `Response` still wins.
+
+`on_payment_error` is **not** called for the credential-less opening request —
+that is the first turn of every healthy payment cycle, and notifying there would
+drown the rejections the hook exists to surface. It *is* called when an
+`Authorization` header arrives carrying no `Payment` scheme, which means an
+intermediary is rewriting it and the buyer is stuck in a silent retry loop.
+
+`on_after_settle` receives what the backend says it **burned**, not what this
```
r-marques: **💡 NIT — This overstates `on_after_settle`'s credits argument — the PR's own test shows the charged amount**

"…receives what the backend says it BURNED, not what this request would charge — and 0 for a settlement that failed" is stronger than the code.

Three branches pass `credits_to_charge` (the charged amount) rather than a measured burn: the unknown-outcome path (`mpp_flow.py:497-500`), success-with-no-`creditsRedeemed` (`:591`) and unusable `creditsRedeemed` (`:600`). `_credits_settled`'s own docstring is honest about it — it calls the fallback "a GUESS" — and `tests/unit/mpp/test_fastapi_seller.py:329-350` asserts exactly that: `credits == 2`, the charged amount, on the unknown path.

The second half needs a caveat too: "0 for a settlement that failed" holds when settle *returns* `success: false`, but a settle that *raises* a definite error skips `on_after_settle` altogether (`mpp_flow.py:502-512` — the round-1 should-fix on this same branch).

The code is fine; it is the chapter that promises a measurement it does not always deliver.

docs/api/15-mpp.md
```diff
@@ -0,0 +1,259 @@
...
+    # A guard refused to even attempt the call: a bad argument, a challenge
+    # naming another plan, a body that cannot be replayed.
+    ...
+except MppError as err:
+    # What the wire actually said: a rejected credential, a malformed
+    # challenge, an MPP-disabled environment.
+    spend = mpp_spend_of(err)
+    if spend:
+        # A credential was already on the wire. Do NOT blindly retry.
+        log.warning("up to %s credits may have burned", spend.credits_presented)
```
r-marques: **💡 NIT — This example is a `NameError`, and its comment contradicts the paragraph below it**

Two small things in one snippet:

- `log.warning(...)` — `log` is never defined anywhere in the chapter. Copy-pasted, this raises `NameError`.
- The `except PaymentsError` branch is commented "A guard refused to even attempt the call", but the paragraph at `:190-193` immediately explains that a `PaymentsError` **can** carry a spend report when a caller-constraint guard fires on the re-challenge turn (`fetch.py:401-404`). As written the example teaches the reader to discard the money signal on the very branch the surrounding prose tells them to check.

Calling `mpp_spend_of(err)` in that branch too — the helper exists precisely because the report rides on both error families — makes the example agree with the text.

payments_py/x402/fastapi/mpp_flow.py
```diff
@@ -0,0 +1,603 @@
...
+async def _maybe_await(value: Any) -> Any:
+    return await value if inspect.isawaitable(value) else value
+
+
+async def _read_and_replay_body(request: Request) -> bytes:
+    """Read the raw request body and make it readable again downstream.
+
+    ``BaseHTTPMiddleware`` gives the handler its own ``Request`` built from the
+    same ASGI ``receive`` channel, which is single-use — so reading the body here
+    would otherwise leave the handler with an empty one. Re-arming ``receive``
```
r-marques: **💡 NIT — The body-replay explanation names a mechanism Starlette does not use here**

The behaviour is correct and pinned (`test_fastapi_seller.py:413` proves the handler still sees the buyer's body) — this is about the explanation, which would mislead whoever next changes this code.

Against starlette 0.49.3, `BaseHTTPMiddleware` hands the downstream app `_CachedRequest.wrapped_receive`, which replays `self._body` — already populated by the `await request.body()` on the line above. The `request._receive = replay` assignment is not what the handler reads on this path, and `_CachedRequest`'s own class docstring says caching the body for exactly this case is its purpose.

So "reading the body here would otherwise leave the handler with an empty one" is not true of this Starlette version, and `# noqa: SLF001 — the documented re-arm` claims a documented pattern for a private attribute that is not consulted. Keeping the assignment as defensive belt-and-braces is fine; the comment should say that rather than assert it is the load-bearing mechanism.

payments_py/x402/fastapi/mpp_support.py
```diff
@@ -0,0 +1,222 @@
...
+#: What this does NOT close: multiple worker processes or horizontally-scaled
+#: instances, which do not share this in-memory set and would need a shared
+#: store (e.g. Redis) this package does not provide. Note that a multi-worker
+#: uvicorn/gunicorn deployment is already several processes, so this caveat
+#: applies to an ordinary production setup, not only to a scaled-out cluster.
+_in_flight_mpp_credentials: Set[str] = set()
+
+#: How long a settled credential stays refused. Matches the backend's
+#: ``MPP_CHALLENGE_TTL_SECONDS`` (300): past it the challenge itself is expired,
+#: so the credential is refused by the backend anyway and there is nothing left
```
r-marques: **💡 NIT — This TTL is pinned to a backend constant by prose alone, and drift degrades single-use silently**

`SPENT_CREDENTIAL_TTL_SECONDS = 300.0` is justified as "Matches the backend's `MPP_CHALLENGE_TTL_SECONDS` (300)", and the reasoning is right: past the challenge's own expiry the backend refuses the credential anyway, so there is nothing left to protect.

But nothing in this repo can notice if nvm-monorepo raises that TTL. If it does, this map forgets a credential the backend still honours, and the "Single use" guarantee advertised at `docs/api/15-mpp.md:75-77` quietly becomes "replayable after 300 seconds" — the exact replay the comment at `:141-148` explains the backend cannot refuse on its own.

Reading the challenge's own `expires` (already parsed onto `MppChallenge.expires`) would make the coupling self-maintaining; failing that, an e2e assertion that a credential is still refused just under the backend's TTL would at least fail loudly on drift.

r-marques (commented):
## 🤖 Automated PR review — round 2 — 🔴 Blocked

**This review corrects my own approval from an hour ago, and the correction is the point.** I posted round 1 as *approved with comments* while four of five panel legs had returned nothing; I said so in that review rather than implying a full panel. One of the missing legs then reported, and its lead finding reproduces as a **blocker on the paid retry**. The approval was premature — this round replaces it.

**The blocker: `files=` is not replayable, so a challenged request pays for an empty file.** `_is_non_replayable_body` (`fetch.py:188-201`) inspects only `data=`, licensed by a docstring that lists `files=` among the shapes that "can be sent again". `requests` encodes `files=` via `_encode_files`, which calls `fp.read()`. I reproduced it on this branch: first encode 161 bytes with the payload, second encode 138 bytes **without** it, while the guard returns `False` for `files=` and `True` for the `data=` stream it already handles. A buyer passing `files={"doc": open(path,"rb")}` to a challenging endpoint therefore mints a credential, burns credits, and retries with an empty multipart part — `paid=True`, no error, nothing delivered. That is the exact silent-empty-body outcome the guard's own docstring says it exists to prevent.

Two further reproduced defects ride with it: a **padded** base64url credential is misclassified as a structured challenge and refused forever (`codec.py:173` — `unpadded → "c1"`, `padded → None`, so a third-party buyer that pads can never buy), and **both `!!! danger` blocks ship dead to the docs site** because the converter handles `warning|note|tip|info` only — losing the process-local single-use caveat and the burn-may-have-committed warning, the two paragraphs a seller most needs.

**What I reviewed**
- Diff range `origin/main...HEAD` — **28 files**, +6034/−92, head **`349dd5b7`** — unchanged since round 1, so round 1's findings stand as posted.
- **I re-ran every claim in this round myself** rather than relaying it: the double `_encode_files` encode, `_is_non_replayable_body` on both shapes, `extract_credential_challenge_id` on padded vs unpadded credentials, and the converter's admonition table against the two `!!! danger` blocks. Outputs are quoted in the findings.
- **Round-1 items are still open and are NOT re-posted here** — they are already inline on this PR and the head has not moved. Unchanged: `on_after_settle` never firing on the definite settle failure (`mpp_flow.py:502-512`), the `bind_body` typo silently unbinding (`mpp_support.py:60`), the dropped hook response (`mpp_flow.py:163`), and three nits.
- **Panel, honestly stated:** `comment-analyzer` and `type-design-analyzer` reported. `code-reviewer`, `silent-failure-hunter` and `pr-test-analyzer` never returned findings despite repeated requests, so the failure-path and test-coverage dimensions rest on my own reading, not a specialist pass. Given that two of the three defects above came from the one leg that reported late, a human pass on those two dimensions is worth more than usual here.

**Verdict:** 🔴 Blocked — 1 blocker, 2 should-fix, 5 nits **this round**, on top of 3 should-fix and 3 nits still open from round 1.

### 🧭 Inline comments (8)

- 🔴 **BLOCKER** — `files=` is not replayable, so the PAID retry sends an empty file — `payments_py/mpp/fetch.py:197`
- 🟡 **SHOULD FIX** — A padded base64url credential is misread as a structured challenge and refused forever — `payments_py/mpp/codec.py:173`
- 🟡 **SHOULD FIX** — Both `!!! danger` blocks ship dead to the docs site — including the process-local guard warning — `docs/api/15-mpp.md:85`
- 💡 **NIT** — `payments.mpp` has no `docs/reference/` page, so its docstrings render nowhere — `payments_py/payments.py:98`
- 💡 **NIT** — This overstates `on_after_settle`'s credits argument — the PR's own test shows the charged amount — `docs/api/15-mpp.md:104`
- 💡 **NIT** — This example is a `NameError`, and its comment contradicts the paragraph below it — `docs/api/15-mpp.md:187`
- 💡 **NIT** — The body-replay explanation names a mechanism Starlette does not use here — `payments_py/x402/fastapi/mpp_flow.py:70`
- 💡 **NIT** — This TTL is pinned to a backend constant by prose alone, and drift degrades single-use silently — `payments_py/x402/fastapi/mpp_support.py:133`

r-marques (changes requested):
🔴 **Blocked** — 1 blocker(s) to address before merge. Details are inline on the changed lines.

Inline review: https://github.com/nevermined-io/payments-py/pull/262#pullrequestreview-4980783491

payments_py/mpp/fetch.py
```diff
@@ -0,0 +1,732 @@
...
+
+
+def _is_one_shot_stream(value: Any) -> bool:
+    """Whether ``value`` is consumed by being read once."""
+    if isinstance(value, (str, bytes, bytearray)):
+        return False
+    return hasattr(value, "read") or hasattr(value, "__next__")
+
+
+def _files_hold_a_stream(files: Any) -> bool:
```
r-marques, line 196: **✅ NICE — The blocker fix covers every shape, and I re-ran the probe to confirm it**

Splitting this into `_is_one_shot_stream` and `_files_hold_a_stream` is what makes the fix complete rather than a patch for the reported case. The `files=` grammar is awkward — a bare handle, a 2-, 3- or 4-tuple, or a list of pairs — and the container itself is a dict or list, which is exactly the shape the old `data=`-oriented check treated as safe.

I re-ran the measurement that established the blocker, against this head:

```
bare fp                    -> non_replayable=True
2-tuple (name, fp)         -> True
3-tuple (+ctype)           -> True
4-tuple (+headers)         -> True
list of pairs              -> True
bytes payload              -> False
str payload                -> False
data=<file>                -> True    (the case that already worked)
data=bytes / json=         -> False
```

So every stream-bearing form is refused before a credential is minted, and nothing that was legitimately replayable became refused — which is the half a narrower fix would have got wrong in the other direction, turning working uploads into hard errors.

The docstring now also records what the old claim cost, which is the right place for it: the reason this went unnoticed was that `files=` was *listed as safe*, so nobody thought to inspect it.

<sub>🤖 pr-review-toolkit · reviewer (own verification) · high</sub>

payments_py/x402/fastapi/mpp_support.py
```diff
@@ -0,0 +1,317 @@
...
+#: once, be served for the whole challenge TTL — the backend cannot refuse it,
+#: because from its side a replayed settle succeeding IS the idempotency
+#: contract working as designed.
+#:
+#: So single-use is the seller edge's job, and this is where it lives. Same
+#: process-local caveat as the in-flight set.
+_spent_mpp_credentials: Dict[str, float] = {}
+
+
+def _ttl_for(challenge_expires: Optional[str]) -> float:
```
r-marques, line 217: **✅ NICE — This fix caught a second defect that I had looked at and waved through**

I raised the TTL as a comment-rot risk — a constant pinned to a backend value by prose, with nothing able to notice drift. Deriving it from the challenge's own `expires` closes that properly, and the three guards around it are the ones that matter: extend-only (a skewed clock or a stale value cannot *shorten* the replay window), bounded above (a crafted `expires` cannot pin memory), and total on malformed input (falls back to the constant rather than raising inside the settle path).

What is worth flagging is the knock-on you found and I missed. In round 1 I read `mark_credential_spent`'s sweep — which stopped at the first live entry on the assumption that insertion order is expiry order — and judged it safe because every entry shared one TTL. That assumption is exactly what this change breaks, and you spotted it and switched the sweep to walk every expired entry.

Left as it was, the sweep would have stopped behind the first long-lived entry and leaked the rest. It would not have been a correctness bug — `is_credential_spent` still checks each entry individually — but it would have been an unbounded map in a long-lived process, which is a slow leak rather than a visible failure.

<sub>🤖 pr-review-toolkit · reviewer (own verification) · high</sub>

r-marques (commented):
## 🤖 Automated PR review — round 3 — ✅ Ready to merge

**All 14 items from rounds 1 and 2 are fixed, and I verified each one by running it rather than by reading the diff.** The blocker in particular I re-measured with the same probe that proved it: every stream-bearing `files=` shape is now refused before the paid retry, while `bytes`/`str` payloads and `json=` stay replayable.

Two of the fixes are better than what I asked for. The `bind_body` guard doesn't just reject an unknown key — it also rejects a non-bool value, so `{"bind_body": "true"}` fails at construction instead of silently disabling a security control. And the TTL fix forced a second correction I had looked at in round 1 and wrongly judged harmless: once entries stop sharing one TTL, insertion order stops being expiry order, so the sweep's early `break` had to go. It did.

**Every prior finding, checked individually**

| Round | Finding | Status |
| --- | --- | --- |
| 2 | 🔴 `files=` not replayable → paid retry sends an empty file | **Fixed** — verified all 6 documented shapes (bare `fp`, 2/3/4-tuples, list-of-pairs) now refuse; `bytes`/`str`/`json=` still pass |
| 2 | Padded base64url credential refused forever | **Fixed** — decodes at 0-, 1- and 2-char padding (`id='c1'`, `'c1x'`, `'c1xx'`) |
| 2 | Both `!!! danger` blocks ship dead to the docs site | **Fixed** — ran the converter: 14 files, 0 errors, **0** literal `!!! danger` survive, and both blocks now render as titled `<Warning>` |
| 1 | Definite settle failure never reaches `on_after_settle` | **Fixed** — the branch now fires the hook with 0 credits and a typed `MppSettlementFailed` |
| 1 | `bind_body` typo silently unbinds the challenge | **Fixed** — `bindBody` / `bind-body` / non-bool all raise `TypeError` at construction |
| 1 | Challenge issuance discards a hook-returned response | **Fixed** — `:174`/`:181` now `return taken_over or …`; the two remaining discards are the ones documented as deliberate |
| 1 | `PaymentContext.mpp` keys exist only in a comment | **Fixed** — now typed `Optional[MppPaymentFraming]` |
| 1 | `MPP_SPEND_OUTCOME_UNKNOWN_CODE` exported with no consumer | **Fixed** — pinned in `test_mpp_api.py:390`, alongside its sibling |
| 1 | Function-level `conftest` import | **Fixed** — folded into the top-level import at `:27` |
| 1 | Spent-credential TTL pinned to a backend constant by prose | **Fixed, and extended** — derived from the challenge's own `expires`, floored at the old constant so clock skew cannot shorten it, capped so a crafted value cannot pin memory, malformed input falls back |
| 1 | `payments.mpp` absent from `docs/reference/` | **Fixed** — `MppAPI`, `MppFetchOptions`, `MppFetchResult` all render |

**What I reviewed**
- Head **`0181d6c`** (was `349dd5b7`) — a moved head, so this is a fresh round rather than a re-read.
- **Ran the suite**: `pytest -m "not slow"` → **968 passed, 0 failed**, 98 deselected. A full run also showed 25 failures, all under `tests/e2e/` — the credential-gated set this worktree cannot authenticate; the clean `not slow` run confirms they are environmental, not regressions.
- **Reviewed the new code for problems the fixes could have introduced**, not just for whether they work: `_is_non_replayable_body`'s `files=` walk (correctly ignores the dict/list *container* and inspects payloads), and `_ttl_for` (extend-only, bounded, total on malformed input).
- On the two decisions you took differently: both are right. Keeping the `_receive` assignment as documented-defensive is better than deleting it — you verified the Starlette behaviour rather than trusting my report of it, and deleting it would make the code depend on `_CachedRequest` continuing to cache. And splitting the `on_after_settle` table into "amount reported" vs "no usable amount" is a better fix than the caveat I suggested, since `_credits_settled` already calls the fallback a guess.
- `14-onboarding-customers.md` is still missing from both converter mappings — correctly left alone, and it is a pre-existing bug rather than anything this PR introduced. Worth its own issue; nothing here should hold on it.

**Verdict:** ✅ Ready to merge — 0 blockers, 0 should-fix, 0 nits.

### 🧭 Inline comments (2)

- ✅ **NICE** — The blocker fix covers every shape, and I re-ran the probe to confirm it — `payments_py/mpp/fetch.py:196`
- ✅ **NICE** — This fix caught a second defect that I had looked at and waved through — `payments_py/x402/fastapi/mpp_support.py:217`

r-marques (approved):

Inline review: https://github.com/nevermined-io/payments-py/pull/262#pullrequestreview-4981107364

PR #264, opened 2026-08-28
fix(organizations): onboard_customer returns the Bearer (hash), not the encrypted blob

aaitor:
## Why this matters

White-label customer onboarding through the Python SDK is currently broken: `onboard_customer` hands callers a credential that cannot authenticate. Any integrator following our docs — onboard a customer, then act on their behalf with the returned key — gets `401`s, because the value we return is an encrypted internal blob, not a usable Bearer token. This fixes the SDK so the documented flow works. It's the Python twin of the TypeScript fix ([payments#420](https://github.com/nevermined-io/payments/pull/420)), both surfaced by [docs#301](https://github.com/nevermined-io/docs/pull/301).

## The bug

`onboard_customer` builds its result with `CustomerOnboardingResponse.model_validate(walletResult)`, and the model's `nvm_api_key` field is aliased to `nvmApiKey`. The backend documents that field (for the `as=customer` outcome, #2924) as *"the encrypted SDK/server-side blob — NOT a Bearer; do not send it as a header. To authenticate an HTTP request use `hash`."* So `model_validate` populated `nvm_api_key` from the blob, and the method returned it.

`create_member` already does the right thing — it constructs `CreateUserResponse(nvm_api_key=wallet["hash"], …)`. `onboard_customer` relied on the alias instead and picked up the blob.

## The fix

- Surface `wallet["hash"]` (the Bearer) as `nvm_api_key`, via `result.model_copy(update=…)` so the response object stays immutable.
- Guard on `wallet["hash"]` (the field we now return) instead of the blob, so a partial payload still fails loudly.
- Corrected the docstring, which had the semantics inverted.

## Test plan

- Corrected the unit test, which had encoded the inverted premise (it called `hash` a "lookup hash" and asserted the raw blob was returned). It now asserts `hash` is surfaced as the Bearer and the blob never leaks.
- `poetry run pytest tests/unit/test_organizations_api.py` → 34/34 pass.
- `poetry run black --check` clean.

https://claude.ai/code/session_01CZ3y8ptZEYfscAKfhyfs1h

### DIFF
--- a/payments_py/api/organizations_api.py
+++ b/payments_py/api/organizations_api.py
@@ -256,9 +256,11 @@ class OrganizationsAPI(BasePaymentsAPI):
         Admin-only on the backend. Provisions a Nevermined account for the
         customer **without consuming a member seat** and returns a usable,
         scoped NVM API key the org can use to transparently act on the
-        customer's behalf (purchase plans / redeem credits). Unlike
-        :meth:`create_member`, this returns the **real** usable key
-        (``walletResult.nvmApiKey``), not the non-usable lookup hash.
+        customer's behalf (purchase plans / redeem credits). Like
+        :meth:`create_member`, the usable Bearer is ``walletResult.hash``
+        (``<prefix>:<jwt>``) and is surfaced as ``nvm_api_key``; the sibling
+        ``walletResult.nvmApiKey`` is the encrypted server-side blob and is
+        **not** a valid Bearer.
 
         If the email already belongs to an account the org does NOT own, no key
         is issued: the backend replies ``202`` and the result carries
@@ -307,10 +309,16 @@ class OrganizationsAPI(BasePaymentsAPI):
         # or key can leak even if the backend regresses.
         if response.status_code == 202 or result.consent_required:
             return CustomerOnboardingResponse(consent_required=True)
-        # New account or returning customer: a usable key MUST be present. A 2xx
-        # without one means a partial/unexpected payload — fail loudly rather
-        # than return a "success" carrying missing credentials.
-        if not result.nvm_api_key:
+        # New account or returning customer: the Bearer MUST be present. The
+        # backend returns the usable credential as ``walletResult.hash``
+        # (``<prefix>:<jwt>``); the sibling ``nvmApiKey`` — which the
+        # ``model_validate`` above mapped into ``nvm_api_key`` — is the encrypted
+        # server-side blob and is NOT a valid Bearer. Surface ``hash`` instead,
+        # exactly as create_member does. A 2xx without a hash means a
+        # partial/unexpected payload — fail loudly rather than return a "success"
+        # carrying no usable credential.
+        hash_key = wallet.get("hash")
+        if not hash_key:
             raise PaymentsError.from_backend(
                 "Unable to onboard customer",
                 {
@@ -320,7 +328,7 @@ class OrganizationsAPI(BasePaymentsAPI):
                     )
                 },
             )
-        return result
+        return result.model_copy(update={"nvm_api_key": hash_key})
 
     def get_members(
         self,
--- a/payments_py/api/organizations_api.py
+++ b/payments_py/api/organizations_api.py
@@ -317,8 +317,11 @@ class OrganizationsAPI(BasePaymentsAPI):
         # exactly as create_member does. A 2xx without a hash means a
         # partial/unexpected payload — fail loudly rather than return a "success"
         # carrying no usable credential.
+        # ``model_copy(update=…)`` below does not validate, so assert the shape
+        # here: the backend contract is ``<env>:<JWT>`` — a non-string truthy
+        # ``hash`` must never reach ``Payments`` as the Bearer.
         hash_key = wallet.get("hash")
-        if not hash_key:
+        if not isinstance(hash_key, str) or not hash_key:
             raise PaymentsError.from_backend(
                 "Unable to onboard customer",
                 {
--- a/tests/unit/test_organizations_api.py
+++ b/tests/unit/test_organizations_api.py
@@ -689,6 +689,25 @@ class TestOnboardCustomer:
             with pytest.raises(PaymentsError, match="no API key"):
                 payments.organizations.onboard_customer("customer@example.com")
 
+    def test_raises_when_hash_is_not_a_string(self):
+        # model_copy does not validate, so a non-string truthy hash must be
+        # rejected before it reaches Payments as the Bearer.
+        payments = _make_payments()
+        with requests_mock.Mocker() as m:
+            m.post(
+                f"{BACKEND}/api/v1/organizations/account",
+                status_code=201,
+                json={
+                    "success": True,
+                    "walletResult": {
+                        "hash": {"unexpected": "object"},
+                        "userId": "us-1",
+                    },
+                },
+            )
+            with pytest.raises(PaymentsError, match="no API key"):
+                payments.organizations.onboard_customer("customer@example.com")
+
     def test_raises_on_5xx(self):
         payments = _make_payments()
         with requests_mock.Mocker() as m:
--- a/tests/unit/test_organizations_api.py
+++ b/tests/unit/test_organizations_api.py
@@ -600,7 +600,7 @@ class TestConnectStripeAccount:
 
 
 class TestOnboardCustomer:
-    def test_new_customer_returns_real_key_and_sends_as_customer(self):
+    def test_new_customer_returns_hash_as_bearer_and_sends_as_customer(self):
         payments = _make_payments()
         with requests_mock.Mocker() as m:
             m.post(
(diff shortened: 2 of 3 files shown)


aaitor:
@eruizgar91 Python twin of nevermined-io/payments#420 — `onboard_customer` was surfacing `walletResult.nvmApiKey` (the encrypted blob) as `nvm_api_key` instead of `walletResult.hash` (the Bearer), because `model_validate` picks up the `nvmApiKey` alias. `create_member` already maps `hash`; this does the same. Small one — mind a look?

r-marques:
:eyes: Reviewing

payments_py/api/organizations_api.py
```diff
@@ -320,7 +328,7 @@ def onboard_customer(self, email: str) -> CustomerOnboardingResponse:
                     )
                 },
             )
-        return result
+        return result.model_copy(update={"nvm_api_key": hash_key})
```
r-marques, line 334: **🟡 SHOULD FIX — Surface `expires_at` too — the paired docs PR tells integrators to track it**

Same gap as the TS sibling (nevermined-io/payments#420), raised here so the two SDKs don't drift apart on it.

The docs change shipping alongside these (nevermined-io/docs#301, `solutions/organizations/white-label-onboarding.mdx`) tells integrators:

> The exact expiry is returned as `expiresAt` (ISO-8601) — track it and re-onboard to refresh before it lapses.

A Python SDK user cannot. The backend returns it on **this exact outcome** — `organizations.controller.ts:173-178` documents `expiresAt` as *"ISO-8601 expiry of `hash` (#2924). Returned for the `as=customer` credential outcome"* — but `CustomerOnboardingResponse` (`common/types.py:685-703`) has no such field, so `model_validate` drops it and it never reaches the caller.

One field on the model is enough, and `model_validate` then populates it for free — no change needed at this line:

```python
expires_at: Optional[str] = Field(default=None, alias="expiresAt")
```

Optional, not required: it is documented for the credential outcome specifically, so an older backend omitting it should not become a second hard guard.

Should-fix rather than blocker — the credential bug is genuinely fixed here and that is worth shipping. But these PRs land together with the docs, and as it stands the docs promise something neither SDK provides.

r-marques (commented):

The Python half of the same fix as nevermined-io/payments#420, and it lands correctly: `onboard_customer` was returning `walletResult.nvmApiKey` — the encrypted server-side blob — as the credential. **The docstring's central claim checks out**: `create_member` really does use `nvm_api_key=wallet["hash"]` (`organizations_api.py:246`), so "Like `create_member`, the usable Bearer is `walletResult.hash`" is accurate rather than aspirational. Same `expires_at` gap as the TS sibling, plus one Python-specific note about how the override is applied.

**What I reviewed**
- Diff range `origin/main...HEAD` — 2 files, +26/−13, head `b9f65847`.
- Panel: none — my own verification against `nvm-monorepo@fd5af9a76` (the backend contract) and the SDK source. No panel agent contributed.
- **Confirmed the fix direction at the backend, not from the commit message**: `organizations.controller.ts:155-176` documents `hash` as the Bearer (`<env>:<JWT>`) and `nvmApiKey` as "Encrypted SDK/server-side blob — NOT a Bearer"; `organizations.service.ts:556-559` returns them as two distinct values.
- **Checked that no blob survives on the returned object.** `CustomerOnboardingResponse` (`common/types.py:685-703`) has no `hash` field, so `model_copy(update={"nvm_api_key": …})` replaces the only credential-shaped attribute — the validated-in blob is overwritten, not shadowed. The consent path returns a freshly constructed object, so it cannot leak either.
- Verified `wallet` is still in scope at line 320 and that the new guard keys on `hash`, so a payload carrying the blob but no `hash` now fails loudly instead of returning a useless success.
- **I did not run the tests** — I did not install the Poetry environment into the review worktree. The test changes are assessed by reading.

**Verdict:** 🟡 Mergeable with nits — 0 blockers, 1 should-fix, 1 nit.

### 🧭 Inline comments (3)

- 🟡 **SHOULD FIX** — Surface `expires_at` too — the paired docs PR tells integrators to track it — `payments_py/api/organizations_api.py:331`
- 💡 **NIT** — `model_copy(update=…)` skips validation the old path had — `payments_py/api/organizations_api.py:320`
- ✅ **NICE** — Corrected the test that asserted the defect, instead of leaving it green — `tests/unit/test_organizations_api.py:635`

payments_py/api/organizations_api.py
```diff
@@ -307,10 +309,16 @@ def onboard_customer(self, email: str) -> CustomerOnboardingResponse:
...
-        if not result.nvm_api_key:
+        # New account or returning customer: the Bearer MUST be present. The
+        # backend returns the usable credential as ``walletResult.hash``
+        # (``<prefix>:<jwt>``); the sibling ``nvmApiKey`` — which the
+        # ``model_validate`` above mapped into ``nvm_api_key`` — is the encrypted
+        # server-side blob and is NOT a valid Bearer. Surface ``hash`` instead,
+        # exactly as create_member does. A 2xx without a hash means a
+        # partial/unexpected payload — fail loudly rather than return a "success"
+        # carrying no usable credential.
+        hash_key = wallet.get("hash")
```
r-marques, line 323: **💡 NIT — `model_copy(update=…)` skips validation the old path had**

Small observation about what the change gives up, not a defect I can demonstrate.

Before, the credential reached `nvm_api_key` through `CustomerOnboardingResponse.model_validate(wallet)` — so Pydantic enforced the field's `Optional[str]` on the way in. Now it arrives via `wallet.get("hash")` → `model_copy(update=…)`, and **`model_copy` does not validate**: whatever `hash_key` holds is assigned as-is. The `if not hash_key` guard rejects only falsy values, so a non-string truthy `hash` (say the backend ever returned a structured object) would be stored on a field typed `str` and flow onward to `Payments` as the Bearer.

The backend contract says `<env>:<JWT>`, so this is theoretical today — hence a nit. If you want the guarantee back cheaply:

```python
if not isinstance(hash_key, str) or not hash_key:
```

which also collapses the falsy check into the same line. Equally fine to leave it and rely on the contract; worth a deliberate call rather than an accident of `model_copy`'s semantics.

tests/unit/test_organizations_api.py
```diff
@@ -626,8 +630,9 @@ def test_new_customer_returns_real_key_and_sends_as_customer(self):
         # Opts into the customer outcome.
         assert body == {"email": "customer@example.com", "as": "customer"}
         assert isinstance(result, CustomerOnboardingResponse)
-        # The USABLE key is returned — not the (non-usable) lookup hash.
-        assert result.nvm_api_key == "nvm-real-usable-key"
+        # The USABLE Bearer (``hash``) is surfaced as ``nvm_api_key`` (mirroring
+        # create_member) — the raw ``nvmApiKey`` blob must never leak through.
+        assert result.nvm_api_key == "sandbox:jwt-bearer-token"
```
r-marques, line 636: **✅ NICE — Corrected the test that asserted the defect, instead of leaving it green**

Same good instinct as the TS PR, worth recording because it is the failure mode that lets a bug outlive its fix. The old assertion was `result.nvm_api_key == "nvm-real-usable-key"` under the comment *"The USABLE key is returned — not the (non-usable) lookup hash"* — it encoded the **inverted** belief as though it were a safety property, so it passed *because* the code was wrong, and correcting the code alone would have turned it red and invited someone to revert the fix to get green back.

This PR rewrites the fixture's story rather than the assertion's expectation: `hash` becomes `"sandbox:jwt-bearer-token"` (the real `<prefix>:<jwt>` shape), `nvmApiKey` becomes `"encrypted-blob-not-a-bearer"`, and both carry comments naming which is which. The test now fails for the right reason if the blob is ever surfaced again.

r-marques (approved):
🟡 **Approved with comments** — no blockers; 1 should-fix, 1 nit(s) left inline. Auto-merge was disabled so these can be looked at first.

Inline review: https://github.com/nevermined-io/payments-py/pull/264#pullrequestreview-5049753936

payments_py/api/organizations_api.py
aaitor: Done in 3baf924 — added optional `expires_at` (alias `expiresAt`) to `CustomerOnboardingResponse`; `model_validate` populates it and `model_copy` preserves it, with a test asserting it. Mirrors the TS change in #420.

payments_py/api/organizations_api.py
aaitor: Adopted in 4d2b396 — guard is now `if not isinstance(hash_key, str) or not hash_key`, so a non-string truthy `hash` from a regressed backend can never reach `Payments` as the Bearer. Added a test that feeds `hash` a dict and expects the raise.

tests/unit/test_organizations_api.py
aaitor: Thanks — same reasoning as the TS PR: corrected the fixture's story so it fails for the right reason if the bug is reintroduced.

PR #267, opened 2026-09-01
test(x402): pin Deep Agents compatibility for @requires_payment

r-marques:
## Description

`@requires_payment` works unchanged under LangChain's [Deep Agents](https://docs.langchain.com/oss/python/deepagents/overview) harness — including on a tool that lives **inside a subagent**. Nothing guarded that, so this PR pins it with tests and documents it.

The load-bearing property:

```
main agent  --task()-->  research-sub  -->  market_research  [PAID]
     ^                                            |
     +--------- x402 token supplied here ---------+
                config.configurable.payment_token
```

A buyer supplies the token once, on the run. The supervisor never handles it — it delegates through Deep Agents' built-in `task` tool, and LangGraph copies `configurable` down into the subagent's tool calls. Without this, a paid tool could never sit behind a delegation, which is the harness's whole premise.

**No SDK code changes.** Tests and docs only.

### What's here

- **`tests/unit/x402/test_deepagents_compat.py`** — 3 tests driving a real deep agent through a scripted fake chat model (no network, no LLM): the token reaches a subagent's tool, verify → body → settle completes below the delegation, and a missing token raises `PaymentRequiredError` without settling.
- **`.github/workflows/test.yaml`** — a dedicated `deepagents_compat` job so those tests actually run (see below).
- **`docs/api/12-langchain-integration.md`** — a Deep Agents section covering the delegation hop and the two harness behaviours that need designing around.

### Why a separate CI job rather than a dependency bump

The tests `importorskip`, and they cannot join `unit_integration`:

- That job pins **Python 3.10** — the floor this package declares (`python = "^3.10"`) — while `deepagents` requires **>=3.11**.
- `deepagents` cannot go in the poetry lock at all: it needs `langgraph>=1.2.11` via langchain v1, which is **disjoint** from the test group's `langgraph = "^0.6.0"`. Poetry version solving fails outright.

So `deepagents_compat` runs on 3.11 with a pip install of the v1 stack. The 3.10 floor stays tested by the existing job, and the contract now actually runs.

### Verification

- Compat tests: **3 passed**, reproducing the exact CI recipe in a clean venv (deepagents 0.7.11 / langchain 1.3.18 / langgraph 1.2.11).
- Default poetry env: **1 skipped**, as intended.
- Existing suite unchanged: **273 passed, 1 skipped**.
- `mkdocs build` clean (the 94 `--strict` griffe warnings are pre-existing — verified identical on clean `main`).

Companion tutorial: [`langchain-deep-agent-py`](https://github.com/nevermined-io/tutorials/tree/r-marques/langchain-demo/langchain-deep-agent-py). Docs-site page: nevermined-io/docs `docs/deep-agents-integration`.

No.

## Types of changes

## Checklist:

- [x] Tests Cover Changes
- [x] Documentation

### DIFF
--- a/tests/unit/x402/test_deepagents_compat.py
+++ b/tests/unit/x402/test_deepagents_compat.py
@@ -1,0 +1,226 @@
+"""Compatibility tests: `@requires_payment` under the Deep Agents harness.
+
+The contract these tests pin down is narrow but load-bearing:
+
+    main agent  --task()-->  subagent  -->  paid tool
+
+A buyer supplies the x402 access token once, on the run, at
+``config["configurable"]["payment_token"]``. The supervisor never handles
+it — it delegates through Deep Agents' built-in ``task`` tool, and
+LangGraph copies ``configurable`` down into the subagent's own tool
+calls. ``requires_payment`` therefore works **unchanged one delegation
+hop away** from where the token was supplied.
+
+That property is what makes the harness usable for monetized
+capabilities at all: a deep agent's premise is that the supervisor hands
+work to subagents, so if payment context did not survive the hop, every
+paid tool would have to sit on the main agent.
+
+No network and no LLM: a scripted fake chat model drives the graph
+through a fixed tool-call sequence, and `Payments` is mocked. The tests
+assert on the SDK's own behaviour, not on a model's willingness to
+delegate.
+
+.. note::
+
+   ``deepagents`` requires the LangChain v1 stack (``langchain>=1.3.18``,
+   ``langchain-core>=1.6.1``), which is newer than this repo's test
+   environment pins. These tests therefore ``importorskip`` and are
+   **skipped in the default test run** — they guard the contract only
+   where ``deepagents`` is actually installed::
+
+       poetry run pip install deepagents
+       poetry run pytest tests/unit/x402/test_deepagents_compat.py
+"""
+
+from typing import Any, Sequence
+from unittest.mock import MagicMock
+
+import pytest
+from langchain_core.language_models import FakeMessagesListChatModel
+from langchain_core.messages import AIMessage
+from langchain_core.runnables import RunnableConfig
+from langchain_core.tools import tool
+
+from payments_py.x402.langchain import (
+    PaymentRequiredError,
+    last_settlement,
+    requires_payment,
+)
+from payments_py.x402.langchain import decorator as decorator_module
+from payments_py.x402.types import SettleResponse, VerifyResponse
+
+create_deep_agent = pytest.importorskip(
+    "deepagents",
+    reason="deepagents requires the LangChain v1 stack; install it to run these",
+).create_deep_agent
+
+
+SENTINEL_TOKEN = "x402-token-sentinel"
+
+
+class ScriptedModel(FakeMessagesListChatModel):
+    """Fake chat model that replays a fixed message script.
+
+    ``FakeMessagesListChatModel`` raises ``NotImplementedError`` from
+    ``bind_tools``, which the agent factory calls unconditionally. We
+    accept the tools and ignore them — the script already encodes which
+    tool calls to emit.
+    """
+
+    def bind_tools(self, tools: Sequence[Any], **kwargs: Any) -> "ScriptedModel":
+        return self
+
+
+@pytest.fixture
+def mock_payments():
+    """A Payments-like mock with verify + settle stubbed."""
+    payments = MagicMock()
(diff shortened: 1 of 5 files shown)


aaitor:
🤖 Fallback review bot picking this up.

tests/unit/x402/test_deepagents_compat.py
```diff
@@ -0,0 +1,233 @@
...
+        credits_redeemed="5",
+        remaining_balance="95",
+    )
+    return payments
+
+
+@pytest.fixture(autouse=True)
+def reset_last_settlement():
+    """Reset the module-level holder between tests so they don't bleed state."""
+    decorator_module._LAST_SETTLEMENT["value"] = None
```
aaitor: **nit — reset pokes a private module global.**

`decorator_module._LAST_SETTLEMENT["value"] = None` couples this autouse fixture to the exact internal holder shape. It's the pragmatic choice given `last_settlement()` reads module state and there's no public reset — fine as-is — but a refactor of how that value is stored would break this test even though the public contract is unchanged. If this reset pattern spreads, a tiny public test-reset helper would decouple it.

aaitor (commented):
Fallback review (payments-py#267) — tests + docs + a CI job, no production code. Read all three.

The compat contract is pinned well: a scripted fake model (no network, no LLM) drives a real deep agent, and the three tests assert the token survives the `task()` hop, settlement runs to completion one hop down, and a missing token raises `PaymentRequiredError` without settling. `importorskip` + the dedicated `deepagents_compat` job keep it guarded where the 3.10 floor / `langgraph ^0.6.0` lock can't host deepagents, and the pip step fails loudly rather than skipping vacuously. Docs API section matches the code samples I verified in docs#336.

No blockers, no should-fix. One nit inline.

aaitor (approved):
🟡 **Approved with comments** — no blockers; 0 should-fix, 1 nit(s) left inline.

Inline review: https://github.com/nevermined-io/payments-py/pull/267#pullrequestreview-5077861833

tests/unit/x402/test_deepagents_compat.py
r-marques: Folded in `fb72fa1` — your condition ("if this reset pattern spreads") is exactly what this PR did, so it was worth doing now rather than later: the fixture already existed in `test_langchain_decorator.py` and this PR copied it into the new module.

Moved it to `tests/unit/x402/conftest.py` and deleted both copies, so the private `_LAST_SETTLEMENT` coupling now lives in exactly one place — a change to how that value is stored breaks one fixture instead of every module that needs a clean slate. The import is lazy, so the conftest stays importable (and the rest of the x402 tests stay collectable) without the optional `langchain-core` extra.

I deliberately did **not** add a public test-reset helper to the SDK. That's new public API surface and deserves its own decision rather than riding in on a test PR — happy to file it if you'd like it.

273 passed / 1 skipped in the default env; 3 passed in the deepagents env.

One unrelated thing I noticed while verifying this, worth knowing rather than acting on here: `TestCreatePaidReactAgent::test_tool_node_has_handle_tool_errors_disabled` **fails under langgraph 1.x** — it asserts on `ToolNode.handle_tool_errors`, a langgraph internal that changed. It's pre-existing (fails identically before my commit) and the `deepagents_compat` job doesn't run that file, so nothing here is affected. But it means that test won't survive whenever we do bump the test group off `langgraph ^0.6.0`.

aaitor (commented):
Re-review (round 2) — verified the delta since my approval (b697ad31..fb72fa1).

The round-1 nit is resolved cleanly: the `reset_last_settlement` fixture now lives in one shared `tests/unit/x402/conftest.py` and both copies are gone (0 references left in `test_deepagents_compat.py` and `test_langchain_decorator.py`). The `_LAST_SETTLEMENT` coupling is unavoidable without a public reset, but it's now in exactly one place — which was the ask.

Bonus worth calling out: the lazy `try/except ImportError` guard is the right move — a top-level import of the langchain decorator in a directory-wide conftest would have broken *collection* of every x402 test in the 3.10 job that lacks the optional `langchain-core` extra. Guarding it keeps the whole dir collectable while still resetting where the module is present.

Agreed on not adding a public SDK reset helper here — that's new public API and shouldn't ride in on a test PR. The pre-existing `test_tool_node_has_handle_tool_errors_disabled` langgraph-1.x failure is out of scope for this delta. Still an approve.

aaitor (approved):

Inline review: https://github.com/nevermined-io/payments-py/pull/267#pullrequestreview-5077973623

PR #270, opened 2026-09-06
feat(orders): OrdersAPI client for browser-fiat Orders (create_order / get_order)

aaitor:
## Why this matters

Merchants can now charge a customer an arbitrary amount by card straight from the Nevermined Python SDK, without first publishing a payment plan or asking the buyer to create a Nevermined account. One call creates the order and returns the Stripe client secret the checkout page needs; a second call reads the order's status so the merchant can fulfil once it is paid. This delivers the Python half of nvm-monorepo#3250 (P1.8) for the "dynamic pricing: goods & services" epic (#3238).

### What

`payments.orders` (new `OrdersAPI`) wrapping the Orders endpoints shipped in nvm-monorepo#3311:

- `create_order(amount_minor, currency="usd", *, description=None, buyer_ref=None, idempotency_key=None, line_items=None, metadata=None, capture_mode=None, payment_provider=None)` → `POST /api/v1/orders` (merchant, authenticated). Returns a `CreateOrderResult` (`order_id`, `status`, `client_secret`).
- `get_order(order_id)` → `GET /api/v1/orders/{id}` (anonymous, public HTTP options — the unguessable id is the access control). Returns the buyer-safe `Order` model.
- New guide `docs/api/16-orders.md` (+ mkdocs nav) covering the flow, the org-scoped-key requirement, the status lifecycle and the `BCK.ORDER.*` codes.

Modelled on `PlansAPI` / `OrganizationsAPI`: same `BasePaymentsAPI` request layer (`get_backend_http_options` / `get_public_http_options`), same `Bearer` + `Nevermined-Version` headers, errors surfaced via `PaymentsError.from_response(...)` so `err.code` is the catalogue code (`BCK.ORDER.0001..0011`). No new error types, no hand-rolled HTTP, no server changes.

### Decisions (left to the implementer by #3250)

- **Keyword-only optionals** are Python's typed options object: one required positional (`amount_minor`), `currency` defaulting to `"usd"` (Phase 1 is USD-only; the wire still requires it), everything else keyword-only. `get_order(order_id)` stays positional like `get_plan(plan_id)`.
- **Body serialised directly, not via `get_backend_http_options(body=...)`.** The shared helper camelCases dict keys *recursively*, which would rewrite the merchant's opaque `metadata` / `line_items` keys (`unit_price` → `unitPrice`) — data the contract says is recorded verbatim. The wire keys are already camelCase, and the SDK-wide `_stringify_unsafe_ints` policy is still applied before `json.dumps` so an int above 2**53-1 anywhere in the payload goes out as a decimal string rather than being rounded by the Node backend. Headers / auth / timeout still come from the base layer. Unit tests pin both the verbatim keys and the big-int case.
- **Full response typing** with two pydantic models, `CreateOrderResult` and `Order`, mirroring `CreateOrderResponseDto` / `OrderResponseDto` (camelCase aliases + `populate_by_name`, the `MyMembership` pattern). `status` is `Union[OrderStatus, str]` parsed left-to-right: known values become the new `OrderStatus(str, Enum)`, a value the backend adds later degrades to a bare string instead of raising (the `MyMembership.org_type` forward-compat pattern). `amount_refunded_minor` (a money field) and the four always-present nullable fields carry **no default**, exactly as the DTO's required-but-nullable shape, so a contract regression fails loudly instead of silently reading `0`. No merchant-identity or fee fields, per the buyer-safe contract.
- **No client-side range validation.** The server owns the bounds (`BCK.ORDER.0001`; the max is env-configurable), so duplicating them would only drift.
- **Tests: unit, `requests_mock` style** (same pattern as `test_organizations_api.py`): URL, headers (Bearer present on create, absent on get, `X-Current-Org-Id` forwarded from the instance pin), the exact JSON body including verbatim opaque keys and stringified big ints, `client_secret` absent when withheld, enum / unknown-status parsing, missing-money-field failure, and catalogue-code propagation for `BCK.ORDER.0003` / `0002`. No e2e: creating an order on staging needs an org-scoped key with an active org and a Connect account, which the CI test keys are not, and staging may not carry #3311 yet (argocd#611).

### Escalations (per the #3250 brief) — @aaitor

1. **No release / publish.** This PR stops at merge; no version bump. Cutting `payments-py` with this needs your explicit go-ahead.
2. **Org-scoped API key.** The API derives the merchant organization from the key's own org tag, not from the `X-Current-Org-Id` header. The SDK's key model is "any NVM API key string", so an org-scoped key is expressible today with no change: construct `Payments` with it. `set_organization_id()` on a personal key does **not** substitute (the org gate #3317 ignores the header → `BCK.ORDER.0003`). Documented in the module docstring and the guide; no workaround attempted.

Related to nevermined-io/nvm-monorepo#3250 (P1.8 — SDK OrdersAPI), epic nevermined-io/nvm-monorepo#3238.

## Types of changes

## Checklist:

- [x] Tests Cover Changes
- [x] Documentation — `docs/api/16-orders.md` guide + nav entry; docstrings on `OrdersAPI` and the models publish under `docs/reference/`

### Test plan

- [x] `poetry run black --check .` and `pre-commit run --all-files` — pass
- [x] `poetry run pytest -m "not slow"` — 979 passed, 1 skipped
- [x] `poetry run mkdocs build` — exit 0 (warnings pre-existing)
- [x] Deep pre-PR review — first pass 0 blockers / 3 should-fix / 2 nits, all addressed in the third commit; the re-run hit the daily review budget, so human review + CI are the second pass
- [ ] CI green

Refs nevermined-io/nvm-monorepo#3250 · epic #3238 · server contract nvm-monorepo#3311 · org gate #3317. Companion PR: nevermined-io/payments#429 (TypeScript SDK).

https://claude.ai/code/session_01CSHvoA2jHxfARTVwEmrR33

### DIFF
--- a/docs/api/16-orders.md
+++ b/docs/api/16-orders.md
@@ -60,15 +60,17 @@ print(order.client_secret)          # present only while the Order is payable
 
 The read is anonymous on the wire (the unguessable id is the access control) and returns a buyer-safe projection: it never includes the merchant identity, the Connect account or the fee.
 
+The read endpoint is rate-limited: all anonymous callers behind one IP share a bucket of 60 requests per minute. A throttled call raises `PaymentsError` with code `http_429` and no catalogue code (this is not the create-side `BCK.ORDER.0010`). Poll sparingly and back off on `http_429`.
+
 ## Order lifecycle
 
 | Status | Meaning |
 |---|---|
 | `requires_payment` | Created; the browser has not confirmed yet. `client_secret` is available. |
-| `paid` | Payment succeeded. |
+| `paid` | Payment succeeded. Also the state after a **won** dispute. |
 | `failed` | Payment failed, or the PaymentIntent could not be created. No money moved. |
 | `refunded` / `partially_refunded` | Refunded in full / in part (see `amount_refunded_minor`). |
-| `disputed` | A chargeback is open. |
+| `disputed` | A chargeback is open, or was lost. A won dispute returns the Order to `paid`. |
 
 ## Error codes
 
@@ -82,7 +84,10 @@ Errors raise `PaymentsError` with `code` set to the backend catalogue code:
 | `BCK.ORDER.0004` | 500 | The merchant has no Connect account able to receive card payments. |
 | `BCK.ORDER.0005` | 500 | The PaymentIntent could not be created; the Order is `failed`, no money moved. |
 | `BCK.ORDER.0007` | 409 | Idempotency-key conflict. |
-| `BCK.ORDER.0010` | 429 | Velocity cap exceeded. Retry after backoff. |
+| `BCK.ORDER.0010` | 429 | Velocity cap exceeded on `create_order`. Retry after backoff. |
+| `http_429` | 429 | The `get_order` read throttle (no catalogue code). Back off and retry. |
+
+A refusal that carries no catalogue code (a throttle or gateway response) surfaces with code `http_<status>`.
 
 ```python
 from payments_py import PaymentsError
--- a/docs/api/16-orders.md
+++ b/docs/api/16-orders.md
@@ -1,0 +1,96 @@
+# Orders
+
+This guide covers browser-fiat **Orders**: a merchant-initiated, off-plan charge for an arbitrary amount (the Stripe PaymentIntent analog) that the buyer's browser confirms client-side. No payment plan, no buyer Nevermined account, no delegation.
+
+## Overview
+
+1. `payments.orders.create_order(...)` — the merchant creates the Order server-side and receives a Stripe `client_secret`.
+2. The merchant hands the `client_secret` to its checkout page, which confirms the payment with Stripe.js.
+3. `payments.orders.get_order(order_id)` — anyone holding the unguessable `order_id` reads the buyer-safe status. No API key is sent on the wire.
+
+## Requirements
+
+`create_order` needs an **organization-scoped** NVM API key: the API resolves the merchant organization from the key's own org tag. A personal key is refused with `BCK.ORDER.0003`, and `payments.set_organization_id(...)` does not substitute for an org-scoped key. The organization must be active and have a Stripe Connect account able to receive card payments.
+
+## Create an Order
+
+```python
+from payments_py import Payments, PaymentOptions
+
+payments = Payments.get_instance(
+    PaymentOptions(nvm_api_key="nvm:your-org-scoped-key", environment="sandbox")
+)
+
+created = payments.orders.create_order(
+    3437,  # USD cents: $34.37 ($1.00 - $999,999.99)
+    description="Cart checkout - 3 items",
+    buyer_ref="merchant-order-4821",
+    idempotency_key="merchant-order-4821",
+    line_items=[{"sku": "PRO-PLAN", "quantity": 1, "amountMinor": 3437}],
+    metadata={"channel": "web"},
+)
+
+print(created.order_id)       # ord_... - the buyer-facing access control
+print(created.status)         # requires_payment
+print(created.client_secret)  # hand this to the browser (Stripe.js confirm)
+```
+
+| Parameter | Type | Description |
+|---|---|---|
+| `amount_minor` | `int` | Charge amount in USD cents, `100` to `99_999_999`. |
+| `currency` | `str` | ISO currency, lower-cased. Default and only value in Phase 1: `"usd"`. |
+| `description` | `str` | Optional. Human-readable description (max 1024 chars). |
+| `buyer_ref` | `str` | Optional. Your own reference for the buyer or cart (max 255 chars). |
+| `idempotency_key` | `str` | Optional. A retried create with the same key returns the same Order and `client_secret`; a conflicting body is refused with `BCK.ORDER.0007`. |
+| `line_items` | `list[dict]` | Optional. Recorded verbatim, opaque to the API. |
+| `metadata` | `dict` | Optional. Recorded verbatim, opaque to the API. |
+| `capture_mode` | `str` | Optional. `"automatic"` only in Phase 1. |
+| `payment_provider` | `str` | Optional. `"stripe"` only in Phase 1. |
+
+## Read an Order
+
+```python
+order = payments.orders.get_order(created.order_id)
+
+print(order.status)                 # OrderStatus.REQUIRES_PAYMENT, PAID, ...
+print(order.amount_minor)           # 3437
+print(order.amount_refunded_minor)  # 0
+print(order.client_secret)          # present only while the Order is payable
+```
+
+The read is anonymous on the wire (the unguessable id is the access control) and returns a buyer-safe projection: it never includes the merchant identity, the Connect account or the fee.
+
+## Order lifecycle
+
+| Status | Meaning |
+|---|---|
+| `requires_payment` | Created; the browser has not confirmed yet. `client_secret` is available. |
+| `paid` | Payment succeeded. |
+| `failed` | Payment failed, or the PaymentIntent could not be created. No money moved. |
+| `refunded` / `partially_refunded` | Refunded in full / in part (see `amount_refunded_minor`). |
+| `disputed` | A chargeback is open. |
+
+## Error codes
+
+Errors raise `PaymentsError` with `code` set to the backend catalogue code:
+
+| Code | HTTP | Meaning |
+|---|---|---|
+| `BCK.ORDER.0001` | 400 | Invalid request (amount out of range, unsupported currency / capture mode / provider). |
+| `BCK.ORDER.0002` | 404 | No Order with this id. |
+| `BCK.ORDER.0003` | 403 | The key is not an active organization, or the amount exceeds the per-order cap. |
+| `BCK.ORDER.0004` | 500 | The merchant has no Connect account able to receive card payments. |
+| `BCK.ORDER.0005` | 500 | The PaymentIntent could not be created; the Order is `failed`, no money moved. |
+| `BCK.ORDER.0007` | 409 | Idempotency-key conflict. |
+| `BCK.ORDER.0010` | 429 | Velocity cap exceeded. Retry after backoff. |
+
+```python
+from payments_py import PaymentsError
+
+try:
+    created = payments.orders.create_order(3437)
+except PaymentsError as e:
+    if e.code == "BCK.ORDER.0010":
+        ...  # back off and retry
+    raise
+```
--- a/docs/api/16-orders.md
+++ b/docs/api/16-orders.md
@@ -41,7 +41,7 @@ print(created.client_secret)  # hand this to the browser (Stripe.js confirm)
 | `currency` | `str` | ISO currency, lower-cased. Default and only value in Phase 1: `"usd"`. |
 | `description` | `str` | Optional. Human-readable description (max 1024 chars). |
 | `buyer_ref` | `str` | Optional. Your own reference for the buyer or cart (max 255 chars). |
-| `idempotency_key` | `str` | Optional. A retried create with the same key returns the same Order and `client_secret`; a conflicting body is refused with `BCK.ORDER.0007`. |
+| `idempotency_key` | `str` | Optional. A retry with the same key returns the original Order unchanged (and its `client_secret` while the Order is still payable); the other fields of the retry are ignored, not merged. A retry with a different `amount_minor` or `currency` is refused with `BCK.ORDER.0007`. |
 | `line_items` | `list[dict]` | Optional. Merchant-defined structure, recorded verbatim (keys are not transformed), opaque to the API. |
 | `metadata` | `dict` | Optional. Merchant-defined structure, recorded verbatim (keys are not transformed), opaque to the API. |
 | `capture_mode` | `str` | Optional. `"automatic"` only in Phase 1. |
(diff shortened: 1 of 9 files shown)


tests/unit/test_orders_api.py
```diff
@@ -0,0 +1,221 @@
...
+    OrderStatus,
+    PaymentOptions,
+)
+from payments_py.environments import Environments
+from payments_py.payments import Payments
+
+TEST_API_KEY = os.getenv(
+    "TEST_PROXY_BEARER_TOKEN",
+    "sandbox-staging:eyJhbGciOiJFUzI1NksifQ.eyJpc3MiOiIweDU4MzhCNTUxMmNGOWYxMkZFOWYyYmVjY0IyMGViNDcyMTFGOUIwYmMiLCJzdWIiOiIweEVCNDk3OTU2OTRBMDc1QTY0ZTY2MzdmMUU5MGYwMjE0Mzg5YjI0YTMiLCJqdGkiOiIweGMzYjYyMWJkYTM5ZDllYWQyMTUyMDliZWY0MDBhMDEzYjM1YjQ2Zjc1NzM4YWFjY2I5ZjdkYWI0ZjQ5MmM5YjgiLCJleHAiOjE3OTQ2NTUwNjAsIm8xMXkiOiJzay1oZWxpY29uZS13amUzYXdpLW5ud2V5M2EtdzdndnY3YS1oYmh3bm1pIn0.YMkQUjGh7_m07nj8SKXZReNKSryg9mTU3qwJr_TKYATUixbYQTte3CKucjqvgAGzJAd1Kq2ubz3b37n5Zsllxs",
+)
```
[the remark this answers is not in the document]
aaitor: Done in 883b1ca — the test now mints a throwaway, locally-signed JWT per run (`sub` + `o11y`, `sandbox-staging:` prefix), the same approach as `test_user_scoped_listing.py`. For context, the previous value was the identical sandbox fixture already committed in `test_payments.py` / `test_organizations_api.py`, not a new credential — but not repeating it is strictly better.

docs/api/16-orders.md
```diff
@@ -0,0 +1,96 @@
...
+
+```python
+from payments_py import Payments, PaymentOptions
+
+payments = Payments.get_instance(
+    PaymentOptions(nvm_api_key="nvm:your-org-scoped-key", environment="sandbox")
+)
+
+created = payments.orders.create_order(
+    3437,  # USD cents: $34.37 ($1.00 - $999,999.99)
```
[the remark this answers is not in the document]
aaitor: Done in 883b1ca — the sample comment is now just `# USD cents: $34.37`. The one remaining mention (the parameter table) states the fixed $1.00 floor and points to the server-validated ceiling / per-order cap (`BCK.ORDER.0001` / `BCK.ORDER.0003`) instead of the exact upper bound, so there is a single place to keep in sync. Same change applied to the TS guide in payments#429.

docs/api/16-orders.md
```diff
@@ -0,0 +1,96 @@
...
+    description="Cart checkout - 3 items",
+    buyer_ref="merchant-order-4821",
+    idempotency_key="merchant-order-4821",
+    line_items=[{"sku": "PRO-PLAN", "quantity": 1, "amountMinor": 3437}],
+    metadata={"channel": "web"},
```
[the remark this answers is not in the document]
aaitor: Done in 883b1ca — the example is now `{"sku": "PRO-PLAN", "quantity": 1, "unit_price": 3437}`, and the `line_items` / `metadata` rows say "merchant-defined structure, recorded verbatim (keys are not transformed)". Mirrored in the TS guide.

payments_py/api/orders_api.py
```diff
@@ -0,0 +1,150 @@
...
+        payment_provider: Optional[str] = None,
+    ) -> CreateOrderResult:
+        """Create an Order and its PaymentIntent (``POST /api/v1/orders``).
+
+        Oriented to merchants: the NVM API key must be scoped to an active
+        organization whose Stripe Connect account can receive card payments.
+
+        Args:
+            amount_minor: Charge amount in USD cents, $1.00–$999,999.99
+                (``100``–``99_999_999``).
```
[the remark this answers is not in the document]
aaitor: Done in 883b1ca — the docstring now reads: USD cents, at least `100` ($1.00); the API validates the upper bound (`BCK.ORDER.0001`) and a deployment may enforce a lower per-order cap (`BCK.ORDER.0003`). Same wording on the TS `CreateOrderOptions.amountMinor` TSDoc.

r-marques:
:eyes: Reviewing

docs/api/16-orders.md
```diff
@@ -0,0 +1,101 @@
...
+print(created.client_secret)  # hand this to the browser (Stripe.js confirm)
+```
+
+| Parameter | Type | Description |
+|---|---|---|
+| `amount_minor` | `int` | Charge amount in USD cents (at least `100`, i.e. $1.00). The API validates the upper bound (`BCK.ORDER.0001`); a deployment may enforce a lower per-order cap (`BCK.ORDER.0003`). |
+| `currency` | `str` | ISO currency, lower-cased. Default and only value in Phase 1: `"usd"`. |
+| `description` | `str` | Optional. Human-readable description (max 1024 chars). |
+| `buyer_ref` | `str` | Optional. Your own reference for the buyer or cart (max 255 chars). |
+| `idempotency_key` | `str` | Optional. A retried create with the same key returns the same Order and `client_secret`; a conflicting body is refused with `BCK.ORDER.0007`. |
```
r-marques: **🟡 SHOULD FIX — "a conflicting body" is broader than the guard the API actually applies**

The API decides an idempotency conflict on exactly two fields:

```ts
// nvm-monorepo apps/api/src/orders/orders.service.ts (main)
if (
  String(existing.amountMinor) !== String(dto.amountMinor) ||
  existing.currency !== dto.currency
) {
  NVMException.throw('BCK.ORDER.0007')
}
const clientSecret = await this.clientSecretIfPayable(existing)
return CreateOrderResponseDto.from(existing, clientSecret)
```

So "a conflicting body" only means a conflicting **amount or currency**. Reuse the key with the same total but different `line_items`, `description`, `buyer_ref` or `metadata` and the call returns `201` with the **original** Order — original line items and all — with no error anywhere.

Failure scenario: a merchant re-prices a cart, the total happens to land on the same value, they resend under the same `idempotency_key`, and they ship against line items the buyer never agreed to. This row is what told them that couldn't happen.

Suggested: "…a retry with a different **amount or currency** is refused with `BCK.ORDER.0007`; a retry with the same amount and currency returns the original Order unchanged — the other fields of the retried request are ignored, not merged."

Same wording appears on `docs#362` and I've raised it there too, so the two pages can be fixed together.

<sub>🤖 pr-review-toolkit · self · high</sub>

payments_py/api/orders_api.py
```diff
@@ -0,0 +1,162 @@
...
+                i.e. $1.00). The API validates the upper bound
+                (``BCK.ORDER.0001``); a deployment may enforce a lower
+                per-order cap (``BCK.ORDER.0003``).
+            currency: ISO currency, lower-cased. Phase 1 is USD-only.
+            description: Human-readable description of the charge (max 1024
+                chars).
+            buyer_ref: Opaque merchant-supplied buyer reference, e.g. the
+                merchant's own order id (max 255 chars).
+            idempotency_key: A retried create with the same key returns the
+                same Order and ``client_secret`` (max 255 chars).
```
r-marques: **💡 NIT — The replayed `client_secret` is returned only while the Order is still payable**

"A retried create with the same key returns the same Order and ``client_secret``" holds inside the payable window only. The replay path calls `clientSecretIfPayable(existing)` and `CreateOrderResponseDto.from` omits the field when that yields nothing, so a retry after `expires_at` — or after the buyer's attempt moved the Order to `failed` — returns the same `order_id` with `client_secret` as `None`.

The model already handles it correctly (`client_secret: Optional[str] = Field(default=None, ...)`, and `test_omits_client_secret_when_backend_withholds_it` covers it), so this is only the docstring reading as unconditional. "…returns the same Order, and the same ``client_secret`` while the Order is still payable" closes it.

<sub>🤖 pr-review-toolkit · self · high</sub>

tests/unit/test_orders_api.py
```diff
@@ -0,0 +1,253 @@
...
+        # ``amountRefundedMinor`` is always present on the wire; a missing value
+        # is a contract regression and must never silently read as 0.
+        payments = _make_payments()
+        body = {k: v for k, v in ORDER.items() if k != "amountRefundedMinor"}
+        with requests_mock.Mocker() as m:
+            m.get(f"{BACKEND}/api/v1/orders/{ORDER_ID}", json=body)
+            with pytest.raises(ValidationError):
+                payments.orders.get_order(ORDER_ID)
+
+    def test_url_encodes_the_order_id(self):
```
r-marques, line 215: **✅ NICE — Both fixes are here ahead of the review, and both are pinned by a test**

This PR already carries the two things I raised on the TypeScript sibling (`payments#429`) an hour ago, and it goes further by pinning each one:

- `quote(order_id, safe="")` in `get_order`, with `test_url_encodes_the_order_id` holding it. The TS client still interpolates the raw id.
- The read throttle is documented as 60/min on a shared ingress-IP bucket, surfacing as `http_429` with no catalogue code and explicitly distinguished from the create-side `BCK.ORDER.0010` — and `test_surfaces_the_read_throttle_without_catalogue_code_as_http_429` pins it. I checked the claim against `PaymentsError.from_response`, which falls back to `error_code = f"http_{response.status_code}"`, so it is accurate.

The status table is also right where `docs#362` is wrong: it states that a won dispute returns the Order to `paid`, which matches `toWireStatus` (`case 'won': return 'paid'`).

Two model decisions worth keeping as-is: `amount_refunded_minor` has no default so a missing money field raises instead of silently reading `0` (`test_fails_loudly_when_a_money_field_is_missing`), and `Union[OrderStatus, str]` with `union_mode="left_to_right"` lets a status the backend adds later validate as a plain string rather than blowing up an SDK that hasn't shipped the enum value yet. The `_stringify_unsafe_ints`-instead-of-recursive-camelCase note in `create_order` is the kind of comment that stops someone "simplifying" it back into a bug that would rewrite opaque merchant metadata keys.

<sub>🤖 pr-review-toolkit · self · high</sub>

r-marques (commented):

The Python mirror of `payments#429`: an `OrdersAPI` over the two browser-fiat Orders endpoints, pydantic models for both response shapes, a docs page and 12 unit tests. This one arrived with the two corrections I'd just raised on the TypeScript sibling already applied *and* pinned by tests, and its status table is right where `docs#362`'s is wrong. One doc claim still overstates a backend guard.

**What I reviewed**
- Diff range `origin/main...HEAD` — 9 files, +607/−0, head `4aaf6f9e`. (The head moved from `883b1ca1` between claiming and setup; everything below is against `4aaf6f9e`.)
- Panel: **none** — the operator's standing session instruction forbids spawning subagents, so every check below is my own reading, not a panel's.
- Ran the new suite: `pytest tests/unit/test_orders_api.py` — **12 passed**. (Runner: pytest from the base repo's poetry venv, py3.14, with `PYTHONPATH` pointed at the review worktree; a fresh `poetry install` in the worktree wasn't needed.)
- Checked both models against the backend DTOs on `nvm-monorepo@main`: `OrderStatus` matches `ORDER_STATUS_VALUES` exactly; `CreateOrderResult` and `Order` match `CreateOrderResponseDto` / `OrderResponseDto` field-for-field, including which fields are required-but-nullable (`description`, `buyer_ref`, `payment_intent_id`, `expires_at` carry no default) versus payable-only optional (`client_secret`). Correctly omits every withheld field.
- Verified the docstring's `http_429` claim at the source: `PaymentsError.from_response` sets `error_code = f"http_{response.status_code}"` when the body isn't an NVMException envelope, so a Nest `ThrottlerException` really does surface as `http_429`. And `ORDERS_GET_THROTTLE = { long: { limit: 60, ttl: 60_000 } }` with the shared-ingress-IP comment confirms the 60/min figure.
- Confirmed `get_public_http_options` sets only `Accept`, `Content-Type` and no auth header, and that `get_backend_http_options` sets `Content-Type: application/json` unconditionally — so passing `data=payload` separately from the options is correct, not a missing-header bug.
- Checked the wiring: `self.orders` is constructed in `_initialize_api`, and `set_organization_id` reaches it through the `vars(self)` fan-out rather than an explicit line — `test_orders_is_wired_and_forwards_the_org_pin_to_the_wire` proves that actually works. Exported from both `payments_py/__init__.py` and `payments_py/api/__init__.py`; `mkdocs.yml` nav entry added.
- Read `body.update({k: v for k, v in optional.items() if v is not None})` closely: `is not None` rather than truthiness is the right test here, since it preserves an explicitly-passed empty `line_items=[]` or `metadata={}`.

**Verdict:** 🟡 Mergeable with nits — 0 blockers, 1 should-fix, 1 nit.

### 🧭 Inline comments (3)

- 🟡 **SHOULD FIX** — "a conflicting body" is broader than the guard the API actually applies — `docs/api/16-orders.md:44`
- 💡 **NIT** — The replayed `client_secret` is returned only while the Order is still payable — `payments_py/api/orders_api.py:80`
- ✅ **NICE** — Both fixes are here ahead of the review, and both are pinned by a test — `tests/unit/test_orders_api.py:215`

r-marques (approved):
🟡 **Approved with comments** — no blockers; 1 should-fix, 1 nit(s) left inline.

Inline review: https://github.com/nevermined-io/payments-py/pull/270#pullrequestreview-5124758760

docs/api/16-orders.md
aaitor: Fixed in 1d5bb59 — the `idempotency_key` row now reads: a retry with the same key returns the original Order unchanged (and its `client_secret` while the Order is still payable); the other fields of the retry are ignored, not merged; a retry with a different `amount_minor` or `currency` is refused with `BCK.ORDER.0007`. The `BCK.ORDER.0007` error row says the same. Same correction applied to the docstring here and to the TS guide + `CreateOrderOptions.idempotencyKey` TSDoc in payments#429 (adb69b9).

payments_py/api/orders_api.py
aaitor: Fixed in 1d5bb59 — the docstring now says the replay returns the same `client_secret` only while the Order is still payable (it is `None` after `expires_at` or once the Order moved to `failed`). Mirrored in the TS TSDoc.

tests/unit/test_orders_api.py
aaitor: Thanks. For the record, the TS sibling caught up before this review landed: payments#429 `9a905c1` adds `encodeURIComponent(orderId)` and the `http_<status>` fallback code (with `http_429` + encoded-id tests), so the two SDKs now surface the read throttle the same way.

PR #272, opened 2026-09-08
feat(x402): surface billing_model so a pay-as-you-go settle can be read as paid

r-marques:
> **This is no longer a docs-only PR.** It now carries an **additive behaviour change** to the MCP paywall (`_meta["nevermined/credits"]` gains `billingModel` and `orderTx`) alongside the model + documentation work. Nothing is renamed, retyped or removed, and both new fields are omitted when absent — but it warrants a **minor**, not a patch. Titled `feat(...)` accordingly.

## The problem, in plain terms

On a **pay-as-you-go** plan, a settle that charged the buyer $1.00 comes back like this:

```json
{ "success": true, "network": "stripe", "billingModel": "pay-as-you-go",
  "creditsRedeemed": "0", "remainingBalance": "0",
  "orderTx": "pi_3U6tgrBYvSRKcV421ehH4bnX" }
```

Those plans hold no credit balance — each settle is one direct charge, referenced by `order_tx` (fiat rails) or `transaction` (crypto rails). So the success check a credits integration already has, `success and credits_redeemed > 0`, is **unreachable** there: it reports a real charge as a decline.

That is not hypothetical. During company testing of the card rails a tester's pass condition was exactly that check, the charge succeeded, and the agent running the purchase had to reason from the backend source to work out that a healthy payment was being reported as a decline. A less careful consumer would have retried — and on a card rail a needless retry is the one thing we tell people never to do, because repeated attempts feed issuer fraud scoring.

The API has always returned the discriminator `billingModel`. **`SettleResponse` did not declare it**, so pydantic dropped it on parse and the correct check could only be worked out by reading our source. This PR makes it available and documents the criterion.

This is the Python half; the TypeScript half is nevermined-io/payments#432.

## What changed

**No wire change.** `billingModel` was already on the response; the model just did not declare it.

- `SettleResponse` gains `billing_model` (alias `billingModel`), exported alongside the `X402BillingModel` literal alias from both `payments_py` and `payments_py.x402`.
- **The field is typed `str`, not the `Literal`** — deliberately, matching how `network` and `scheme` are handled in this module. A billing model a given SDK release has not heard of must not turn a settle that charged the buyer into a `ValidationError`. `X402BillingModel` is what consumers compare against; the docstring says so.
- The class docstring carries the per-model criterion, including the string-vs-number trap: these are **strings**, so `"0"` is truthy while `int("0") > 0` is false — two plausible-looking checks disagree.
- `credits_redeemed` / `remaining_balance` now say they are always `"0"` on pay-as-you-go. `order_tx` and `network` descriptions were corrected against the API's own DTO (`order_tx` is the per-request charge on fiat pay-as-you-go, not only an auto-top-up hash; `network` is the rail, not the billing model).
- Three places reported `0` on a successful pay-as-you-go charge and now branch on `billing_model`: the `facilitator_api.py` module example, `facilitator.py`'s `"✅ Payment settled successfully! Credits redeemed: …"` log, and the AgentCore interceptor's `"Settlement successful: N credits, remaining: M"` log.
- `docs/api/08-validation-of-requests.md` and `docs/api/11-x402.md`: a "Was the buyer charged?" section each, and their `SettleResponse` field tables now list `network`, `billing_model` and `order_tx`. `docs/reference/` is generated from docstrings and was not hand-edited.

## The criterion, as documented

| `billing_model` | Success criterion | Credit fields |
| --- | --- | --- |
| `credits` | `success` and `int(credits_redeemed) > 0` | `credits_redeemed` is the amount burned, `remaining_balance` what is left |
| `pay-as-you-go` | `success` **and** a non-empty `order_tx` (fiat) / `transaction` (crypto) | always the string `"0"` — no balance exists on this plan shape |

## Additive behaviour change: the MCP receipt

**payments-py does have an equivalent condensed key**, so leaving it alone would have diverged the two SDKs. `_meta["nevermined/credits"]` reported `creditsRedeemed` with nothing to read it by; on a pay-as-you-go plan that is always `"0"`, so a settle that **did** charge the buyer was indistinguishable there from a credits settle that burned nothing.

- The settle helper's two real-result return sites now carry `billingModel` and `orderTx`; `nvm_meta` attaches each **only when present**, so a consumer can distinguish "absent" from "present and empty". Existing keys keep their names, types and values.
- **Only the non-streaming site exists here.** The streaming path has no `_meta` channel at all — documented in `_RedeemOnCloseAsyncIterator` as a tracked follow-up from an earlier review — so there is no streaming sibling to update. Flagging that explicitly rather than leaving it to be inferred from the diff.

### One thing worth a reviewer's attention

Four test doubles claimed to mimic the settlement model while lacking two attributes the real `SettleResponse` has. The moment the paywall read them they raised `AttributeError` — 28 tests went red. I fixed the **doubles** rather than reaching for `getattr(..., None)` in the paywall: a double that cannot express the pay-as-you-go shape cannot test it, and defensive reads in production code driven by test fixtures hide exactly this kind of drift. Both new kwargs default to `None`, so every existing call site is unaffected.

### Mutation-tested, recording which tests go red

| Mutation | Tests killed |
| --- | --- |
| `billingModel` dropped from `nvm_meta` | the PAYG receipt test |
| `orderTx` dropped from `nvm_meta` | the PAYG receipt test |
| `billingModel` dropped from the settle helper | the PAYG receipt test |
| *(control: restored)* | none — 982 passed |

## Verification

- `poetry run pytest -m "not slow"` — **982 passed, 1 skipped, 98 deselected** (1 new test).
- `poetry run black --check .` — 185 files, all unchanged.
- `poetry run mkdocs build` — builds; the griffe warnings are pre-existing and all point at `payments_py/common/types.py`, none at the files touched here.
- `poetry run python scripts/lint_doc_links.py` — `docs/api/**` has no escaping or unmapped relative links.
- `./scripts/check_synced_doc_links.sh` — mintlify `broken-links` over the staged `api-reference/python/` tree: no broken links.
- Round-tripped the model directly: `billingModel` parses to `billing_model`, dumps back under the alias, and an unknown value (`"future-model"`) is carried through rather than raising.

Refs nevermined-io/nvm-monorepo#2999

https://claude.ai/code/session_01CVpimo1dBXnTJo9tQSy8SA


r-marques:
## Merged `main` after #276 landed — four test-file conflicts, resolved toward the factory

#276 (real `SettleResponse` / `VerifyResponse` doubles) merged, and it deleted exactly the hand-rolled bags this PR had been *adding* `billing_model` to. Four conflicts, all in test files, all the same shape.

**Resolved to main's side in every case, not this branch's.** #276's factory builds the real model, and `billing_model` is a field on it — which this PR itself adds to `types.py` — so a double carrying it needs no bag at all. Keeping this branch's `SettleResult` / `VerifyResult` classes would have reintroduced the drift #273 was opened against, in the same commit that merged its fix.

Resolved the conflict **hunks** rather than taking whole files with `--theirs`, so this PR's non-conflicting test changes survive.

One orphan fell out and is fixed: `test_payg_receipt_carries_billing_model_and_order_tx` referenced `_FakeSettlement`, this branch's bag. It now uses main's `_settlement()` helper, which returns a real `SettleResponse` — strictly stronger, since the double is now validated against the model rather than impersonating it.

## The guard survived the merge — measured, not assumed

A merge that silently weakens a test is the risk here, so I mutation-tested this PR's own guard afterwards rather than trusting a green suite:

| | |
| --- | --- |
| stop emitting `billingModel` on the receipt (`paywall.py`) | **kills** `test_payg_receipt_carries_billing_model_and_order_tx` |
| unmutated control | **1081 passed**, 1 skipped |

1081 rather than main's 1080 — this PR's test is the extra one, and it is still load-bearing.

`black --check` clean across 191 files.

tests/unit/mcp_tests/test_x402_inband.py
```diff
@@ -248,9 +248,48 @@ def handler(args, extra, ctx):
         assert nvm["creditsRedeemed"] == "5"
         assert nvm["remainingBalance"] == "100"
         assert nvm["planId"] == "plan-123"
+        # Absent on a credits settle that carried neither — omitted, not None,
+        # so a consumer can tell "absent" from "present and empty".
+        assert "billingModel" not in nvm and "orderTx" not in nvm
```
aaitor, line 253: **Should-fix — test coverage gap:** the two settlement-receipt tests in this class cover `billing_model` absent (this test, via `_settlement(success=True)` which never sets it) and `billing_model="pay-as-you-go"` (`test_payg_receipt_carries_billing_model_and_order_tx` below), but nothing here exercises the explicit `billing_model="credits"` value through the `nevermined/credits` receipt path. The PR's own doc example (`docs/api/09-mcp-integration.md`) shows `"billingModel": "credits"` in the sample response, but no test asserts that a settle with `billing_model="credits"` actually surfaces `billingModel: "credits"` in `nvm_meta` — only that it's omitted when absent. The `if credits_result.get("billingModel"):` check in `paywall.py` is value-agnostic so this is low functional risk, but it's the one discriminator value silently left untested by the PR that introduces the discriminator.

payments_py/mcp/core/paywall.py
```diff
@@ -275,6 +280,10 @@ async def wrapped(*all_args: Any) -> Any:
                 "subscriberAddress": auth_result.get("subscriber_address"),
                 "success": credits_result.get("success", False),
             }
+            if credits_result.get("billingModel"):
```
aaitor: **Nit:** `if credits_result.get("billingModel"):` (and the `orderTx` check two lines below) use truthiness, not `is not None`. The comment just above this block says the omit-when-absent logic lets "a consumer tell 'absent' from 'present and empty'" — but as written, a hypothetical empty-string value would be treated identically to `None`/absent and silently dropped, which is the opposite of that stated distinction. Low practical risk since `billing_model` is a non-empty enum-like string and `order_tx` is either `None` or a real charge reference in today's API, so this is unlikely to bite — but `is not None` would match the comment's actual intent and be robust if that ever changes.

payments_py/x402/types.py
```diff
@@ -252,28 +252,101 @@ class VerifyResponse(BaseModel):
...
+            when the settle had to order credits first (auto top-up). Treat it as
+            an opaque string.
     """
 
     success: bool
     error_reason: Optional[str] = Field(None, alias="errorReason")
     payer: Optional[str] = None
     transaction: str = ""
     network: str = ""
+    billing_model: Optional[str] = Field(None, alias="billingModel")
```
aaitor, line 349: **Praise:** good call typing `billing_model` as plain `str` rather than the new `X402BillingModel` Literal, consistent with how `network`/`scheme` are already handled in this module. Verified directly: `SettleResponse.model_validate({"billingModel": "future-model", ...})` parses cleanly and round-trips through `model_dump(by_alias=True)` without raising, so a future billing model this SDK release hasn't heard of won't turn a settle that actually charged the buyer into a `ValidationError` on the client. That's exactly the right tradeoff for a discriminator on a payment success path.

aaitor (commented):
Reviewed the diff plus full context of every touched file (types.py, facilitator.py, facilitator_api.py, paywall.py, agentcore/interceptor.py), confirmed the companion PR #276 (real SettleResponse/VerifyResponse model factories replacing attribute-bag test doubles) is already on `main` and is this branch's merge-base, cloned the branch and ran the actual test suite (`poetry run pytest -m "not slow"`: 1081 passed, 1 skipped, 102 deselected on the current main-merged branch; the x402/MCP-focused subset: 380 passed), and verified the parsing/round-trip claims directly against the Pydantic model (`billingModel` alias parses correctly, and an unrecognized value round-trips without raising since the field is typed `str`, not the `Literal`).

`billing_model` threads correctly end-to-end: `SettleResponse.model_validate(response.json())` in `facilitator_api.py` picks it up via `Field(None, alias="billingModel")`, it's `Optional[str]` (default `None`) so there's no backward-compat break for existing callers or test doubles, and every consumer (facilitator.py logging, the AgentCore interceptor logging, and paywall.py's two `_meta` receipts) reads it off the real object/dict rather than a hand-rolled double that could be missing the attribute — confirmed #276 already replaced the old attribute-bag doubles with real model factories (`tests/x402_responses.py`), so the "four test doubles" paragraph in the PR description refers to work that's now a no-op on this diff (verified: `git diff origin/main pr272` is empty for the three files that commit touched).

Two minor points left inline (one test-coverage gap worth closing, one low-severity nit on a truthiness check) plus one praise note. Nothing blocking.

aaitor (approved):
🟡 **Approved with comments** — no blockers; 1 should-fix, 1 nit(s) left inline.

Inline review: https://github.com/nevermined-io/payments-py/pull/272#pullrequestreview-5177496942

tests/unit/mcp_tests/test_x402_inband.py
r-marques: Folded in `2093640` — added `test_credits_receipt_carries_the_explicit_discriminator`.

Your reasoning is the part worth keeping, and it is the opposite of the intuitive one: *"Low functional risk since the emit check is value-agnostic"* is precisely why it needed a test rather than an argument. A value-agnostic check means nothing would have caught a regression that dropped **only** this value — the argument for safety is also the argument for no coverage.

You are also right about the shape of the gap: the class covered `billing_model` **absent** (read as `credits`) and `"pay-as-you-go"`, so the PR introducing the discriminator left untested the one value its own documentation example shows.

payments_py/mcp/core/paywall.py
r-marques: Fixed in `2093640`. Both that check and the `orderTx` one below it now use `is not None`.

You caught a comment and its code disagreeing: the comment directly above promises a consumer can tell *"absent"* from *"present and empty"*, and a truthiness test collapses exactly those two — an empty string dropped identically to a missing key. So the code did the opposite of what the line above it claimed.

I took the fix rather than relaxing the comment, because the distinction it states is the useful one and costs nothing to honour. The reasoning is recorded inline so it does not get "simplified" back: inert today (`billingModel` is enum-like and non-empty, `orderTx` is either `None` or a real charge reference), correct if that ever changes.

payments_py/x402/types.py
r-marques: Thank you — and thank you especially for **verifying** it rather than just agreeing with it. `SettleResponse.model_validate({"billingModel": "future-model", ...})` parsing cleanly and round-tripping is the claim that matters, and it is not obvious from the type alone.

The failure mode being avoided is worth stating for whoever reads this later: had `billing_model` been the `X402BillingModel` Literal, a billing model added server-side after this SDK release would raise `ValidationError` on the client — turning a settle that **actually charged the buyer** into a client-side exception. On a payment success path a permissive parse is the safe direction, and it matches how `network` and `scheme` are already handled in this module.

PR #276, opened 2026-09-10
test(x402): build real Settle/VerifyResponse models instead of attribute-bag doubles

### LINKED ISSUE #273: Every settle/verify test double has drifted from the model it claims to mimic (12 of 12)
Found while adding `billing_model` / `order_tx` to the settle response for nevermined-io/nvm-monorepo#2999 (PR #272). Adding two fields to `SettleResponse` turned **28 tests red** — not because the change was wrong, but because the test doubles standing in for that model did not have the fields the real one does.

The four that broke were fixed in that PR. This issue is about the other eight, and about the class.

## The sweep

Parsing every test class's `__init__` self-assignments and diffing against `SettleResponse.model_fields` / `VerifyResponse.model_fields`:

**12 settle/verify doubles exist. All 12 have drifted.**

| Kind | Count | Missing |
| --- | --- | --- |
| settle doubles | 5 | all 5 lack `error_reason`, `network`, `payer` (4 also lacked `billing_model` / `order_tx` before PR #272) |
| verify doubles | 7 | all 7 lack some combination of `invalid_reason`, `network`, `payer`, `url_matching`, `agent_request`, `agent_request_id` |

Worst case: `tests/integration/a2a/test_payments_server.py::SettleResult` is missing six of nine settle fields.

## Why this is worth fixing rather than patching case by case

The four repaired in #272 were **not "the drifted ones"** — they were the ones whose gap production code happened to reach for. The rest are latent by exactly the same mechanism: an `AttributeError` the first time any code path reads one of those fields. So the next person to add a field to either model gets the same red wall, and the one after that.

It also means these doubles are not testing what they appear to test. A double missing `network` cannot catch a regression in anything that reads `network`.

Note the tempting non-fix: making production code use `getattr(result, "field", None)`. That bends the SDK around its fixtures and hides exactly this drift — it was considered and rejected in #272.

## Suggested fix

Build a real `SettleResponse` / `VerifyResponse` in the tests instead of hand-rolling attribute bags. Pydantic then rejects an unknown field and supplies the defaults, so this class of drift becomes a parse error at the point of construction rather than an `AttributeError` somewhere downstream months later.

A shared factory helper (`make_settle_response(**overrides)`) would keep the call sites as short as the current bags.

## Related

- nevermined-io/payments-py#272 — where this surfaced; fixed 4 of the 12
- https://github.com/nevermined-io/payments/issues/433 — the sibling issue for the TypeScript SDK, same class, different shape
- nevermined-io/nvm-monorepo#2999 — the originating issue

r-marques:
Closes #273.

> ⚠️ **This body replaces an incorrect one.** The original was the TypeScript sibling's (nevermined-io/payments#437) — it described `pnpm build`, `tsc`, `tests/tsconfig.json` and a `typecheck:tests` CI step, none of which exist in this repo, and its `Closes #433` is a 404 here. Caught in review. The correct issue is #273; see the note at the bottom on how it happened.

## The defect

Every test that stubs `facilitator.verify_permissions` / `settle_permissions` hand-rolled an attribute bag:

```python
class SettleResult:
    def __init__(self, success=True, transaction="0x123"):
        self.success = success
        self.transaction = transaction
```

Those bags carry only the fields the production code happened to read on the day they were written. All **12** in the suite had drifted from `SettleResponse` / `VerifyResponse`. The failure mode is asymmetric and nasty: the first code path to read a field a bag never had blows up with `AttributeError`, and adding a field to the real model turns a wall of unrelated tests red for a reason that has nothing to do with the change.

## The fix

Build the **real Pydantic model** instead, via `tests/x402_responses.py`. Every field the model declares exists on the double, populated by the model's own defaults, so a new field is inert in tests that do not care about it.

**The model alone cannot supply the other half of the guarantee.** Both response models use Pydantic's default `extra="ignore"`, so a typo'd or removed field is *silently dropped* rather than raised:

```python
SettleResponse(success=True, creditsRedeemd='9').credits_redeemed   # -> None, no raise
```

So `_reject_unknown` validates override names against `model_fields` — **and runs over the defaults dicts at import time**, which is the hole that would otherwise let the factory reintroduce the very bug it exists to fix. A bad default is then a collection error across every importing file, not one quiet degradation.

It also rejects the camelCase wire aliases, deliberately: `populate_by_name=True` means Pydantic *accepts* `creditsRedeemed`, and `{**defaults, **overrides}` keys on the string — so two spellings of one field survive as two keys and Pydantic prefers the alias **whatever the dict order**, silently dropping the override. Verified in both orders on pydantic 2.12.5.

An AST sweep (`tests/unit/test_x402_response_doubles.py`) fails if a hand-rolled bag reappears.

## Verification

Runner: **pytest via the payments-py poetry venv** (not testmcp).

- `poetry run pytest -m "not slow"` — **1080 passed**, 1 skipped, 102 deselected
- `poetry run black --check .` — 191 files unchanged
- `poetry run mkdocs build` — clean

### Mutation results — which tests die

| mutant | killed |
| --- | --- |
| sweep detector → `return set()` | `test_sweep_detects_a_hand_rolled_double[settle]`, `[verify]` |
| typo a key in `SETTLE_DEFAULTS` | **12 collection errors** (import-time) |
| `TESTS_ROOT` → nonexistent path | `test_no_hand_rolled_…` (`assert 0 > 50`) |
| strip the name check from `_reject_unknown` | `test_unknown_override_is_rejected[settle]`, `[verify]`, `test_alias_spelling_is_rejected`, `test_defaults_are_validated_too[SETTLE_DEFAULTS]`, `[VERIFY_DEFAULTS]` |
| unmutated control | 1080 passed |

Mutating `DEFAULT_NETWORK`, `DEFAULT_PAYER`, `transaction`, `credits_redeemed` and `remaining_balance` each leaves 1080 passing — no assertion is pinned to a factory value. `success: True` **is** load-bearing (16 red), correctly: it is the "this is the success shape" default that bare `make_settle_response()` call sites depend on.

## Known blind spot, stated rather than implied

The sweep parses `ast.ClassDef` only, so dict / `Mock` / `SimpleNamespace` stubs are invisible to it. The tree is clean of those now, and `CLAUDE.md` says so explicitly rather than implying coverage it lacks.

## How the wrong body happened

Several agents in one session shared a scratchpad directory, and a PR-body file was overwritten by another agent working in a different repo. That near-miss was recorded earlier in the same session on another PR and caught before pushing; here it was not. The lesson is namespacing per agent, not "check more carefully" — `$CLAUDE_JOB_DIR/tmp` is shared by every agent in a session.

https://claude.ai/code/session_01CVpimo1dBXnTJo9tQSy8SA

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -106,6 +106,32 @@ poetry run pytest tests/unit/test_example.py -v
 poetry run pytest --cov=payments_py --cov-report=term-missing
 ```
 
+### Never hand-roll a settle/verify test double
+
+Tests that stub `facilitator.verify_permissions` / `settle_permissions` must
+build the **real** model via `tests/x402_responses.py`, never an attribute bag:
+
+```python
+from tests.x402_responses import make_settle_response, make_verify_response
+
+facilitator.settle_permissions = lambda **k: make_settle_response(transaction="0xabc")
+```
+
+A hand-rolled class only ever carries the fields production code read the day it
+was written. All 12 such doubles in this suite had drifted from the model
+(#273), and adding two fields to `SettleResponse` turned **28 unrelated tests
+red** with `AttributeError` (#272). Building the model means every field exists,
+populated by the model's own defaults, so a new field is inert in tests that do
+not care about it — measured: the same probe produces 28 failures against the
+old doubles and 0 against these.
+
+Do **not** "fix" that by making production code use `getattr(result, "f", None)`
+— that bends the SDK around its fixtures and hides the drift. Note the factories
+reject an unknown or aliased keyword themselves: both models use Pydantic's
+default `extra="ignore"`, so constructing them directly would silently drop a
+typo'd or since-removed field name. `tests/unit/test_x402_response_doubles.py`
+fails if a hand-rolled bag reappears.
+
 ### Test Markers
 
 - No marker: Unit and integration tests (fast)
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -126,11 +126,26 @@ not care about it — measured: the same probe produces 28 failures against the
 old doubles and 0 against these.
 
 Do **not** "fix" that by making production code use `getattr(result, "f", None)`
-— that bends the SDK around its fixtures and hides the drift. Note the factories
-reject an unknown or aliased keyword themselves: both models use Pydantic's
-default `extra="ignore"`, so constructing them directly would silently drop a
-typo'd or since-removed field name. `tests/unit/test_x402_response_doubles.py`
-fails if a hand-rolled bag reappears.
+— that bends the SDK around its fixtures and hides the drift.
+
+The factories validate keyword **names** themselves, for both the overrides and
+their own defaults, and the check is **not** redundant with Pydantic:
+
+- **Unknown name** — both models use Pydantic's default `extra="ignore"`, so
+  constructing them directly *silently drops* a typo'd or since-removed field.
+  The model does no enforcing; `_reject_unknown` does.
+- **camelCase alias** — this one is **not** dropped. `populate_by_name=True`
+  means `SettleResponse(creditsRedeemed="5")` is accepted and sets
+  `credits_redeemed`. It is rejected because of the defaults/overrides dict
+  merge: two spellings of one field survive as two keys, and Pydantic resolves
+  the alias in preference to the field name *whatever the dict order* — so a
+  mixed-spelling merge silently discards one of them, in the direction where the
+  override is the loser.
+
+`tests/unit/test_x402_response_doubles.py` fails if a hand-rolled bag reappears,
+and carries its own positive control so it cannot silently stop detecting. Its
+sweep parses `ast.ClassDef` only, so `dict` / `Mock` / `SimpleNamespace` stubs
+are **outside** what it can see — those are on review to catch.
 
 ### Test Markers
 
--- a/tests/unit/test_x402_response_doubles.py
+++ b/tests/unit/test_x402_response_doubles.py
@@ -1,0 +1,120 @@
+"""Guards for the x402 response factories, and against the drift they replaced.
+
+payments-py#273: every one of the suite's 12 settle/verify test doubles was a
+hand-rolled attribute bag carrying only the fields the production code happened
+to read the day it was written. Adding two fields to ``SettleResponse`` turned
+28 unrelated tests red with ``AttributeError``, and a double missing ``network``
+could never have caught a regression in anything reading ``network``.
+
+The fix was to build the real Pydantic models (``tests/x402_responses.py``).
+These tests pin the two properties that keeps it fixed.
+"""
+
+import ast
+import pathlib
+
+import pytest
+
+from payments_py.x402.types import SettleResponse, VerifyResponse
+from tests.x402_responses import make_settle_response, make_verify_response
+
+TESTS_ROOT = pathlib.Path(__file__).resolve().parents[1]
+
+
+class TestFactories:
+    def test_settle_factory_carries_every_model_field(self):
+        response = make_settle_response()
+        assert isinstance(response, SettleResponse)
+        for field in SettleResponse.model_fields:
+            # Not a tautology about Pydantic: the point is that a *new* field on
+            # the model is present on every double the day it is added, rather
+            # than raising AttributeError in whichever test first reads it.
+            assert hasattr(response, field)
+
+    def test_verify_factory_carries_every_model_field(self):
+        response = make_verify_response()
+        assert isinstance(response, VerifyResponse)
+        for field in VerifyResponse.model_fields:
+            assert hasattr(response, field)
+
+    def test_overrides_win_over_defaults(self):
+        assert make_settle_response(transaction="0xfeed").transaction == "0xfeed"
+        assert make_verify_response(is_valid=False).is_valid is False
+
+    @pytest.mark.parametrize(
+        "factory",
+        [make_settle_response, make_verify_response],
+        ids=["settle", "verify"],
+    )
+    def test_unknown_override_is_rejected(self, factory):
+        # Both models use Pydantic's default extra="ignore", so constructing them
+        # directly would silently DROP a typo'd or since-removed field name. The
+        # factory has to supply that half of the guarantee itself.
+        with pytest.raises(TypeError, match="has no field"):
+            factory(no_such_field="x")
+
+    def test_alias_spelling_is_rejected(self):
+        # Call sites use the Python names; the camelCase wire alias would be
+        # silently ignored by the model, so reject it loudly.
+        with pytest.raises(TypeError, match="has no field"):
+            make_settle_response(creditsRedeemed="5")
+
+
+def _self_assigned_attrs(node: ast.ClassDef) -> set[str]:
+    """Attribute names a class sets on instances (class body + ``__init__``)."""
+    attrs: set[str] = set()
+    for stmt in node.body:
+        if isinstance(stmt, ast.AnnAssign) and isinstance(stmt.target, ast.Name):
+            attrs.add(stmt.target.id)
+        elif isinstance(stmt, ast.Assign):
+            attrs.update(t.id for t in stmt.targets if isinstance(t, ast.Name))
+        elif isinstance(stmt, ast.FunctionDef) and stmt.name == "__init__":
+            for sub in ast.walk(stmt):
+                targets = (
+                    sub.targets
+                    if isinstance(sub, ast.Assign)
+                    else [sub.target] if isinstance(sub, ast.AnnAssign) else []
+                )
+                attrs.update(
+                    t.attr
+                    for t in targets
+                    if isinstance(t, ast.Attribute)
+                    and isinstance(t.value, ast.Name)
+                    and t.value.id == "self"
+                )
+    return attrs
+
+
+def test_no_hand_rolled_settle_or_verify_doubles():
+    """No test class may stand in for a settle/verify response by hand.
+
+    This is the regression guard for the whole class of defect, not for the 12
+    instances that existed when it was written: a new attribute bag is caught on
+    the PR that adds it rather than by the next person to touch either model.
+
+    If this fails, use ``make_settle_response`` / ``make_verify_response`` from
+    ``tests/x402_responses.py`` instead of declaring a class.
+    """
+    settle_fields = set(SettleResponse.model_fields)
+    offenders = []
+
+    for path in sorted(TESTS_ROOT.rglob("*.py")):
+        if "__pycache__" in path.parts:
+            continue
+        for node in ast.walk(ast.parse(path.read_text())):
+            if not isinstance(node, ast.ClassDef):
+                continue
+            attrs = _self_assigned_attrs(node)
+            # "success" alone is far too common to key on; require it to look
+            # like a settle response. "is_valid" is specific enough on its own.
+            if ("success" in attrs and len(attrs & settle_fields) >= 2) or (
+                "is_valid" in attrs
+            ):
+                rel = path.relative_to(TESTS_ROOT.parent)
+                offenders.append(f"{rel}:{node.lineno} {node.name} {sorted(attrs)}")
+
+    assert not offenders, (
+        "Hand-rolled settle/verify response double(s) found; build the real "
+        "model with tests.x402_responses instead (payments-py#273):\n  "
+        + "\n  ".join(offenders)
+    )
--- a/tests/unit/test_x402_response_doubles.py
+++ b/tests/unit/test_x402_response_doubles.py
@@ -11,6 +11,7 @@ These tests pin the two properties that keeps it fixed.
 """
 
 import ast
+import importlib
 import pathlib
 
 import pytest
@@ -54,10 +55,25 @@ class TestFactories:
             factory(no_such_field="x")
 
     def test_alias_spelling_is_rejected(self):
-        # Call sites use the Python names; the camelCase wire alias would be
-        # silently ignored by the model, so reject it loudly.
+        # NOT because the model would ignore it — populate_by_name=True means
+        # SettleResponse(creditsRedeemed="5") is accepted and sets
+        # credits_redeemed. The reason is the defaults/overrides dict merge:
+        # two spellings of one field survive as two keys and Pydantic resolves
+        # the alias in preference to the field name whatever the dict order, so
+        # a mixed-spelling merge silently drops one of them.
         with pytest.raises(TypeError, match="has no field"):
             make_settle_response(creditsRedeemed="5")
+
+    @pytest.mark.parametrize("defaults_name", ["SETTLE_DEFAULTS", "VERIFY_DEFAULTS"])
+    def test_defaults_are_validated_too(self, defaults_name):
+        # The overrides check alone left the factory carrying this PR's own
+        # defect: a typo'd default key is dropped by extra="ignore", every
+        # double degrades to the model's default, and the suite stays green.
+        module = importlib.import_module("tests.x402_responses")
+        model = SettleResponse if defaults_name == "SETTLE_DEFAULTS" else VerifyResponse
+        assert not set(getattr(module, defaults_name)) - set(model.model_fields)
+        with pytest.raises(TypeError, match="has no field"):
+            module._reject_unknown(model, {"remainingBalanceTYPO": 1}, "in defaults")
 
 
 def _self_assigned_attrs(node: ast.ClassDef) -> set[str]:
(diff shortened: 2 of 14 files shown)


r-marques:
## Self-review before flipping ready

Verified at head `382ee96e`. The interesting question on this PR is not "do the tests pass" — it is whether the new factory can itself drift, since the two response models use Pydantic's default `extra="ignore"` and would **silently swallow** a bad field name. I measured that directly rather than reading the code for it.

```
✅ bogus field  -> TypeError: SettleResponse has no field(s) ['totallyNotAField'] (passed as an override)
✅ raw model silently ignored the same field   (hasattr -> False)
✅ camelCase alias rejected by the factory
✅ ...but pydantic itself DOES accept it       (SettleResponse(creditsRedeemed='5').credits_redeemed == '5')
```

Lines 2 and 4 are the load-bearing ones. Line 2 proves `extra="ignore"` is real, so `_reject_unknown` is doing genuine work rather than duplicating a check Pydantic already performs. Line 4 proves the alias rejection is a deliberate addition and not redundant: `populate_by_name=True` means Pydantic would happily accept `creditsRedeemed`, so without the factory's check a mixed-spelling merge could silently drop an override — and, as the docstring works out, it drops it in the direction that matters.

**The defaults are validated at import time too** (`_reject_unknown(SettleResponse, SETTLE_DEFAULTS, …)`), which closes the hole where the factory reintroduces the very defect it exists to fix: a typo'd default key would be dropped by `extra="ignore"`, every double would silently degrade to the model's own default, and the suite would stay green while testing less than it claims.

## Suite

`poetry run pytest -m "not slow" -q` — **1080 passed**, 1 skipped, 102 deselected, 3.69s. Runner: pytest via poetry.

## What I did not check

I did not re-derive whether each of the 12 converted doubles asserts the same things it did before conversion; I checked the mechanism and the suite result. If a reviewer wants one spot-checked, `test_paywall_invalid_token_flow.py` is the one with the most behavioural surface.

aaitor (commented):
Reviewed: CI green (lint/unit_integration/e2e/deepagents_compat). Diff is test-only, adds `tests/x402_responses.py` factories for `SettleResponse`/`VerifyResponse`. The import-time `_reject_unknown` validation against `model_fields` correctly closes the `extra="ignore"` drift hole this PR exists to fix, and the camelCase-alias rejection in `_reject_unknown`'s docstring reasoning checks out (alias resolution in `{**defaults, **overrides}` would otherwise silently drop the override). No blockers found.

Note: this session's PR-review bot tooling has a persistent "sensitive file" permission gate on its run-dir artifact path, so it cannot complete the deterministic verdict step (act.py) right now — leaving this for a human approve or a later tick.

eruizgar91:
👀 Reviewing this now.

CLAUDE.md
```diff
@@ -106,6 +106,47 @@ poetry run pytest tests/unit/test_example.py -v
 poetry run pytest --cov=payments_py --cov-report=term-missing
 ```
 
+### Never hand-roll a settle/verify test double
```
eruizgar91, line 109: **MEDIUM** — The PR description is the TypeScript sibling's, not this one's, and it links the wrong issue.

The body describes `pnpm build` / `tsc`, a root `tsconfig.json` `exclude`, ts-jest `isolatedModules`, `tests/tsconfig.json`, a new `pnpm typecheck:tests` script and a step added to `testing.yml`'s `lint_build` job — none of which exist in `payments-py`, and none of which are in this diff. This repo's gate is `poetry run pytest -m "not slow"` in `.github/workflows/test.yaml`, and nothing in this PR adds a typecheck gate. Its whole verification table (`pnpm test:unit` 635 passed, `pnpm lint`, the M1/M2/M3 `tsc` mutations) is measured against a different repository.

That matters beyond tidiness in two concrete ways:

1. **`Closes #433` names an issue that does not exist here** (`GET /repos/nevermined-io/payments-py/issues/433` → 404), so nothing is auto-closed — while [#273](https://github.com/nevermined-io/payments-py/issues/273), the issue this PR actually resolves and which `tests/x402_responses.py` and this CLAUDE.md section both cite by number, is left unlinked and will stay open after merge.
2. A maintainer reading the body comes away believing the test tree is now type-gated in CI. It is not, and the real guard shipped here (`tests/unit/test_x402_response_doubles.py`) is never mentioned.

The actual work is well described by this CLAUDE.md section — replacing the description with the Python story, and swapping `Closes #433` for `Closes #273`, is the whole fix.

eruizgar91 (commented):
Test-only PR replacing 12 hand-rolled settle/verify attribute bags with real `SettleResponse` / `VerifyResponse` models behind `tests/x402_responses.py`, plus an AST sweep that fails if a bag reappears. No production code changes. Findings are inline: **0 CRITICAL, 0 HIGH, 1 MEDIUM, 2 LOW** — nothing in the payment or settlement paths.

### The guards were verified mechanically, not taken on trust

Every claim the module makes about itself was reproduced by constructing the failure it says it prevents (Python 3.14, pydantic 2.12.5, worktree at `382ee96`):

| probe | result |
|---|---|
| Baseline defect is real | `SettleResponse(success=True, creditsRedeemd='9').credits_redeemed` → `None`, no raise. `model_config` carries no `extra`, so pydantic's `ignore` applies. |
| Typo'd override | `make_settle_response(credits_redeemd='9')` → `TypeError: SettleResponse has no field(s) ['credits_redeemd']` ✅ |
| camelCase alias override | `make_settle_response(creditsRedeemed='5')` → `TypeError` ✅ |
| Removed-name override | `make_verify_response(isValid=False)` → `TypeError` ✅ |
| **Bad default (the import-time check)** | `SETTLE_DEFAULTS['credits_redeemed']` → `'creditsRedeemed'`: `import tests.x402_responses` raises `TypeError … (in SETTLE_DEFAULTS)`, and pytest reports it as a **collection error** across every importing file — loud, not silent ✅ |
| Is the import-time check load-bearing or duplicated? | Removing both `_reject_unknown(…DEFAULTS…)` lines **and** typo'ing a default still fails `test_defaults_are_validated_too[SETTLE_DEFAULTS]`. The property is double-covered; the import-time line makes it a hard error rather than one red test. |
| Sweep positive control | Neutering `_self_assigned_attrs` to `return set()` turns `test_sweep_detects_a_hand_rolled_double[settle|verify]` red ✅ — the detector cannot silently stop detecting. |

The docstring's non-obvious claim — that pydantic resolves the **alias in preference to the field name regardless of dict order** — is also true as written: `SettleResponse(success=True, **{'credits_redeemed': A, 'creditsRedeemed': B})` yields `B` in **both** key orders. So rejecting the alias spelling really is required for "overrides win" to hold, and is not defensive noise.

### The PR's premise was checked independently

Added `probe_field: Optional[str]` to `SettleResponse` and a read of it in `paywall.py` (`settle_result.probe_field`), then ran the same mcp subset on `origin/main` and on this head, subtracting each side's own baseline (24 failures either side, an artefact of running the subset in isolation under 3.14):

- `origin/main` (hand-rolled bags): 24 → **33** = **+9** new `AttributeError`-class failures
- this head (real models): 24 → **24** = **+0**

That is the claimed property, reproduced. Trees restored; both worktrees clean.

### Checked and deliberately discarded

- **Coverage loss on `_coerce_settle_response`'s duck-typed branch.** Converting `test_payments_stream_and_push` / `test_push_notifications` from dicts to real models moves them onto the `isinstance(SettleResponse)` branch. The duck-typed branch is still directly covered by `tests/unit/a2a/test_inband.py::test_coerce_settle_response_from_duck_typed_object`, which the PR correctly leaves as a `SimpleNamespace`. Not a regression.
- **`txHash is None` → `txHash == ""`** in `test_mcp.py`. Verified faithful, not a weakened assertion: `paywall.py` emits `settle_result.transaction if settle_success else None`, and `SettleResponse.transaction` is a non-Optional `str` defaulting to `""`. The old assertion pinned a state production cannot produce.
- **The `exclude_none` assertion in `test_x402_inband.py`.** Still load-bearing — `order_tx` / `error_reason` are genuinely `None` on the real model and get dropped, so dropping `exclude_none` in production would still surface `orderTx`.
- **`credits_redeemed or str(credits)` fallback in `paywall.py`** is unexercised by the new defaults — but it was equally unexercised by the old `SettleResult(credits_redeemed="1")` default. No change.
- **Files constructing `SettleResponse(...)` / `VerifyResponse(...)` directly** (`test_agentcore_decorator`, `test_langchain_decorator`, `test_deepagents_compat`, `test_strands_decorator`) bypass `_reject_unknown` and so remain exposed to the silent-drop typo. They are real models, so the `AttributeError` class this PR targets is already closed for them; routing them through the factory is a separate sweep, not this PR's.
- **CI actually runs the guard**: `test.yaml` runs `poetry run pytest -m "not slow"` against `testpaths = ["tests"]`, and `python_files = ["test_*.py"]` matches the new file. All four checks are green at this SHA.
- **Local suite**: 1007 passed on this head; the 5 `tests/unit/langsmith/test_spans.py` failures reproduce identically on `origin/main` (missing optional extra in my venv), so they are pre-existing and unrelated.

A prior `COMMENTED` review on this PR reported no blockers; this review was done independently from the diff and the runtime probes above, and reaches the same conclusion on the load-bearing question while adding three non-blocking items.

tests/x402_responses.py
```diff
@@ -0,0 +1,127 @@
...
+
+#: CAIP-2 network the rest of the suite's fixture tokens are minted against.
+DEFAULT_NETWORK = "eip155:84532"
+#: Subscriber address the fixture tokens are issued to.
+DEFAULT_PAYER = "0xTestSubscriber"
+
+#: A settle response shaped the way the facilitator really answers a successful
+#: burn: every field the backend populates is populated here too, so a test that
+#: reads one gets a value rather than an attribute error.
+SETTLE_DEFAULTS: Dict[str, Any] = {
```
eruizgar91, line 54: **LOW** — `make_settle_response(success=False)` inherits the *success* shape, producing a response the facilitator never sends.

`SETTLE_DEFAULTS` is documented (and correctly built) as "the way the facilitator really answers a successful burn". But `success` is just another overridable key, so `make_settle_response(success=False)` returns `transaction="0x123"`, `credits_redeemed="1"`, `remaining_balance="100"` and `error_reason=None` — contradicting `SettleResponse`'s own docstring (`transaction`: *"empty string if settlement failed"*; `error_reason`: *"only present if success is false"*).

No current call site hits it — but only because both places that needed a failure shape worked around the factory by hand:

- `tests/unit/mcp_tests/test_x402_inband.py`'s `_settlement()` re-derives `transaction` from `success` and adds `error_reason` itself;
- `tests/integration/mcp_tests/test_paywall_invalid_token_flow.py:60` has to pass `payer=None` explicitly, with a comment saying exactly this — *"VERIFY_DEFAULTS is the successful shape, so clear it rather than inherit it"*.

Two call sites independently compensating is the signal. The next one to write a bare `make_settle_response(success=False)` silently encodes an impossible response — which is the fixture-fidelity failure this PR exists to remove, one level up.

Bounded fix inside `_build`/`make_settle_response`: when `success` is overridden to `False` and the caller did not say otherwise, default `transaction` to `""` and require/default an `error_reason` — or add an explicit `make_failed_settle_response()` alongside the two factories.

tests/unit/test_x402_response_doubles.py
```diff
@@ -0,0 +1,178 @@
...
+                    and t.value.id == "self"
+                )
+    return attrs
+
+
+def _looks_like_a_double(attrs: set[str]) -> bool:
+    # "success" alone is far too common to key on; require it to look like a
+    # settle response. "is_valid" is specific enough on its own.
+    settle_fields = set(SettleResponse.model_fields)
+    return ("success" in attrs and len(attrs & settle_fields) >= 2) or (
```
eruizgar91, line 120: **LOW** — A class-shaped double carrying only `success` is not detected, and that blind spot is undocumented.

`success` is itself in `SettleResponse.model_fields`, so `len(attrs & settle_fields) >= 2` means *`success` plus at least one other model field*. A bag with `success` alone slips through — asymmetrically with `is_valid`, which is caught on its own.

Reproduced against this file's own `_sweep`:

```
MISSED    class SettleStub:
              def __init__(self): self.success = True
DETECTED  class SettleStub:
              def __init__(self): self.success = True; self.transaction = '0x1'
MISSED    class SettleStub:
              success: bool = True
```

(the third miss has the same cause, not an AST gap — `_self_assigned_attrs` picks the `AnnAssign` up fine, the `>= 2` threshold then drops it.)

That exact shape is not hypothetical: the TypeScript sibling's own list of drifted doubles includes one that was *"`{ success: true }` and nothing else"*. So the docstring on `test_no_hand_rolled_settle_or_verify_doubles` — *"No test class may stand in for a settle/verify response by hand"* — overstates what the sweep enforces.

The `>= 2` threshold is a defensible trade: `success` alone is common enough that keying on it would false-positive across unrelated result stubs, and a false positive here fails an innocent PR. So the cheap fix is honesty rather than sensitivity — the CLAUDE.md section already enumerates the blind spots (`dict` / `Mock` / `SimpleNamespace` stubs are *"on review to catch"*); add this one to that same list, and soften the docstring to "...carrying two or more of the model's fields".

eruizgar91 (approved):
🟡 **Approved with comments** — no blockers; 1 should-fix, 2 nit(s) left inline.

Inline review: https://github.com/nevermined-io/payments-py/pull/276#pullrequestreview-5177097892

CLAUDE.md
r-marques: You are right, and this is the worst of the three because it is the one a reader would have believed. **The body was payments#437's — a different repository's PR.** Replaced.

Everything you listed checks out: `pnpm build`, `tsc`, `tests/tsconfig.json`, `isolatedModules`, the `typecheck:tests` script, the `testing.yml` `lint_build` step — none of them exist here, and the entire verification table (635 jest tests, the M1/M2/M3 `tsc` mutations) was measured against the TypeScript repo. And `Closes #433` really is a 404 in `payments-py`, so **nothing** was going to auto-close while #273, the issue this actually resolves, sat unlinked.

The new body is this repo's: `Closes #273`, pytest/black/mkdocs as the gates, the real numbers (1080 passed, 191 files black-clean), and the mutation table that was actually run here — including the import-time `_reject_unknown` probe and the `scanned > 50` floor.

**How it happened, since a wrong body is a process failure rather than a typo:** several agents in one session shared a scratchpad directory and a PR-body file was overwritten by an agent working in another repo. The same near-miss was caught on a different PR earlier in this session; here it was not. The fix is namespacing the scratchpad per agent, not resolving to check more carefully — `$CLAUDE_JOB_DIR/tmp` is shared by every agent in a session. I have written that note into the body itself rather than quietly swapping it, so the record shows what went wrong.

Your point that it matters *beyond tidiness* is the right emphasis. A body describing a typecheck gate this PR does not add would have sent the next reader looking for one.

tests/x402_responses.py
r-marques: Folded in `83b4d38`. Reproduced it first — `make_settle_response(success=False)` really did return `transaction='0x123'`, `credits_redeemed='1'`, `remaining_balance='100'`, which contradicts `SettleResponse`'s own docstring on both fields.

`SETTLE_FAILURE_DEFAULTS` is now selected when `success=False`, with explicit overrides still winning over either base:

```
success=False         -> transaction='' credits_redeemed='0' remaining_balance='0'
success=False + error -> error_reason='insufficient credits', transaction=''
default               -> success=True transaction='0x123' credits_redeemed='1'
```

**The part I would have missed without your own earlier finding:** a second defaults dict reintroduces exactly the hole the import-time `_reject_unknown` closes. So `SETTLE_FAILURE_DEFAULTS` is validated at import too, and I mutation-tested that specifically rather than assuming it inherited the property — typo a key in it and collection fails with **12 errors**, not one quiet degradation.

Your read of the evidence is what made this worth doing rather than theoretical: both call sites that needed a failure shape had already worked around the factory by hand, and one carried a comment saying so. A helper that every caller has to correct is not a helper.

tests/unit/test_x402_response_doubles.py
r-marques: Documented rather than closed, in `83b4d38` — and I want to be explicit that this is a deliberate reading of your finding, not a narrowing of it.

You wrote that the blind spot is *undetected and undocumented*. Closing it is the wrong fix: `success` is itself a `SettleResponse` field, so keying on it alone — which is what symmetry with `is_valid` would require — flags every unrelated class in the tree that happens to carry a `success` attribute. The predicate's existing comment already says that, which is why I read the finding as being about the silence rather than the gap.

So the note now states it plainly: the `>= 2` means *success plus at least one OTHER settle field*; a `success`-only class is missed; that is asymmetric with `is_valid`; it is the price of not generating false positives; and **widening the predicate is not the fix — measure the false positives first**. It also points at the module docstring's other blind spots so they read as one list.

One observation in support of leaving it: such a bag is close to unusable anyway. Production reads `.transaction` / `.credits_redeemed` off a settle result, so a double carrying only `success` AttributeErrors on first access — it fails loudly at the point of use rather than drifting silently, which is the opposite of the failure mode this sweep exists to catch.

If you would still rather it were detected, say so and I will measure the false-positive count across the tree before changing the predicate.

r-marques: Resolving this one, and saying why rather than doing it silently.

Your finding was that the blind spot is *undetected and **undocumented***. It is now documented, in the predicate itself, with the reason it is not closed and an explicit instruction not to widen it without measuring first. That is the deliverable, so the thread is addressed.

My offer to go further — measure the false-positive count and detect it — stands and is not withdrawn by resolving. If you want it, say so here and I will reopen and do it; this is not me closing past you.

I am resolving rather than holding because `ship.py` refuses to merge while any reviewer thread is open, and leaving the PR blocked on a courtesy question I volunteered would be the wrong reason for it to sit.

Issue #280, opened 2026-09-14
ci: a failing post-merge E2E on main reports the run as green (continue-on-error applies to push runs)


eruizgar91:
## Summary

`test.yaml`'s `e2e` job carries `continue-on-error: true`, which applies to the post-merge `push` run on `main` as well as to pull requests. A failing post-merge E2E therefore leaves the workflow run concluding **`success`**: green badge, no notification, nobody told.

Measured on this repo: run [`34108722286`](https://github.com/nevermined-io/payments-py/actions/runs/34108722286) (push to `main`, 2026-09-07) has job `e2e: failure` inside a run that concluded `success`.

## Why it matters now

#278 makes `e2e` skip on Dependabot-triggered runs, because GitHub does not expose repository secrets to `dependabot[bot]` and the job cannot pass there. That is the right call for the required check, but it means the full chain on a dependency bump is:

1. `e2e: skipped` satisfies the required check on the PR
2. the bump merges with E2E never executed
3. the post-merge E2E on `main` fails
4. the run still reports green

So the one place the money path would be exercised after a bump is also the one place a failure is silent. #278's a2a-sdk bump is the concrete example of the class: a dependency changed a return arity on the credit-burn path.

## Options

1. **Scope the escape hatch to PRs** — one line:

   ```yaml
   continue-on-error: ${{ github.event_name == 'pull_request' }}
   ```

   CLAUDE.md notes that "E2E test failures due to pending backend deployments are expected" and asks for coordination with the team, which argues for keeping the hatch on the PR side and letting a `main` failure be loud. This does mean `main` goes red while staging lags behind a backend deploy — a team call, which is why it is filed rather than folded into #278.

2. **Populate the Dependabot secret store** with `TEST_SUBSCRIBER_API_KEY` / `TEST_BUILDER_API_KEY`. E2E then runs on Dependabot PRs, the `github.actor` clause added in #278 can be dropped, and the signal moves back before the merge instead of after it. This is the better end state and is already named in the workflow comment.

The two are complementary; 2 is the one that actually restores pre-merge coverage.

## Related

- #278 — added the Dependabot gate and the comment pointing here.
- #279 — separate: non-blocking / `auth_required` requests cancel the producer before the burn runs.

r-marques:
[...]

PR #278, opened 2026-09-14
chore(deps): refresh dependencies and unblock the Dependabot queue

eruizgar91:
## Problem

payments-py has **18 open Dependabot PRs** and **87 open security alerts**. Two independent problems kept them there.

### 1. The `e2e` required check can never pass on a Dependabot PR

GitHub does not expose repository secrets to `dependabot[bot]` — they would have to live in the separate Dependabot secret store, which this repo does not populate. So `TEST_SUBSCRIBER_API_KEY` / `TEST_BUILDER_API_KEY` arrive empty and every E2E test errors identically:

```
TEST_SUBSCRIBER_API_KEY:
...
>           [_, key] = self.nvm_api_key.split(":")
E           ValueError: not enough values to unpack (expected 2, got 1)
payments_py/api/base_payments.py:193: ValueError
```

(run [`31309887118`](https://github.com/nevermined-io/payments-py/actions/runs/31309887118), PR #261). `e2e` is a required status check, so every Dependabot PR is `BLOCKED` on a failure it cannot fix. `main` itself is green.

The job already skipped **fork** PRs for exactly this reason; it now also skips **Dependabot-triggered** runs. A job skipped by a job-level `if` reports conclusion `skipped`, which satisfies the required check — the fork gate came in with #206, and #215 / #181 are the merged PRs that show `e2e: conclusion=skipped` against protected `main`. (#269, cited in the first revision, is the wrong precedent — its own `e2e` ran and passed.)

The gate reads `github.actor` (the actor of the triggering event), **not** the PR author, so when a human pushes to a `dependabot/*` branch to fix up a bump, secrets are available and E2E runs as normal. That was the failure mode called out on the TypeScript side in nevermined-io/payments#449.

### 2. `.github/dependabot.yml` configured the wrong ecosystem

It declared `npm` in a Poetry repository with **no package.json**, unchanged since the first commit. It matched nothing, so no version updates were ever configured — every PR this repo has received came from *security* updates, which run off the dependency graph and ignore this file, producing one ungrouped PR per advisory.

Replaced with grouped `pip` (Poetry: pyproject.toml + poetry.lock) and `github-actions` entries. Security updates get their own group because `applies-to` defaults to version-updates.

## Dependencies refreshed

All 87 alerts clear against the new lock (checked mechanically: every alert's `first_patched_version` against the locked version). 91 packages moved inside existing constraints; these needed the constraint raised:

| Package | Was | Now | Why |
|---|---|---|---|
[...]

PR #283, opened 2026-09-14
ci: auto-approve and queue routine Dependabot bumps

eruizgar91:
## Why

#278 made Dependabot PRs *able* to pass CI. They still needed a human approval each, because `main` requires one approving review and this repo has no automation for it. #281 is the shape of what was left: every check green, then waiting for someone to click approve.

This ports the pattern already running in nevermined-io/payments and nevermined-io/nvm-monorepo — adapted to this repo, not copied.

## What differs from the two existing implementations, and why

**It can actually work here.** `main` requires 1 approving review with `require_code_owner_reviews: false`, so the release bot's approval satisfies branch protection. That is precisely what does *not* hold in the TypeScript repo, where CODEOWNERS is required with no `bypass_pull_request_allowances` and a GitHub App cannot be a team member (payments#449). The workflow header says so, so nobody copies it back and wonders why it stopped working.

**No `dependabot-update-branches.yml`.** The TypeScript repo needs one because its `main` sets `strict: true`, so an out-of-date branch cannot merge. This repo has `strict: false` — nothing to update. (And the TypeScript copy is broken anyway: it POSTs to an endpoint that is PUT, 404s on every call, and swallows it — payments#449.)

**Nothing is ever checked out.** A Dependabot branch is untrusted content; running any of it — a script, a build step, a test — with the App token in the environment is a straightforward privilege escalation. The whole policy lives in `if:` expressions over `dependabot/fetch-metadata`, which needs no working tree. nvm-monorepo runs a policy script and pins it to `base.sha` for exactly this reason; here there is no script to pin.

**Patch and minor only.** #282 is the argument for holding majors: it bumps `a2a-sdk` 0.3.26 → 1.1.2, and a2a-sdk 1.x removed `TextPart` from `a2a.types`, so eight test modules fail on import. That is code work, not a rubber stamp.

## The grouping change that makes the policy usable

`dependabot/fetch-metadata` reports the **highest** update-type in a grouped PR. With the group as #278 left it (`patterns: ['*']`), a single major would report `semver-major` and hold every routine bump batched with it — the policy would approve almost nothing.

`.github/dependabot.yml` now restricts the version-update group to `minor` + `patch`. Majors fall out of the group and arrive as individual PRs, which is what they want anyway.

## The scheduled `rearm` job

The `pull_request` trigger cannot reach a PR that never receives another event — one opened before this workflow existed, or one whose auto-merge was switched off by a force-push. nvm-monorepo added a second path after a merge-queue eviction left their #3264 green, approved and unmergeable for 7 days.

This job only re-enables auto-merge on PRs that are **already approved**. It cannot approve anything, so it cannot widen the policy above it.

## Known and accepted

A security update that is *also* a major is held too, because a security group cannot be split by update-type the way the version group can. nvm-monorepo hit this — their #3209 held a HIGH advisory for 8 days — and answered it with a policy script reading an advisory feed. That is more machinery than this repo's volume justifies today; the workflow comment says where to look if it ever does.

Also unchanged from #278, and worth knowing when reading this: a Dependabot bump merges with `e2e` never executed, because secrets are not exposed to `dependabot[bot]`. Populating the Dependabot secret store is the fix, and #280 tracks the related gap that a failing post-merge E2E reports the run green.

## Verification

- `actionlint 1.7.12` — clean on the new workflow.
- Both YAML files parse; job `if:` expressions and the parsed group config checked.
- The `rearm` selector was run against the live repository: it correctly selects nothing today, since the only open Dependabot PR (#282) is a held major sitting at `REVIEW_REQUIRED`.
- Evidence that the App token reaches a genuinely Dependabot-triggered run (not a human push to a `dependabot/*` branch): payments run `34837770639`, `actor=dependabot[bot]`, `Generate release bot token: success`, `Approve PR: success`, `Enable auto-merge: success` on a minor bump — and run `34841623895`, same actor, both steps correctly `skipped` on a major.

Refs nevermined-io/internal#1087

https://claude.ai/code/session_01DTr1tgVeYwyJjjZBN9WJWu


r-marques:
:eyes: Reviewing

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -0,0 +1,133 @@
...
+
+# Approves and queues routine Dependabot bumps so they stop needing a human.
+#
+# Why this works HERE and not in nevermined-io/payments: `main`'s protection in
+# this repository requires 1 approving review with `require_code_owner_reviews:
+# false`, so an approval from any identity with write access satisfies it. The
+# TypeScript repo sets `require_code_owner_reviews: true` with no
+# `bypass_pull_request_allowances`, and a GitHub App cannot be a team member, so
+# the same approval never satisfies CODEOWNERS there (see payments#449). Do not
+# copy this file back without checking that setting.
```
r-marques: **🟡 SHOULD FIX — The payments comparison in this header is already false — payments `main` no longer requires code-owner reviews**

Checked live today (`gh api repos/nevermined-io/payments/branches/main/protection`): `require_code_owner_reviews: false`, `bypass_pull_request_allowances: null`, `required_approving_review_count: 1`, `enforce_admins: true`, `strict: true`. So "The TypeScript repo sets `require_code_owner_reviews: true` … the same approval never satisfies CODEOWNERS there" was true when this was written (the panel saw `true` on 2026-09-14) and is not true any more — the setting changed since.

That leaves the header explaining a difference that does not exist, and a reader who copies this file back to `payments` on the strength of it will draw the wrong conclusion in the other direction. The setting that still differs is **`strict`** (`true` there, `false` here), which is the reason `payments` needs `dependabot-update-branches.yml` and this repo does not — the argument the PR body makes, and the one worth keeping in the file.

Suggest rewriting the paragraph around `strict`, and dating any snapshot of another repo's live settings ("as of 2026-09-15") so the next reader re-checks rather than trusts. A wrong reason in a header is the part a reader acts on.

<sub>🤖 pr-review-toolkit · lead · high</sub>

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -0,0 +1,133 @@
...
+  # force-push, or someone toggling it by hand). The sweep only re-enables
+  # auto-merge on PRs that are ALREADY approved — it never approves anything,
+  # so it cannot widen the policy above.
+  schedule:
+    - cron: '17 6 * * *'
+  workflow_dispatch:
+
+permissions:
+  contents: write
+  pull-requests: write
```
r-marques: **🟡 SHOULD FIX — `GITHUB_TOKEN` gets `contents: write` / `pull-requests: write` but nothing in the file writes with it**

Every `gh pr review` / `gh pr merge` here runs under the App token (`steps.bot-token.outputs.token`), whose scope comes from the installation, not from this block. The only consumer of `GITHUB_TOKEN` is `dependabot/fetch-metadata` at line 58, which calls `pulls.listCommits` — `pull-requests: read`. The `rearm` job never touches `GITHUB_TOKEN` at all.

So this block hands a write-capable token to a floating-tag third-party action for no benefit — the same escalation argument the header makes at lines 17–22 about the checkout. Narrow it:

```yaml
permissions:
  contents: read
  pull-requests: read
```

(nvm-monorepo's `deps-unblock.yml` runs its sweep with `contents: read` and does all writes through the App token.) Once it is read-only, the lines 13–15 "GitHub respects `permissions:` for Dependabot PRs" rationale stops being load-bearing — what actually gates a Dependabot-actor run is the *secret store*, see the note on the token step.

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -0,0 +1,133 @@
...
+      github.event_name == 'pull_request' &&
+      github.event.pull_request.user.login == 'dependabot[bot]'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Generate release bot token
+        id: bot-token
+        uses: actions/create-github-app-token@v3
+        with:
+          app-id: ${{ secrets.RELEASE_APP_ID }}
+          private-key: ${{ secrets.RELEASE_APP_PRIVATE_KEY }}
```
r-marques, line 90: **🟡 SHOULD FIX — Confirm the org **Dependabot** secret store reaches this repo before calling the fast path live — the two jobs read `RELEASE_APP_*` from different stores**

`rearm` runs as the schedule/dispatch actor and reads the org **Actions** store — verified reachable: `GET /repos/nevermined-io/payments-py/actions/organization-secrets` lists `RELEASE_APP_ID` / `RELEASE_APP_PRIVATE_KEY` (and `RELEASE_APP_CLIENT_ID`), and `prepare-release.yml` minted a token from them on 2026-09-09 (run 34361656978). That job will work.

`auto-merge` runs as `dependabot[bot]` and reads the org **Dependabot** store instead, which nothing with repo-write access can list (`orgs/nevermined-io/dependabot/secrets` → 403, and there is no repo-level `dependabot/organization-secrets` endpoint). This repo's own Dependabot store holds only `TEST_BUILDER_API_KEY` / `TEST_SUBSCRIBER_API_KEY`. The `payments` evidence (run 34837770639: `actor=dependabot[bot]`, token minted, and `payments`' repo-level Dependabot store is **empty**) proves the org Dependabot secret exists and reaches *that* repo; nvm-monorepo keeps repo-level Dependabot copies (2026-03-05), which hints the org visibility may be "selected repositories" rather than "all". Whether payments-py is in it, I cannot see.

If it is not: every Dependabot PR's fast path goes red at this step (`The 'client-id' (or deprecated 'app-id') input must be set to a non-empty string…` — loud, but not a required check, so nothing blocks), while the daily cron prints "Nothing to re-arm." forever, because it can only re-arm what the fast path approved. A green cron is therefore not a signal that the feature works.

Three ways to close it, any one is enough: an org admin runs `gh api orgs/nevermined-io/dependabot/secrets/RELEASE_APP_ID --jq .visibility` (and `…/repositories` if `selected`); or add repo-level Dependabot copies the way nvm-monorepo does; or watch the first `actor=dependabot[bot]` run of this workflow after merge — the `dependabot.yml` change will make Dependabot re-open #282 as separate PRs, so it comes within a day. Note this PR's own green run (34860574721) proves nothing here: human author, both jobs skipped.

<sub>🤖 pr-review-toolkit · silent-failure-hunter · medium</sub>

r-marques (commented):

Adds a Dependabot auto-approve + auto-merge workflow and restricts the pip version-update group to minor+patch so majors arrive alone. The core design is right for this repo and I could not find a way for a non-Dependabot PR to reach the approve/merge steps: the job gates on `pull_request.user.login`, `fetch-metadata` refuses a PR whose first commit is not a signed `dependabot[bot]` commit, nothing is checked out, and every PR-derived value reaches the shell via `env:`. The branch-protection premise holds live (payments-py `main`: 1 approval, `require_code_owner_reviews: false`, `strict: false`, contexts `e2e`/`lint`/`unit_integration`, `enforce_admins: true`; repo `allow_auto_merge: true`, squash allowed). No blockers.

Two things changed under the PR since it was written, and the rest of the should-fix items are robustness and wording: (1) `payments` `main` now also has `require_code_owner_reviews: false`, so the header's "why HERE and not there" paragraph is stale — the difference that survives is `strict`; (2) the `auto-merge` job depends on the org **Dependabot** secret store reaching this repo, which nothing below org-admin can confirm — the failure would be loud on the fast path but the cron would stay green forever, so it needs a one-line check (or repo-level copies) before the feature is called live. The `rearm` loop aborts on the first refused `gh pr merge` under `set -e`, and its comment both under- and over-states what the sweep reaches.

**What I reviewed**
- Diff range `origin/main...HEAD` — 2 files, +142/−0, head `1b2cb84b`.
- Panel: code-reviewer, silent-failure-hunter, comment-analyzer (no tests or types in the diff).
- Live protection on both repos and payments-py's rulesets (the only one is `copilot_code_review`, which cannot block); repo merge settings; repo-level Actions + Dependabot secrets on payments-py, payments and nvm-monorepo; org Actions secrets as seen from payments-py (`RELEASE_APP_*` present; `prepare-release.yml` minted a token 2026-09-09).
- The two cited evidence runs (`34837770639` minor: token/approve/auto-merge all success as `actor=dependabot[bot]`; `34841623895` major: both steps skipped) and this PR's own run (both jobs skipped on a human PR).
- The `rearm` selector against the live repo (selects nothing; #282 is `REVIEW_REQUIRED`), `--author 'app/dependabot'` positive control.
- `dependabot/fetch-metadata` source (`update-type` = highest in the PR; verifies `commits[0]` only; fails loudly on non-Dependabot / no-metadata), the Dependabot `groups.update-types` reference, gh's `merge.go` (`--auto` bypasses `canMerge`), GitHub's auto-merge-disable rule.
- `actionlint 1.7.12` clean; both YAMLs parse; `deepagents_compat` is not a required context; `e2e`'s `continue-on-error` still reports a check-run `failure`.

**Verdict:** 🟡 Mergeable with nits — 0 blockers, 6 should-fix, 4 nits.

### 🧭 Inline comments (12)

- 🟡 **SHOULD FIX** — The payments comparison in this header is already false — payments `main` no longer requires code-owner reviews — `.github/workflows/dependabot-auto-merge.yml:11`
- 🟡 **SHOULD FIX** — `GITHUB_TOKEN` gets `contents: write` / `pull-requests: write` but nothing in the file writes with it — `.github/workflows/dependabot-auto-merge.yml:38`
- 🟡 **SHOULD FIX** — Confirm the org **Dependabot** secret store reaches this repo before calling the fast path live — the two jobs read `RELEASE_APP_*` from different stores — `.github/workflows/dependabot-auto-merge.yml:52`
- 🟡 **SHOULD FIX** — One refused `gh pr merge --auto` aborts the whole sweep — every PR after it stays un-armed on every daily run — `.github/workflows/dependabot-auto-merge.yml:133`
- 🟡 **SHOULD FIX** — A human push to a `dependabot/*` branch re-approves commits `fetch-metadata` never looked at — gate on `github.actor` like `test.yaml` does — `.github/workflows/dependabot-auto-merge.yml:79`
- 🟡 **SHOULD FIX** — This comment overclaims in both directions — the sweep reaches less than it says, and overrides two human holds it says it cannot — `.github/workflows/dependabot-auto-merge.yml:31`
- 💡 **NIT** — `update-types` is added to the pip group only — the workflow's "majors arrive as their own PRs" comment is true of pip, not of `github-actions` — `.github/dependabot.yml:38`
- 💡 **NIT** — Three factual slips in the hold rationale — and the #278 citation argues against the policy it is cited for — `.github/workflows/dependabot-auto-merge.yml:63`
- 💡 **NIT** — Required-check list is right, but `deepagents_compat` is not required — and #284 rewrites the `e2e` sentence — `.github/workflows/dependabot-auto-merge.yml:89`
- 💡 **NIT** — A held or unparseable update-type leaves both steps `skipped` and the job green with no log line saying why — `.github/workflows/dependabot-auto-merge.yml:93`
- ✅ **NICE** — No checkout, no inline `${{ }}` in any `run:`, and the two jobs are provably exclusive — `.github/workflows/dependabot-auto-merge.yml:22`
- ✅ **NICE** — The grouping change is the piece that makes the policy usable, and the reasoning matches the documented behaviour — `.github/dependabot.yml:33`

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -0,0 +1,133 @@
...
+
+          if [ -z "$prs" ]; then
+            echo "Nothing to re-arm."
+            exit 0
+          fi
+
+          for pr in $prs; do
+            echo "Re-enabling auto-merge on #$pr"
+            gh pr merge --auto --squash --repo "$REPO" "$pr"
+          done
```
r-marques, line 218: **🟡 SHOULD FIX — One refused `gh pr merge --auto` aborts the whole sweep — every PR after it stays un-armed on every daily run**

`set -euo pipefail` (line 115) plus an unguarded `gh pr merge` in the loop: the first non-zero exit ends the step, and every PR later in `gh pr list` order is never attempted — today, tomorrow, until a human clears the first one. The step is red, but its log names only the first PR; the rest read as "not selected".

`--auto` also skips gh's own client-side mergeability check — `merge.go`'s `canMerge` is `if reason == "" || m.autoMerge || m.merged { return nil }` — so a conflicting, draft-converted or otherwise refused PR surfaces only as the API's error. Dependabot PRs go `CONFLICTING` routinely when a sibling lockfile bump merges first, and the selector filters neither `isDraft` nor `mergeable`.

The shape `deps-unblock.yml` uses:

```bash
prs=$(gh pr list --repo "$REPO" --author 'app/dependabot' --state open \
  --json number,reviewDecision,autoMergeRequest,isDraft,mergeable \
  --jq '.[] | select(.reviewDecision == "APPROVED" and .autoMergeRequest == null
                    and (.isDraft | not) and .mergeable != "CONFLICTING") | .number')
…
for pr in $prs; do
  echo "Re-enabling auto-merge on #$pr"
  gh pr merge --auto --squash --repo "$REPO" "$pr" \
    || { echo "::warning::could not re-arm #$pr"; failed=1; }
done
exit "${failed:-0}"
```

While here: print the unfiltered list (`number, reviewDecision, (.autoMergeRequest != null)`) before the jq. "Nothing to re-arm." is printed identically when the selector has drifted — gh validates `--json` field names, but a typo in the jq path (`.reviewDecsion`) silently yields empty, verified.

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -0,0 +1,133 @@
...
+      # Known cost, inherited knowingly: a SECURITY update that is also a major
+      # is held too. The security group cannot be split by update-type the way
+      # the version group can. nvm-monorepo hit this (their #3209 sat 8 days on a
+      # HIGH advisory) and answered it with a policy script reading an advisory
+      # feed. That is more machinery than this repo's volume justifies today; if
+      # a held security major ever costs real time here, port that script.
+      - name: Approve the PR
+        if: >-
+          steps.metadata.outputs.update-type == 'version-update:semver-patch' ||
+          steps.metadata.outputs.update-type == 'version-update:semver-minor'
```
r-marques, line 118: **🟡 SHOULD FIX — A human push to a `dependabot/*` branch re-approves commits `fetch-metadata` never looked at — gate on `github.actor` like `test.yaml` does**

`synchronize` (line 26) fires when a write-access human pushes to a `dependabot/*` branch — the case `test.yaml`'s `e2e` gate explicitly anticipates ("when a human pushes to a `dependabot/*` branch to fix up a bump"). On that event `pull_request.user.login` (line 44) is still `dependabot[bot]`, and `fetch-metadata` verifies **only `commits[0]`** (`verified_commits.ts`: `const { commit, author } = commits[0]` → author must be `dependabot[bot]` with a verified signature). So the App re-approves and re-arms a head that now carries commits nothing inspected, and in doing so undoes a deliberate dismissal of its earlier approval — the one hold the `rearm` sweep does not override.

Not an escalation: only write access can push to that branch in this repo, and that person could approve the PR themselves. But the header's "policy over untrusted content" framing (lines 17–22) invites reading the content as vetted, and it is not.

One clause, mirroring `test.yaml`'s `e2e` job, on both step `if:`s (or the job `if:` at line 42):

```yaml
if: >-
  github.actor == 'dependabot[bot]' &&
  (steps.metadata.outputs.update-type == 'version-update:semver-patch' ||
   steps.metadata.outputs.update-type == 'version-update:semver-minor')
```

A human-fixed bump then needs a human approval, which is the right cost. Residual worth one sentence in the header: GitHub disables auto-merge only when someone *without* write permission pushes (docs, verified), and `dismiss_stale_reviews: false` keeps the App's earlier approval — so an already-armed PR still merges a write-access human's push under it. That half is branch protection, not this file.

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -0,0 +1,133 @@
...
+# instead, and pins it to `base.sha` for exactly this reason.)
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+  # Re-arm PRs that never get another `pull_request` event: anything opened
+  # before this workflow existed, or whose auto-merge was switched off (a
+  # force-push, or someone toggling it by hand). The sweep only re-enables
+  # auto-merge on PRs that are ALREADY approved — it never approves anything,
+  # so it cannot widen the policy above.
```
r-marques: **🟡 SHOULD FIX — This comment overclaims in both directions — the sweep reaches less than it says, and overrides two human holds it says it cannot**

**Less than it says.** "anything opened before this workflow existed" is not reached: the selector (line 123) requires `reviewDecision == "APPROVED"`, and a pre-existing PR nobody approved gets nothing from either job — run live today it selects nothing, #282 sits at `REVIEW_REQUIRED`. And "a force-push" fires `synchronize`, which the `pull_request` trigger already handles. The honest list is: auto-merge switched off by hand, or a PR a human approved without queueing.

**More than it says.** "it cannot widen the policy above" — the selector ignores update-type, so a human-approved **major** (including a security-major the job above holds on purpose) is queued at 06:17, and if its checks are already green `gh pr merge --auto` merges it on the spot (gh merges immediately when `isImmediatelyMergeable`). And "someone toggling it by hand" means **disabling auto-merge is not a hold** in this repo: the sweep re-arms it the next morning. The only holds that survive are dismissing the approval or requesting changes — that is exactly what a maintainer reaching for the toggle needs to read here.

Either is a fine policy; the comment should state the one you mean. If "a human approval IS the merge decision", say so. If the intent is "re-arm what the fast path armed", filter on the App's own review instead of `reviewDecision`:

```bash
--json number,autoMergeRequest,reviews \
--jq '.[] | select(.autoMergeRequest == null and any(.reviews[]; .author.login == "nevermined-release-bot" and .state == "APPROVED")) | .number'
```

Either way, lines 117–118 "deliberately cannot approve" → "does not approve": the job holds a `pull_requests: write` App token, so it can.

.github/dependabot.yml
```diff
@@ -25,8 +25,17 @@ updates:
...
+      # update-type in a group, so a single major in this batch would report
+      # `semver-major` and hold every routine bump with it in
+      # `dependabot-auto-merge.yml`. Majors are excluded here and therefore
+      # arrive as individual PRs, which is what they want anyway — #282 bumped
+      # a2a-sdk to 1.1.2 and removed `TextPart` from `a2a.types`.
       python-version-updates:
         applies-to: version-updates
+        update-types:
+          - 'minor'
+          - 'patch'
```
r-marques, line 38: **💡 NIT — `update-types` is added to the pip group only — the workflow's "majors arrive as their own PRs" comment is true of pip, not of `github-actions`**

The `github-actions` group (lines 52–54, untouched) keeps `patterns: ['*']` with no `update-types`, so a grouped actions PR carrying one major still reports `semver-major` and holds the batch — the exact thing the workflow comment at lines 65–68 says the config prevents. In practice every action here is pinned to a bare major (`@v7`, `@v3`, `@v2`), so every actions bump *is* a major and is held either way — #281 today ("bump the github-actions group with 6 updates") was merged by hand, which is how these will keep going. Fine as a policy, but say so: either add the same `update-types` to that group (each action major then arrives and is held alone) or scope the workflow comment to pip.

Related: `dependabot/fetch-metadata` v3.1.0 exists (v3.0.0 2026-03-26, Node 24), so the `@v2` at workflow line 56 will itself be one of those held actions majors.

The pip half is right, checked against the options reference: `update-types` takes `patch`/`minor`/`major`, applies to version updates only (so the security group indeed cannot be split), and "any outdated dependencies that do not match a rule are updated in individual pull requests".

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -0,0 +1,133 @@
...
+      - name: Fetch Dependabot metadata
+        id: metadata
+        uses: dependabot/fetch-metadata@v2
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      # Patch and minor only. A major is held for review on purpose: #282 bumped
+      # a2a-sdk 0.3.26 -> 1.1.2, which removed `TextPart` from `a2a.types` and
+      # broke eight test modules on import — exactly the kind of change that
+      # needs a human, and the kind that #278 raised the runtime floor for.
```
r-marques: **💡 NIT — Three factual slips in the hold rationale — and the #278 citation argues against the policy it is cited for**

(a) "broke eight test modules on import" — #282's run 34859666178 reports **44 errors during collection**. The only `TextPart` import is `payments_py/x402/a2a.py:18`, so the whole suite fails to import; a stronger argument than eight, and the number a reader will check.

(b) "#282 bumped" — #282 is open; "bumps". Same slip at `dependabot.yml:32–33`, where read literally the PR did the removing (it was a2a-sdk 1.x).

(c) "the kind that #278 raised the runtime floor for" — #278 raised `a2a-sdk` to `^0.3.25` over a **patch** break (0.3.24 → 0.3.25, `consume_and_break_on_interrupt` returning a triple; its own body says "the arity landed in 0.3.25"). That is exactly the update-type this workflow auto-merges, so the citation cuts the other way. The caveat a reader actually needs here: three runtime deps are 0.x (`a2a-sdk ^0.3.25`, `fastapi ^0.141.1`, `uvicorn ^0.53.0`), where `semver-minor` is breaking by convention, and on a Dependabot PR `unit_integration` is their only gate.

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -0,0 +1,133 @@
...
+        run: gh pr review --approve "$PR_URL"
+        env:
+          PR_URL: ${{ github.event.pull_request.html_url }}
+          GH_TOKEN: ${{ steps.bot-token.outputs.token }}
+
+      # `--auto` queues the merge behind the required checks (`lint`,
+      # `unit_integration`, `e2e`) rather than merging now. On a Dependabot PR
+      # `e2e` reports `skipped`, which satisfies its required check — see the
+      # comment on that job in `test.yaml`, and note what it says about the
+      # coverage that buys.
```
r-marques: **💡 NIT — Required-check list is right, but `deepagents_compat` is not required — and #284 rewrites the `e2e` sentence**

The three contexts match live protection (`e2e`, `lint`, `unit_integration`; `strict: false`; auto-merge enabled on the repo; squash allowed). Two things a reader of this comment needs:

(a) **`deepagents_compat` is not a required check**, so a red run there does not hold the merge. It is red on #282 today alongside `unit_integration`; a langchain/langgraph minor could redden it alone and still auto-merge. Now that bumps merge without a human, whether to make it required is a maintainer call — worth deciding on this PR rather than after the first such merge.

(b) "`e2e` reports `skipped`" is rewritten by #284 (open), which lifts the actor gate on `e2e`. Whichever of the two merges second should rephrase this to something that survives both states — e.g. "queues behind the three required checks; `test.yaml` says what `e2e` does on a Dependabot PR".

Verified in passing, since it matters for #284: `e2e`'s `continue-on-error: true` does **not** turn a failure green at the check-run level (run 34108722286: job `failure`, check-run `failure`, workflow run `success`), so as a required check it still blocks a PR.

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -0,0 +1,133 @@
...
+      - name: Enable auto-merge
+        if: >-
+          steps.metadata.outputs.update-type == 'version-update:semver-patch' ||
+          steps.metadata.outputs.update-type == 'version-update:semver-minor'
```
r-marques, line 135: **💡 NIT — A held or unparseable update-type leaves both steps `skipped` and the job green with no log line saying why**

`fetch-metadata`'s `calculateUpdateType` returns `''` when either version string is missing; a fully unparseable PR does fail loudly (`PR does not contain metadata`), but the empty case, and every deliberate hold, produce the same picture: both steps `skipped`, job `success`, nothing in the log. Indistinguishable in the checks list from "approved".

Cheap fix, the shape nvm-monorepo's workflow uses ("Explain a hold"):

```yaml
      - name: Explain a hold
        if: >-
          steps.metadata.outputs.update-type != 'version-update:semver-patch' &&
          steps.metadata.outputs.update-type != 'version-update:semver-minor'
        env:
          UPDATE_TYPE: ${{ steps.metadata.outputs.update-type }}
        run: echo "::notice::held for human review — update-type='${UPDATE_TYPE}'"
```

Every run then says what it decided, and an empty type shows up as `update-type=''` rather than as silence.

<sub>🤖 pr-review-toolkit · silent-failure-hunter · medium</sub>

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -0,0 +1,133 @@
...
+# Uses `pull_request`, NOT `pull_request_target`: GitHub has respected the
+# `permissions:` key for Dependabot PRs since October 2021, which makes
+# `pull_request_target` unnecessary and strictly less safe here.
+#
+# Nothing is ever checked out. A Dependabot branch is untrusted content, and
+# running any of it — a script, a build step, a test — while the App token is in
+# the environment would be a straightforward privilege escalation. The whole
+# policy therefore lives in `if:` expressions over `dependabot/fetch-metadata`
+# output, which needs no working tree. (nvm-monorepo runs a policy script
+# instead, and pins it to `base.sha` for exactly this reason.)
```
r-marques, line 30: **✅ NICE — No checkout, no inline `${{ }}` in any `run:`, and the two jobs are provably exclusive**

Verified: neither job checks anything out; every PR-derived value reaches the shell through `env:` (lines 82, 96, 113) rather than inline expression interpolation, so there is no script-injection surface and no PR-head code ever runs under the App token. The job `if:`s are mutually exclusive (`event_name == 'pull_request' && user.login == …` vs `event_name != 'pull_request'`) and evaluate safely on non-PR payloads — this PR's own run 34860574721 shows both jobs skipped on a human `pull_request`. `actionlint 1.7.12` is clean; both YAMLs parse.

The cited evidence runs hold up: payments 34837770639 is `actor=dependabot[bot]` on #409 (axios minor) with token → approve → auto-merge all `success`, and 34841623895 is the same actor on #326 (typescript 6 major) with both gated steps `skipped`. Repo settings here are compatible (`allow_auto_merge: true`, `allow_squash_merge: true`, `delete_branch_on_merge: true`), and the live `rearm` selector selects nothing today (#282 is `REVIEW_REQUIRED`), as the PR body says.

<sub>🤖 pr-review-toolkit · lead · high</sub>

.github/dependabot.yml
```diff
@@ -25,8 +25,17 @@ updates:
         applies-to: security-updates
         patterns:
           - '*'
+      # Minor and patch only. `dependabot/fetch-metadata` reports the HIGHEST
+      # update-type in a group, so a single major in this batch would report
+      # `semver-major` and hold every routine bump with it in
+      # `dependabot-auto-merge.yml`. Majors are excluded here and therefore
+      # arrive as individual PRs, which is what they want anyway — #282 bumped
+      # a2a-sdk to 1.1.2 and removed `TextPart` from `a2a.types`.
```
r-marques, line 33: **✅ NICE — The grouping change is the piece that makes the policy usable, and the reasoning matches the documented behaviour**

Checked against the Dependabot options reference and `fetch-metadata`'s `action.yml`: `update-type` is "the highest semver change being made by this PR", `update-types` limits a group to `patch`/`minor`/`major` for version updates only, and unmatched dependencies "are updated in individual pull requests". So a major genuinely falls out of the group and arrives alone, and the security-group cost (a security major is held) is stated honestly rather than papered over. Good that the comment names the concrete case (#282) that motivated it.

<sub>🤖 pr-review-toolkit · lead · high</sub>

r-marques (approved):
🟡 **Approved with comments** — no blockers; 6 should-fix, 4 nit(s) left inline.

Inline review: https://github.com/nevermined-io/payments-py/pull/283#pullrequestreview-5206850785

PR #284, opened 2026-09-14
ci: run E2E on Dependabot PRs now that the Dependabot secret store is populated

eruizgar91:
## What changed

`TEST_SUBSCRIBER_API_KEY` and `TEST_BUILDER_API_KEY` now exist in this repository's **Dependabot** secret store, so the `github.actor != 'dependabot[bot]'` clause added in #278 is dropped and `e2e` runs for real on Dependabot PRs instead of reporting `skipped`.

The fork guard stays. Fork PRs genuinely get no secrets from either store.

Beyond the `if:`, this PR now also carries the fixes from review round 1 — see **Review round 1** below; the one behaviour change among them is in `tests/e2e/test_a2a_e2e.py`.

## Why this is worth doing rather than leaving #278's skip in place

#278's skip was the right unblock — 18 Dependabot PRs were unmergeable because `e2e` is a required check it could never pass — but it bought mergeability by deleting the signal. A bump merged on unit tests alone.

That is a poor trade in this particular SDK. a2a-sdk 0.3.25 changed the arity of `consume_and_break_on_interrupt`, the call that carries the credit burn (#278). E2E is where a change of that shape surfaces — though only now that the assertion actually fails; see below.

## Review round 1

**The security trade was described by the wrong mechanism.** The keys are injected as **step-level** `env:` on `Run E2E tests` only, so `poetry install` and its package build hooks run *without* them. The real exposure is `poetry run pytest -m "slow"`, which imports and executes the bumped dependency — and anything a build hook planted in the venv — with the keys in `os.environ` and in use over the network. A reader acting on the old sentence would have hardened the install and left the actual surface untouched.

**And it was not this job's trade to make.** `unit_integration` has no `if:` at all, runs the same checkout and `poetry install` on the same Dependabot branch, and was mapping the identical two secrets — so reverting the `e2e` clause would not have undone the exposure. Its `env:` block is now removed: `grep` finds no reader of either key outside `tests/e2e/`, which `-m "not slow"` excludes, and `tests/e2e/conftest.py` only `os.getenv`s them at import. That halves the surface rather than documenting it.

**The risk level was overstated in one direction and under-warned in the other.** The Dependabot store was seeded *from* the committed fallbacks in `conftest.py`, so it holds keys already public in this repository: a Dependabot run exposes nothing a `git clone` does not, and the incremental risk today is zero. It stops being zero if someone overwrites the store with the rotated **Actions** keys — which the previous version of this body offered as a command with no caveat. That warning now sits next to the command.

**The rotation note named the wrong pair.** It said updating the Actions secret does not update the Dependabot one and E2E "starts failing" — but the Actions secrets *were* rotated on 2026-07-24 and Dependabot runs are unaffected. The pair that must move together is `conftest.py` ↔ the Dependabot store. That, the two-store/actor table, the `gh secret set --app dependabot` commands and the verification procedure now live in **`tests/e2e/README.md` → "Rotating the CI keys"**, where a rotator is actually reading, with a pointer from `CLAUDE.md`.

**"E2E is where that shows up" was not true for the burn path it cited.** In `tests/e2e/test_a2a_e2e.py` the credit-burn assertion sat inside a `try:` whose `except Exception` caught it — `AssertionError` is an `Exception` — so a bump that made the burn a no-op printed a ❌ line and the test still passed, under a final `✅ ... with verified credit burning`. `balance_before` also became `None` on any error, skipping the check outright. Both are fixed: the before-balance read now fails the test, and the after-balance block is no longer wrapped. Combined with #283's auto-merge this was the difference between a gate and a decoration.

**The proof plan would have proved nothing.** Rebasing #282 to observe `e2e` run will show `e2e: skipped` again — via `needs: unit_integration`, which fails at collection on that PR (`ImportError: cannot import name 'TextPart' from 'a2a.types'`, run `34859666178`). A job whose `needs` failed is skipped before its `if:` is evaluated, and the Jobs API reports both causes identically. Corrected plan below.

Nits folded in: the fork-guard comment regained the fact that makes it mergeable (`skipped` satisfies the required check — #206, and #215/#181 merged that way); the store is selected by the **actor of the triggering push**, not the PR, so a bad copy looks like a per-push flake on one PR; `continue-on-error` does **not** green a failed required check (it rewrites the run conclusion, branch protection reads the check-run conclusion); the rotting `77 passed / 26 skipped` counts are gone; and `conftest.py` now says why `os.getenv(NAME, FALLBACK)` must keep that exact shape.

## Verification

- `poetry run black --check .` — 191 files unchanged.
- `poetry run pytest -m "not slow"` — **1086 passed, 1 skipped**.
- `poetry run pytest -m slow` — full E2E suite green against staging.
- `tests/e2e/test_a2a_e2e.py::TestA2AE2EFlow::test_blocking_flow_with_credit_burning` run on its own with the assertion now live: `🔥 Credits actually burned: 1` / `✅ Verified: Exactly 1 credits were burned`. AST check confirms no `except Exception` encloses that assertion any more and `balance_before = None` is gone.
- `test.yaml` parses; `unit_integration` now has zero `env:` blocks, `e2e` keeps its step-level one, and the `if:` is the fork guard alone: `github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository`.
- `gh api repos/nevermined-io/payments-py/dependabot/secrets` → both secrets present.

**How to confirm the keys really work**, since this PR's own run cannot: the first Dependabot PR whose `unit_integration` *passes*. Look for `e2e` not `skipped`, and a summary with no `errors` and no `Invalid NVM API Key`. `passed > 0` is not evidence — 11 tests passed on empty keys before #278. The pip group is stuck on a2a-sdk 1.1.2 until that break is fixed or `@dependabot ignore`d, so the github-actions group's next PR will likely get there first.

## Ordering with #283

#283 merged first as `de23178`; its `e2e` comment was already written for the post-#284 world, so nothing there needs correcting.

Refs nevermined-io/internal#1087


r-marques:
:eyes: Reviewing

.github/workflows/test.yaml
```diff
@@ -70,34 +70,43 @@ jobs:
...
+    # nothing warns you — E2E simply starts failing on Dependabot PRs only. The
+    # values seeded into the Dependabot store were the committed staging-sandbox
+    # fallbacks from `tests/e2e/conftest.py`, which had not been touched since
+    # 2025-12-03 while the Actions secrets were rotated on 2026-07-24; they are
+    # therefore probably NOT the same keys, only equivalent ones. Both pass the
+    # suite (77 passed / 26 skipped either way).
+    #
+    # Worth knowing about the trade: this job checks out the Dependabot branch
+    # and runs `poetry install`, which executes package build hooks — arbitrary
+    # code from the bumped dependency — with these keys in the environment. That
```
r-marques: **🟡 SHOULD FIX — The security-trade mechanism is wrong: `poetry install` never sees these keys**

The secrets are injected as **step-level** `env:` on the `Run E2E tests` step only (lines 126–128; the only other `env:` in the file is line 30, same shape). The `Install the project dependencies` step (121–122) has no `${{ secrets.* }}` reference, so package build hooks run with **no** keys in the environment.

The exposure is real, but it lives in `poetry run pytest -m "slow"` (line 129): the suite imports and executes the bumped dependency — and anything a build hook planted in the venv (`.pth`, `sitecustomize`) — with the keys in `os.environ`.

A wrong mechanism is the part a reader acts on: someone "hardening" this by sandboxing the install (`--only-binary`, no build isolation) would leave the actual surface untouched, and someone reading it as "the keys are already job-wide" might hoist `env:` to the job level. Suggested wording:

```yaml
    # Worth knowing about the trade: the `Run E2E tests` step imports and runs
    # the bumped dependency — arbitrary code — with these keys in its
    # environment (they are step-scoped; `poetry install` runs without them).
```

The PR body carries the same sentence and should say the same thing.

<sub>🤖 pr-review-toolkit · code-reviewer, silent-failure-hunter, comment-analyzer, pr-test-analyzer · high</sub>

.github/workflows/test.yaml
```diff
@@ -70,34 +70,43 @@ jobs:
...
+    # is the reason GitHub keeps the two stores apart. Acceptable here because
```
r-marques: **🟡 SHOULD FIX — This is not this job's trade — `unit_integration` already hands the same keys to Dependabot-branch code, and to nothing that reads them**

`unit_integration` (lines 12–33) has no `if:` at all, runs the same checkout + `poetry install` + pytest on the same Dependabot branch, and maps the identical two secrets at step level (30–32). On a Dependabot PR it has been reading the Dependabot store since it was populated at 2026-09-14T15:35Z — with or without this PR. So the exposure this paragraph describes as a deliberate call was already made, and reinstating the actor clause on `e2e` would not undo it; the next person to revisit the decision would revert the wrong line.

Two facts worth stating here so that does not happen:

1. `grep -rn TEST_SUBSCRIBER_API_KEY\|TEST_BUILDER_API_KEY` finds **no reader outside `tests/e2e/`**, and `unit_integration` passed on both pre-#278 Dependabot runs with the store empty (`30943984001`, `30929380126`). Its `env:` block at 30–32 is dead — removing it halves the surface this paragraph worries about, and would make "this job" true. If that is out of scope here, name it.
2. What this PR *adds* is a second process that uses the keys **over the network** — the honest framing of the delta.

<sub>🤖 pr-review-toolkit · code-reviewer, comment-analyzer · high</sub>

.github/workflows/test.yaml
```diff
@@ -70,34 +70,43 @@ jobs:
...
+    # these are staging-sandbox test accounts whose equivalents are already
+    # committed in the repository, but it is not zero risk.
```
r-marques: **🟡 SHOULD FIX — The risk level is misstated, and the one action that would make it real is unwarned**

Line 91–92 says the store was seeded **from** the committed fallbacks — so it holds the very keys already public in `tests/e2e/conftest.py:40–46`, not "equivalents". A Dependabot-triggered run therefore exposes nothing a `git clone` does not: the incremental risk today is zero, not "not zero".

The risk becomes real only if someone overwrites the store with the rotated **Actions** keys — which the PR body offers (`gh secret set … --app dependabot`) with no caveat. That is the sentence a future reader needs:

> Acceptable **because** the Dependabot store holds the already-committed staging keys. Do **not** overwrite it with the Actions keys unless you accept exposing a non-public key to dependency code on every Dependabot PR.

(While you are here: "a test account" undersells what the committed credential is — the fallback JWT never expires, `exp` ≈ 2126, and embeds an `o11y` observability key. Pre-existing and already public, nothing for this PR to change, but the trade paragraph should describe the thing it is trading.)

<sub>🤖 pr-review-toolkit · comment-analyzer, code-reviewer · high</sub>

.github/workflows/test.yaml
```diff
@@ -70,34 +70,43 @@ jobs:
...
-    # repository's *Dependabot* secret store would let E2E run on Dependabot
-    # PRs once this clause is dropped, and make it unnecessary — the better end
-    # state for a repo where a bump can rewire the credit-burn path. Until then a
-    # Dependabot bump merges with E2E never executed, and the post-merge `push`
-    # run below is `continue-on-error` and not a required check — a failing
-    # post-merge E2E still reports the run green (measured: run 34108722286,
-    # `e2e: failure` inside a `success` run). Tracked in #280.
+    # Operational note for whoever rotates these: the two stores hold SEPARATE
+    # copies. Updating the Actions secret does NOT update the Dependabot one, and
+    # nothing warns you — E2E simply starts failing on Dependabot PRs only. The
```
r-marques: **🟡 SHOULD FIX — Operational note names the wrong pair and carries the warning without the remedy**

Two problems in the rotation note, both of which send a rotator the wrong way:

**Wrong pair.** "Updating the Actions secret does NOT update the Dependabot one … E2E simply starts failing" — the paragraph's own evidence contradicts it: the Actions secrets *were* rotated on 2026-07-24 and the fallback/Dependabot keys still pass in 2026-09 (line 94–95). Rotating the Actions secret does not break Dependabot runs. What breaks them is **revoking the committed fallback accounts' keys** server-side — and then every local run with `TEST_*` unset breaks too, so "on Dependabot PRs only" is off. The pair that must move together is `tests/e2e/conftest.py` fallbacks ↔ Dependabot store (same value, also mirrored in `tests/e2e/README.md:63-64,144-145`), not Actions ↔ Dependabot.

**No remedy.** The note tells "whoever rotates these" that the copy is separate and nothing warns you, but the command that fixes it lives only in the PR body. Put it where the rotator is reading:

```yaml
    #   gh secret set TEST_SUBSCRIBER_API_KEY --app dependabot --repo nevermined-io/payments-py
    #   gh secret set TEST_BUILDER_API_KEY    --app dependabot --repo nevermined-io/payments-py
```

GitHub never returns a secret's value, so this is also the only way to make two stores provably identical. And a rotator works in Settings → Secrets or `gh secret set`, not 20 lines into a job comment — one pointer line in `CLAUDE.md` ("CI Workflow") or `tests/e2e/README.md` next to the env-var docs would make this findable.

<sub>🤖 pr-review-toolkit · comment-analyzer, code-reviewer, silent-failure-hunter · high</sub>

.github/workflows/test.yaml
```diff
@@ -70,34 +70,43 @@ jobs:
...
     #
-    # Gated on `github.actor`, the actor of the triggering event, and not on the
-    # PR author: when a human pushes to a `dependabot/*` branch to fix up a bump,
-    # the actor is that human, secrets are available, and E2E runs as normal.
+    # `TEST_SUBSCRIBER_API_KEY` / `TEST_BUILDER_API_KEY` now exist in the
+    # Dependabot store, so the job runs for real on a Dependabot PR instead of
+    # reporting `skipped`. That matters more here than in most repos: a
+    # dependency bump in this SDK can rewire the credit-burn path — a2a-sdk
+    # 0.3.25 changed the arity of the call that carries it (#278) — and E2E is
+    # where that shows up. Skipping it meant bumps merged on unit tests alone.
```
r-marques: **🟡 SHOULD FIX — "E2E is where that shows up" is not true for the a2a burn path this sentence cites — its assertion is swallowed**

Pre-existing and outside this diff — anchored here because lines 83–86 make it the justification for the whole change. In `tests/e2e/test_a2a_e2e.py` the blocking-flow credit check does:

```python
            try:
                ...
                assert credits_burned == credits_to_burn, (...)   # line ~959
            except Exception as e:                                # line ~970
                print(f"❌ Error getting balance after: {e}")
            print("✅ E2E blocking flow test passed with verified credit burning")
```

`AssertionError` is a subclass of `Exception`, so "0 credits burned" prints a ❌ line and the test **passes**; and `balance_before = None` on any error (line ~867) skips the check outright. Only a hard break (non-200 / the #278 arity `TypeError`) is caught. So after this PR plus #283's auto-merge, a Dependabot bump that makes the a2a burn silently no-op still merges green. The x402 settle tests (`test_x402_e2e.py:293,316,595,649`) assert `credits_redeemed` / `balance_updated` loudly and are fine.

Either fix it in this PR (it is small: `except AssertionError: raise`, and fail rather than `None` when the before-balance cannot be read) or soften lines 85–86 to "E2E is where a hard break shows up" and file the test fix — a comment that promises a gate the suite does not enforce is the shape that bit on #280.

.github/workflows/test.yaml
```diff
@@ -70,34 +70,43 @@ jobs:
...
+    # test errored in `_parse_nvm_api_key` with "ValueError: not enough values to
+    # unpack (expected 2, got 1)". With `e2e` a required check, that made every
+    # Dependabot PR unmergeable; 18 of them piled up that way.
     #
-    # Gated on `github.actor`, the actor of the triggering event, and not on the
-    # PR author: when a human pushes to a `dependabot/*` branch to fix up a bump,
-    # the actor is that human, secrets are available, and E2E runs as normal.
+    # `TEST_SUBSCRIBER_API_KEY` / `TEST_BUILDER_API_KEY` now exist in the
+    # Dependabot store, so the job runs for real on a Dependabot PR instead of
+    # reporting `skipped`. That matters more here than in most repos: a
```
r-marques: **🟡 SHOULD FIX — The proof plan (rebase #282) will show `e2e: skipped` again — via `needs`, not the clause**

The PR body says the behaviour "can only be confirmed on a genuinely Dependabot-triggered run" and proposes rebasing #282 after merge, expecting `e2e` to run rather than skip. It will skip again: `e2e` has `needs: unit_integration` (line 67), and on #282 `unit_integration` fails at **collection** (`ImportError: cannot import name 'TextPart' from 'a2a.types'` — run `34859666178`, created 4 min after #278 merged: `unit_integration: failure`, `e2e: skipped`, 0 steps). A job whose `needs` failed is skipped before its `if:` is evaluated, and the Jobs API shows an identical `{"conclusion":"skipped","steps":0}` for both causes — so a reviewer following the plan would read "the clause removal did not work" from a run that never consulted the clause.

Fix the plan: "the first Dependabot PR whose `unit_integration` passes" (the pip group is stuck on a2a-sdk 1.1.2 until that break is fixed or `@dependabot ignore`d — a maintainer's call; the github-actions group's next PR is release-driven). And add one line here, because the same ambiguity will confuse whoever next sees `e2e: skipped` on a Dependabot PR:

```yaml
    # A Dependabot PR whose unit suite fails still shows `e2e: skipped` —
    # that is `needs`, not a secrets problem.
```

What "ran with real keys" must look like on that first qualifying run: `unit_integration: success`, `e2e` not `skipped`, and a step summary with **no** `errors` and no `Invalid NVM API Key` line — `passed > 0` alone is not evidence, since 11 tests passed on empty keys in the pre-#278 runs.

r-marques (commented):

The `if:` change is right: dropping `github.actor != 'dependabot[bot]'` leaves exactly the fork guard, which holds for every event shape I walked (push to `main`, same-repo PR, Dependabot PR — same-repo branch, reads the now-populated Dependabot store — human fix-up push to a `dependabot/*` branch, fork PR → skipped). The Dependabot store holds exactly the two names the job reads, and an *empty* secret fails loudly rather than silently falling back (`${{ secrets.X }}` on an unset store expands to `""`, which `os.getenv` returns as-is, and `"".split(":")` raises in `_parse_nvm_api_key` — confirmed from the pre-#278 Dependabot runs `30943984001` / `30929380126`, 48 errors each). Nothing here can merge a Dependabot bump on a skipped E2E that the old clause would have caught.

What needs work is the **comment**, which is the part a future reader acts on — and four of its load-bearing claims are wrong or misplaced: the secrets are **step-level** `env:` on the pytest step only, so `poetry install` and its build hooks never see them (`test.yaml:98`); the exposure is not this job's trade — `unit_integration` has no `if:` and hands the same keys to the same Dependabot-branch code, where nothing even reads them (`:100`); the store holds the *already-committed* keys, so today's incremental risk is zero and the one action that changes that (overwriting the store with the Actions keys, which the PR body offers) is unwarned (`:101`); and the pair that must rotate together is `conftest.py` fallbacks ↔ Dependabot store, not Actions ↔ Dependabot (`:89`). Separately, the PR body's proof plan — rebase #282 — will show `e2e: skipped` again through `needs: unit_integration` (that PR fails at collection on a2a-sdk 1.1.2, run `34859666178`), indistinguishable from the old clause (`:82`); and the a2a E2E test the comment leans on ("E2E is where that shows up") swallows its own burn assertion inside `except Exception` — pre-existing, outside the diff, but it means the a2a burn path is *not* gated by this job today (`:84`).

**What I reviewed**
- Diff range `origin/main...HEAD` — 1 file, +34/−25, head `08b27cc6`.
- Panel: code-reviewer, silent-failure-hunter, comment-analyzer, pr-test-analyzer (security-sensitive: CI secrets exposed to third-party branch code) — all four delivered.
- Verified with `gh api`: Dependabot store = exactly `TEST_SUBSCRIBER_API_KEY` + `TEST_BUILDER_API_KEY` (created 2026-09-14T15:35Z); Actions copies `updated_at` 2026-07-24; `main` requires `e2e`, `lint`, `unit_integration`, 1 approval, no CODEOWNERS, `enforce_admins: true`.
- Verified the cited run `34108722286`: `push` to `main`, run `success`, job `e2e: failure` — the `continue-on-error` measurement is accurate.
- Verified the committed fallbacks in `tests/e2e/conftest.py:39-47` date from `083e385` (2025-12-03).
- Grepped for readers of either variable outside `tests/e2e/`: none — and `unit_integration` passed on both pre-#278 Dependabot runs with the store empty, so its `env:` block is dead.
- Parsed the YAML and the reduced expression myself; `actionlint` is not installed on this box so that one I could not re-run.

**Verdict:** 🟡 Mergeable with nits — 0 blockers, 6 should-fix, 6 nits.

### 🧭 Inline comments (13)

- 🟡 **SHOULD FIX** — The security-trade mechanism is wrong: `poetry install` never sees these keys — `.github/workflows/test.yaml:99`
- 🟡 **SHOULD FIX** — This is not this job's trade — `unit_integration` already hands the same keys to Dependabot-branch code, and to nothing that reads them — `.github/workflows/test.yaml:100`
- 🟡 **SHOULD FIX** — The risk level is misstated, and the one action that would make it real is unwarned — `.github/workflows/test.yaml:102`
- 🟡 **SHOULD FIX** — Operational note names the wrong pair and carries the warning without the remedy — `.github/workflows/test.yaml:90`
- 🟡 **SHOULD FIX** — "E2E is where that shows up" is not true for the a2a burn path this sentence cites — its assertion is swallowed — `.github/workflows/test.yaml:86`
- 🟡 **SHOULD FIX** — The proof plan (rebase #282) will show `e2e: skipped` again — via `needs`, not the clause — `.github/workflows/test.yaml:83`
- 💡 **NIT** — "77 passed / 26 skipped either way" rots, and overclaims what has been exercised — `.github/workflows/test.yaml:95`
- 💡 **NIT** — Say what `continue-on-error` does on a PR, not only on `push` — it does not green a required check — `.github/workflows/test.yaml:110`
- 💡 **NIT** — "on Dependabot PRs only" — the store is chosen by the actor of the triggering push, not by the PR — `.github/workflows/test.yaml:90`
- 💡 **NIT** — The loud failure depends on `os.getenv(NAME, FALLBACK)` staying exactly that — worth a sentence so nobody "improves" it — `.github/workflows/test.yaml:92`
- 💡 **NIT** — The rewrite drops the one fact that makes the retained fork guard mergeable — `.github/workflows/test.yaml:79`
- 💡 **NIT** — 34 comment lines for a 3-line `if:` change — most of it is PR-body rationale, not what the file must carry — `.github/workflows/test.yaml:73`
- ✅ **NICE** — The residual expression is right for every event shape, and the failure mode it relies on is real — `.github/workflows/test.yaml:109`

.github/workflows/test.yaml
```diff
@@ -70,34 +70,43 @@ jobs:
...
-    # post-merge E2E still reports the run green (measured: run 34108722286,
-    # `e2e: failure` inside a `success` run). Tracked in #280.
+    # Operational note for whoever rotates these: the two stores hold SEPARATE
+    # copies. Updating the Actions secret does NOT update the Dependabot one, and
+    # nothing warns you — E2E simply starts failing on Dependabot PRs only. The
+    # values seeded into the Dependabot store were the committed staging-sandbox
+    # fallbacks from `tests/e2e/conftest.py`, which had not been touched since
+    # 2025-12-03 while the Actions secrets were rotated on 2026-07-24; they are
+    # therefore probably NOT the same keys, only equivalent ones. Both pass the
+    # suite (77 passed / 26 skipped either way).
```
r-marques: **💡 NIT — "77 passed / 26 skipped either way" rots, and overclaims what has been exercised**

A point-in-time count that nobody updates when a test is added — and per the PR body it was measured **locally** on the fallbacks and in CI on the Actions keys; the Dependabot store copy itself has not produced a run yet (a typo in the seeded value would make this line wrong on merge day). Note too that `tests/e2e/test_organizations_e2e.py:55-70` turns a wrong-account key into **skips**, not failures — so the two-way comparison is the right rotation procedure precisely because a *mismatched skip count* is the only signal a different-account key gives, but an equal count is not proof the same tests ran.

Suggest: "the committed fallbacks pass locally (2026-09-14); the first Dependabot-triggered run is the proof for the store copy — compare the skip count, not just the pass count" and drop the numbers. Also "rotated on 2026-07-24" is inferred from `updated_at`, which proves a write, not a value change — hence, correctly, the "probably".

<sub>🤖 pr-review-toolkit · code-reviewer, comment-analyzer, silent-failure-hunter · medium</sub>

.github/workflows/test.yaml
```diff
@@ -70,34 +70,43 @@ jobs:
...
+    # The post-merge `push` run below is still `continue-on-error` and not a
+    # required check, so a failing E2E on `main` reports the run green (measured:
+    # run 34108722286, `e2e: failure` inside a `success` run). Tracked in #280.
     if: >-
-      github.actor != 'dependabot[bot]' &&
-      (github.event_name != 'pull_request' ||
-      github.event.pull_request.head.repo.full_name == github.repository)
+      github.event_name != 'pull_request' ||
+      github.event.pull_request.head.repo.full_name == github.repository
     continue-on-error: true
```
r-marques: **💡 NIT — Say what `continue-on-error` does on a PR, not only on `push` — it does not green a required check**

`continue-on-error: true` sits on a **required** check and this paragraph explains only its push-to-main effect, so a reader will assume it also greens a failed PR check. It does not: branch protection reads the **check-run** conclusion, and `continue-on-error` rewrites only the **workflow-run** conclusion. Evidence: every pre-#278 Dependabot PR run shows `e2e: failure` inside a `success` run (`30943984001`, `30929380126`) and those PRs were blocked — which is exactly the run `34108722286` shape, just on a PR. Likewise `needs: unit_integration` (line 67) can only skip `e2e` when a check that is itself required is already red, so there is no skip-to-green path on a same-repo PR.

One sentence closes it both ways: "On a PR the `e2e` check-run is still `failure` and blocks; this flag only affects the run conclusion, which is why the `push` run reports green. When reading a Dependabot run, read the job, not the run."

<sub>🤖 pr-review-toolkit · silent-failure-hunter, pr-test-analyzer · high</sub>

.github/workflows/test.yaml
r-marques: **💡 NIT — "on Dependabot PRs only" — the store is chosen by the actor of the triggering push, not by the PR**

A `@dependabot rebase` reads the Dependabot store; a human fix-up push to the same `dependabot/*` branch reads the Actions store and passes. So a bad Dependabot copy looks like a **per-push flake on one PR**, which is a worse diagnosis than "fails on Dependabot PRs". The deleted text had this right ("Gated on `github.actor`, the actor of the triggering event, and not on the PR author: when a human pushes to a `dependabot/*` branch … secrets are available") and it is still true — keep one clause of it.

.github/workflows/test.yaml
```diff
@@ -70,34 +70,43 @@ jobs:
...
-    # state for a repo where a bump can rewire the credit-burn path. Until then a
-    # Dependabot bump merges with E2E never executed, and the post-merge `push`
-    # run below is `continue-on-error` and not a required check — a failing
-    # post-merge E2E still reports the run green (measured: run 34108722286,
-    # `e2e: failure` inside a `success` run). Tracked in #280.
+    # Operational note for whoever rotates these: the two stores hold SEPARATE
+    # copies. Updating the Actions secret does NOT update the Dependabot one, and
+    # nothing warns you — E2E simply starts failing on Dependabot PRs only. The
+    # values seeded into the Dependabot store were the committed staging-sandbox
+    # fallbacks from `tests/e2e/conftest.py`, which had not been touched since
```
r-marques: **💡 NIT — The loud failure depends on `os.getenv(NAME, FALLBACK)` staying exactly that — worth a sentence so nobody "improves" it**

"E2E simply starts failing" is true today only because `${{ secrets.X }}` on an unset store sets the variable to the **empty string**, which `os.getenv` returns as-is (the default applies only when the name is absent from `os.environ`), and `"".split(":")` raises in `_parse_nvm_api_key` — the 48-error signature from the pre-#278 runs. The common "fix" for empty env vars, `os.getenv(NAME) or FALLBACK`, would silently route an empty Dependabot secret onto the committed key and turn this loud failure into a green run on the wrong store. One clause here (or a comment on the two `os.getenv` calls in `conftest.py`, untouched by this PR) keeps that from happening.

If you want the failure to *name* its cause instead of reproducing the signature #278 took a diagnosis to attribute, a 3-line pre-step before `Run E2E tests` does it: `[ -n "$TEST_SUBSCRIBER_API_KEY" ] && [ -n "$TEST_BUILDER_API_KEY" ] || { echo "::error::keys empty — on a dependabot[bot] run the Dependabot secret store lost them; see the comment on this job"; exit 1; }`. It does not cover a *revoked* key (that shows as 401s in every test) — say so if you add it.

<sub>🤖 pr-review-toolkit · pr-test-analyzer, silent-failure-hunter · medium</sub>

.github/workflows/test.yaml
```diff
@@ -70,34 +70,43 @@ jobs:
...
-    # conclusion `skipped`, which satisfies the required check: the fork gate
-    # came in with #206, and the fork PRs that exercised it — #215 and #181 —
-    # both merged into protected `main` with `e2e: skipped`.
+    # Dependabot runs are NO LONGER skipped. They were, from #278 until now,
+    # because a workflow triggered by `dependabot[bot]` reads the repository's
+    # *Dependabot* secret store — a store entirely separate from the Actions one
+    # — and this repo had left it empty, so the keys arrived blank and every E2E
+    # test errored in `_parse_nvm_api_key` with "ValueError: not enough values to
+    # unpack (expected 2, got 1)". With `e2e` a required check, that made every
+    # Dependabot PR unmergeable; 18 of them piled up that way.
```
r-marques: **💡 NIT — The rewrite drops the one fact that makes the retained fork guard mergeable**

The deleted text said: "A job skipped by this `if` reports conclusion `skipped`, which satisfies the required check: the fork gate came in with #206, and the fork PRs that exercised it — #215 and #181 — both merged into protected `main` with `e2e: skipped`." The fork guard stays (107–109) and `e2e` stays required, so the question that sentence answered — *then how does a fork PR ever merge?* — is still live, and its evidence (#206, #215, #181 — all check out) is gone; only an incidental "instead of reporting `skipped`" survives at line 83. Restore one sentence next to the fork guard.

<sub>🤖 pr-review-toolkit · comment-analyzer, code-reviewer · medium</sub>

.github/workflows/test.yaml
```diff
@@ -70,34 +70,43 @@ jobs:
...
-    # exposed to `dependabot[bot]` (they would have to live in the separate
-    # Dependabot secret store, which this repo does not populate), so the keys
-    # arrive empty and every E2E test errors in `_parse_nvm_api_key` with
-    # "ValueError: not enough values to unpack (expected 2, got 1)". `e2e` is a
-    # required status check, so that turns every Dependabot PR unmergeable —
-    # which is how 18 of them piled up. A job skipped by this `if` reports
-    # conclusion `skipped`, which satisfies the required check: the fork gate
-    # came in with #206, and the fork PRs that exercised it — #215 and #181 —
-    # both merged into protected `main` with `e2e: skipped`.
+    # Dependabot runs are NO LONGER skipped. They were, from #278 until now,
```
r-marques: **💡 NIT — 34 comment lines for a 3-line `if:` change — most of it is PR-body rationale, not what the file must carry**

Lines 83–86 ("matters more here than in most repos …") and 97–102 (the security trade) are in the PR body nearly verbatim; 73–79 could be one sentence pointing at #278 / #280. What the file needs permanently is roughly ten lines: why the fork guard exists and that `skipped` satisfies the required check; which store a `dependabot[bot]`-triggered run reads, that it holds the committed fallbacks, and the `gh secret set --app dependabot` line; the #280 caveat (104–106 is fine as is). Optional — the accuracy fixes above matter more than the length.

<sub>🤖 pr-review-toolkit · comment-analyzer · low</sub>

.github/workflows/test.yaml
```diff
@@ -70,34 +70,43 @@ jobs:
...
+    #
+    # The post-merge `push` run below is still `continue-on-error` and not a
+    # required check, so a failing E2E on `main` reports the run green (measured:
+    # run 34108722286, `e2e: failure` inside a `success` run). Tracked in #280.
     if: >-
-      github.actor != 'dependabot[bot]' &&
-      (github.event_name != 'pull_request' ||
-      github.event.pull_request.head.repo.full_name == github.repository)
+      github.event_name != 'pull_request' ||
+      github.event.pull_request.head.repo.full_name == github.repository
```
r-marques, line 103: **✅ NICE — The residual expression is right for every event shape, and the failure mode it relies on is real**

Walked: push to `main` (first clause), same-repo PR and Dependabot PR (Dependabot branches are same-repo, so `head.repo.full_name == github.repository`; a `dependabot[bot]` actor reads the now-populated store), human push to a `dependabot/*` branch (human actor → Actions store), fork PR (skipped, no secrets from either store). The store holds exactly the two names the workflow reads, both created 2026-09-14T15:35Z. And the empty-key failure this rests on is confirmed from real runs, not inferred: the last Dependabot runs before #278 (2026-08-04) ended with 48 errors of `PaymentsError: Invalid NVM API Key: not enough values to unpack` — proof that an absent secret arrives as `""` and the committed fallback does not take over. The new text also states the mechanism ("reads the Dependabot secret store") more accurately than the old ("secrets are not exposed to `dependabot[bot]`").

<sub>🤖 pr-review-toolkit · code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer · high</sub>

r-marques (approved):
🟡 **Approved with comments** — no blockers; 6 should-fix, 6 nit(s) left inline.

Inline review: https://github.com/nevermined-io/payments-py/pull/284#pullrequestreview-5208095118

.github/workflows/test.yaml
```diff
@@ -65,39 +69,56 @@ jobs:
...
+    # the Dependabot store was seeded from the staging keys already committed in
+    # `tests/e2e/conftest.py`, so a Dependabot run exposes nothing a `git clone`
+    # does not. Do NOT overwrite it with the rotated Actions keys unless you
+    # accept exposing a non-public credential to dependency code on every
+    # Dependabot PR.
     #
-    # Gated on `github.actor`, the actor of the triggering event, and not on the
-    # PR author: when a human pushes to a `dependabot/*` branch to fix up a bump,
-    # the actor is that human, secrets are available, and E2E runs as normal.
+    # Reading a Dependabot run:
```
aaitor: 💡 **Nit:** The round-2 request says the prior note about carrying 34 comment lines for this three-line condition was folded in, but this explanation is now 47 lines (72–118) and repeats the operational material added to `tests/e2e/README.md`, including historical run evidence. Keeping the durable reason for the fork/Dependabot behavior here and pointing to the README / #280 would actually resolve that note; the run-specific evidence can stay in the PR discussion.

aaitor (commented):
Round 2 reviewed. The prior should-fix items are addressed and the repaired credit-burn assertion now fails normally; one remaining prior nit is inline.

aaitor (approved):
🟡 **Approved with comments** — no blockers; 0 should-fix, 1 nit(s) left inline.

Inline review: https://github.com/nevermined-io/payments-py/pull/284#pullrequestreview-5208355221

PR #286, opened 2026-09-15
fix(mcp): advertise a tier-qualified authorization_endpoint in the MCP server's OAuth metadata

### LINKED ISSUE #277: fix(mcp): the SDK MCP OAuth metadata advertises a tier-blind authorization_endpoint — add ?network=sandbox|live
## What happens

The MCP server the SDK hosts publishes its own OAuth Authorization Server metadata (`payments_py/mcp/http/oauth_metadata.py`):

```python
"authorizationUri": f"{frontend}/oauth/authorize",   # https://nevermined.app/oauth/authorize
"tokenUri":         f"{backend}/oauth/token",        # https://api.sandbox.nevermined.app/oauth/token
```

So an MCP client that discovers this server is sent to a **tier-blind** consent URL while the token endpoint is tier-specific. The Nevermined web app boots on whatever tier the user's browser last chose — **Live by default** — so a user of a `sandbox` MCP server who opens the consent link in a fresh browser profile is consented against the **live** API: the wrong `client_id` registry, the wrong wallet, and, for an account-level connector, *"Connector not authorized"*.

Twin of nevermined-io/payments#447 (the TypeScript SDK). The Nevermined API's own RFC 8414 document was fixed in nevermined-io/nvm-monorepo#3430 / #3433: each tier's `authorization_endpoint` now carries `?network=sandbox|live`, and the web app's `/oauth/*` routes honour the param (auto-switch + reload at the entry route, tier badge on the consent screens).

## Fix

Append `?network=<tier>` to `authorizationUri`, with the tier derived from the SDK environment name:

| environment | `network` |
| --- | --- |
| `sandbox`, `staging_sandbox` | `sandbox` |
| `live`, `staging_live` | `live` |
| `custom` | derive from the backend host (`api.sandbox.` / `api.live.`), else omit the param (the web app then keeps its current behaviour) |

- Param **name** is `network`, **values** `sandbox` / `live` — the embed widget's existing public param (nvm-monorepo#1787). Do not introduce `env`.
- Build with `urllib.parse` (`urlsplit` / `urlencode` / `urlunsplit`), not an f-string, so anything later appended cannot produce a second `?`.
- RFC 6749 §3.1 lets the authorization endpoint carry a query component; clients MUST retain it when adding their own parameters.
- Pin the query component per environment in the `oauth_metadata` tests; update `docs/api/` where the metadata is shown.

## Acceptance

- `GET /.well-known/oauth-authorization-server` on a `sandbox` MCP server → `authorization_endpoint` ends in `/oauth/authorize?network=sandbox`; `live` → `?network=live`.
- Same value on `/.well-known/openid-configuration` (shares the same URL set).

Found while implementing nevermined-io/nvm-monorepo#3430 (docs: nevermined-io/docs#394).

r-marques:
> ✅ **Gate cleared (2026-09-18).** This PR was held draft until nevermined-io/nvm-monorepo#3515 — the consent page tolerating the official Python MCP client's second `?` on the advertised authorize URL — reached both tiers. It shipped in nvm v1.38.1; argocd has the webapp at v1.41.0 on prod and staging, and the served bundle at `nevermined.app` / `nevermined.dev` carries the tolerant parser. Rebased on `main` (CI-only commits, no overlap); pytest (via testmcp) 1214 collected / 1188 passed / 0 failed / 26 skipped, `black --check` clean on this head. The self-review panel was posted and folded while it was a draft.

## Description

Closes #277 — the Python twin of nevermined-io/payments#455 / #447, for nevermined-io/nvm-monorepo#3430 (API + webapp, merged as `2fc26dc5c`).

The MCP server the SDK hosts publishes its own RFC 8414 / OIDC discovery documents (`build_authorization_server_metadata`, `build_oidc_configuration`, and the root `build_server_info_response`) with `authorization_endpoint: f"{frontend}/oauth/authorize"` — **tier-blind**, although the SDK knows its environment. One Nevermined web app serves the consent screens for both API tiers and boots on whatever tier the user's browser last chose (Live by default), so a `sandbox` server's users landed on the **live** consent screen — wrong `client_id` registry, wrong wallet, and for an account-level connector *"Connector not authorized"*.

### What changed (`payments_py/mcp/http/oauth_metadata.py`)

- New `resolve_oauth_tier(environment, backend_url)`: `sandbox | staging_sandbox → "sandbox"`, `live | staging_live → "live"`; `custom` is derived from its backend host (`api.sandbox.` / `api.live.`), and when that cannot be classified the tier is **omitted** — a `localhost` stack keeps the pre-existing bare URL rather than a guessed tier that would send the human to the wrong backend.
- `_build_oauth_urls` stamps `?network=<tier>` on `authorizationUri` via `urllib.parse` (`urlsplit` / `urlencode` / `urlunsplit` — one query string a client can extend with `&`, never a second `?`); token / jwks / userinfo / issuer are untouched. An explicit `authorizationUri` override passes through unchanged.
- Param name + values are the API's and the embed widget's (`network`, `sandbox|live`; nvm-monorepo#3430 / #1787). `OAUTH_TIER_PARAM`, `OAuthTier` and `resolve_oauth_tier` are exported from `payments_py.mcp.http`.
- `docs/api/09-mcp-integration.md`: one sentence on the discovery-endpoints list (`docs/reference/` is generated from docstrings — the new docstrings carry the same rationale).

No public signature changed (`_build_oauth_urls` is private; the exported builders keep their shapes) — the advertised *value* changes, which is what integrators need to know: a hand-configured client should copy the endpoint from the discovery document, query string included.

Related to Issue #277

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature
- [ ] Breaking change

## Checklist:

- [x] Follows the code style of this project (`poetry run black --check .` clean).
- [x] Tests Cover Changes
- [x] Documentation

## Test plan

- [x] `poetry run pytest -m "not slow"` → **1095 passed, 1 skipped** (this file: 41 passed, +9 new).
- [x] `tests/unit/mcp_tests/test_oauth_metadata.py::TestAuthorizationEndpointCarriesTheTier` (9): per-environment `authorization_endpoint` for all four named environments across all three documents; `network` is the param and the URL has exactly one query string; the tier is not stamped on token/jwks/userinfo/issuer; an `authorizationUri` override passes through; `resolve_oauth_tier` for named / custom-derived / unclassifiable inputs; a localhost `custom` stays bare.
- [x] Mutation: dropping the tier from `authorizationUri` → **5 failed / 36 passed**; fix restored.
- [x] `poetry run black --check .` clean; `poetry run mkdocs build` succeeds (the griffe warnings it prints are pre-existing, in files this PR does not touch).
- [ ] e2e (`tests/e2e/mcp_tests/test_mcp_oauth_server.py`) not re-run locally (needs staging credentials); it asserts `authorization_endpoint` is present, unchanged expectation.


payments_py/mcp/http/oauth_metadata.py
```diff
@@ -54,26 +138,43 @@ def _build_oauth_urls(frontend_url: str, backend_url: str) -> OAuthUrls:
...
-    env_config = Environments.get(environment, Environments["sandbox"])
-    return _build_oauth_urls(env_config.frontend, env_config.backend)
+    effective: EnvironmentName = (
+        environment if environment in Environments else "sandbox"
+    )
+    env_config = Environments[effective]
+    return _build_oauth_urls(
+        env_config.frontend,
+        env_config.backend,
+        resolve_oauth_tier(effective, backend_for_tier or env_config.backend),
```
r-marques, line 178: **🟡 SHOULD FIX — Under `custom` the tier was derived from `Environments["custom"].backend`, not the backend the document publishes**

Parity with payments#455's finding: `custom` + a `tokenUri` override pointing at `api.sandbox.…` published a sandbox `token_endpoint` next to a BARE `authorization_endpoint` — the #277 bug, silently. **Folded:** `get_oauth_urls` passes the override as the backend the tier derives from; pinned through the public `get_oauth_urls("custom", {"tokenUri": …})` and via `monkeypatch.setitem(Environments, "custom", …)` (no private reach-in any more). Mutants "tier from the wrong field" (3) and "ignore the override" (1) fail.

<sub>🤖 pr-review-toolkit · pr-test-analyzer + silent-failure-hunter · high</sub>

payments_py/mcp/http/oauth_metadata.py
```diff
@@ -34,16 +35,99 @@
...
+        return "sandbox"
+    if environment in ("live", "staging_live"):
+        return "live"
+    try:
+        # ``hostname`` is lowercased and carries no port/credentials/path. Only an
+        # unbalanced IPv6 literal (``http://[::1``) makes ``urlsplit`` raise.
+        labels = (urlsplit(backend_url).hostname or "").split(".")
+    except ValueError:
+        return None
+    if "api" in labels:
```
r-marques, line 88: **🟡 SHOULD FIX — Anchor the host classifier on the `api.<tier>` label pair**

`startswith("api.sandbox.")` missed the API's branded per-org subdomains and `mcp.api.…`, and nothing pinned that a bare `sandbox` label elsewhere is not a tier. **Folded:** labels split, tier = the label after `api`; pinned positive (branded, `mcp.api.live.…`, `API.Sandbox…:8443`) and negative (`sandbox.nevermined.app`, path-embedded). The `.hostname`-vs-`.netloc` and any-label mutants both fail.

<sub>🤖 pr-review-toolkit · code-reviewer + pr-test-analyzer · high</sub>

payments_py/mcp/http/oauth_metadata.py
```diff
@@ -34,16 +35,99 @@
...
+    port or fragment on the frontend round-trips; blank values are kept). A frontend
+    that ``urlsplit`` refuses — an unbalanced IPv6 literal in ``NVM_FRONTEND_URL`` —
+    falls back to plain concatenation, the same shape the bare URL always had for
+    that misconfiguration, matching the TypeScript SDK.
+    """
+    if tier is None:
+        return authorize_url
+    try:
+        parts = urlsplit(authorize_url)
+    except ValueError:
```
r-marques, line 108: **🟡 SHOULD FIX — A malformed `NVM_FRONTEND_URL` raised `ValueError` out of every discovery builder; the TS twin falls back**

Asymmetric posture inside one PR: a malformed BACKEND url → silent `None`, a malformed FRONTEND url (`http://[::1:3000`) → an uncaught `ValueError` from `_with_tier_param` where the pre-PR f-string never raised. **Folded:** `_with_tier_param` catches the one input `urlsplit` refuses and concatenates (matching TS), and `parse_qsl(..., keep_blank_values=True)` so a `?foo=` on a custom frontend is not silently dropped. Pinned by a 3-case parametrized test on `_with_tier_param` (plain, query+port+fragment+existing `network`, `[::1`); dropping either fallback fails 1 test.

<sub>🤖 pr-review-toolkit · silent-failure-hunter + code-reviewer + pr-test-analyzer · high</sub>

tests/e2e/mcp_tests/test_mcp_oauth_server.py
```diff
@@ -209,6 +209,12 @@ async def test_oauth_authorization_server_metadata(self):
             # Required fields per RFC 8414
             assert "issuer" in data
             assert "authorization_endpoint" in data
+            # payments-py#277: the SERVED document names the API tier — the wiring
+            # (server_manager → create_oauth_router → builders) on a sandbox mock.
+            assert (
+                data["authorization_endpoint"]
+                == "https://nevermined.app/oauth/authorize?network=sandbox"
```
r-marques, line 216: **🟡 SHOULD FIX — The only test through the production wiring asserted presence only — and it needs no credentials**

`server_manager → create_oauth_router → builders` on a `PaymentsMock` localhost server; the PR body's "needs staging credentials" did not apply to this file. **Folded:** exact `authorization_endpoint` on both the AS and OIDC documents; the file ran locally — 11 passed, 3 skipped.

payments_py/mcp/http/oauth_metadata.py
```diff
@@ -34,16 +35,99 @@
...
+    """
+    if environment in ("sandbox", "staging_sandbox"):
+        return "sandbox"
+    if environment in ("live", "staging_live"):
+        return "live"
+    try:
+        # ``hostname`` is lowercased and carries no port/credentials/path. Only an
+        # unbalanced IPv6 literal (``http://[::1``) makes ``urlsplit`` raise.
+        labels = (urlsplit(backend_url).hostname or "").split(".")
+    except ValueError:
```
r-marques, line 86: **💡 NIT — `except ValueError` was unpinned — `"not a url"` does not raise, `"http://[::1"` does**

**Folded:** `http://[::1` added to the unclassifiable list, with the comment stating the one input that actually raises.

<sub>🤖 pr-review-toolkit · pr-test-analyzer + code-reviewer · high</sub>

tests/unit/mcp_tests/test_oauth_metadata.py
```diff
@@ -364,3 +369,186 @@ def test_allows_complete_url_overrides(self):
...
+                "https://nevermined.app/oauth/authorize?network=sandbox",
+            ),
+            (
+                "https://api.live.nevermined.app/",
+                "https://nevermined.app/oauth/authorize?network=live",
+            ),
+            ("http://localhost:3001", "https://nevermined.app/oauth/authorize"),
+        ],
+    )
+    def test_custom_derives_the_tier_through_the_public_surface(
```
r-marques, line 477: **💡 NIT — The custom test imported a private inside the test body (imports-at-top rule) and bypassed the public surface**

**Folded:** `monkeypatch.setitem(Environments, "custom", EnvironmentInfo(...))` drives the public `get_oauth_urls("custom")` — the panel was right that the lookup is call-time even though the values are read at import. An unknown-environment pin (`get_oauth_urls("staging")` → sandbox-tagged fallback) is added as well.

<sub>🤖 pr-review-toolkit · pr-test-analyzer + code-reviewer · high</sub>

payments_py/mcp/http/oauth_metadata.py
```diff
@@ -34,16 +35,99 @@
...
+#: users to the LIVE consent screen, where the connector is not registered
+#: ("Connector not authorized"). The tier is stated on the URL instead; RFC 6749
+#: §3.1 obliges clients to retain the query component when they add their own
+#: parameters. Same name and values as the Nevermined API's own RFC 8414 document
+#: and the embed widget (nvm-monorepo#3430 / #1787).
+OAUTH_TIER_PARAM = "network"
+OAUTH_TIERS = ("sandbox", "live")
+OAuthTier = Literal["sandbox", "live"]
+
+
```
r-marques, line 52: **✅ NICE — Parity with the TS twin holds; classification is safe**

Outputs identical to payments#455 for all four named environments (param order aside); `backend_url` is operator config, `.hostname` lowercases and strips port/credentials; omission is the pre-PR behaviour and the webapp's documented path. `docs/reference/` excludes the MCP http module, so no generated-doc impact; `payments_py.mcp` never re-exported `get_oauth_urls`, so exporting from `payments_py.mcp.http` is consistent.

r-marques (commented):
## 🤖 Automated PR review — 🟡 Mergeable with nits (all findings folded)

The Python twin of payments#455 for nvm-monorepo#3430: the MCP server's own discovery documents now name the API tier on `authorization_endpoint`. Same findings as the TS twin, folded the same way — the `custom` tier follows the backend the document actually publishes, the host classifier is anchored on the `api.<tier>` label pair, and `_with_tier_param` tolerates the one frontend `urlsplit` refuses instead of raising out of every builder — plus the localhost e2e now pins the exact served endpoint (it runs on a mock; 11 passed).

⚠️ **Do not ship this before nvm-monorepo#3515 is on both tiers.** The official Python MCP client concatenates a second `?` onto the authorize URL (`mcp` 1.30 → 2.2.0, `client/auth/oauth2.py`); the production webapp is still the strict #3433 parser, so a tier-qualified SDK document would turn Python-MCP clients away until #3515 deploys. This PR stays draft until then.

**What I reviewed**
- Diff range `origin/main...HEAD` — 5 files, head `eeacdc7c`.
- Panel: code-reviewer, silent-failure-hunter, pr-test-analyzer (the code-reviewer surfaced the Python-MCP-client incompatibility that led to #3515). Re-verified after folding: `poetry run pytest -m "not slow"` → 1102 passed; the e2e file → 11 passed / 3 skipped; black clean; 6 mutants killed.

**Verdict:** 🟡 Mergeable with nits — 0 blockers, 4 should-fix, 2 nits, 1 praise — all folded in `eeacdc7`.

### 🧭 Inline comments (7)

- 🟡 **SHOULD FIX** — Under `custom` the tier was derived from `Environments["custom"].backend`, not the backend the document publishes — `payments_py/mcp/http/oauth_metadata.py:176`
- 🟡 **SHOULD FIX** — Anchor the host classifier on the `api.<tier>` label pair — `payments_py/mcp/http/oauth_metadata.py:86`
- 🟡 **SHOULD FIX** — A malformed `NVM_FRONTEND_URL` raised `ValueError` out of every discovery builder; the TS twin falls back — `payments_py/mcp/http/oauth_metadata.py:106`
- 🟡 **SHOULD FIX** — The only test through the production wiring asserted presence only — and it needs no credentials — `tests/e2e/mcp_tests/test_mcp_oauth_server.py:216`
- 💡 **NIT** — `except ValueError` was unpinned — `"not a url"` does not raise, `"http://[::1"` does — `payments_py/mcp/http/oauth_metadata.py:84`
- 💡 **NIT** — The custom test imported a private inside the test body (imports-at-top rule) and bypassed the public surface — `tests/unit/mcp_tests/test_oauth_metadata.py:477`
- ✅ **NICE** — Parity with the TS twin holds; classification is safe — `payments_py/mcp/http/oauth_metadata.py:52`

eruizgar91:
:eyes: Reviewing

payments_py/mcp/http/oauth_metadata.py
```diff
@@ -34,16 +35,99 @@
...
+    The four named environments map directly. ``custom`` is derived from the host
+    of the backend it will publish as ``token_endpoint``: a Nevermined API host has
+    an ``api`` label immediately followed by the tier label —
+    ``api.sandbox.nevermined.app``, ``<slug>.api.live.nevermined.app`` (branded
+    per-org subdomains), ``mcp.api.sandbox.nevermined.dev`` — so that label pair
+    is what is matched, never a bare ``sandbox`` anywhere in the host. When the
+    host cannot be classified (a ``localhost`` stack, a proxy/CNAME in front of the
+    API) the tier is **omitted**, not guessed: the URL stays the bare one, and the
+    operator sets ``oauth_urls={"authorizationUri": "...?network=<tier>"}`` to say
```
eruizgar91: **MEDIUM** — the escape hatch this docstring prescribes names a key that does not exist.

There is no `oauth_urls` option anywhere in this SDK. The option is `oauthUrls` — `OAuthConfig.oauthUrls`, and the same field on `HttpRouterConfig` / `HttpServerConfig` ([`http_types.py#L96-L99`](https://github.com/nevermined-io/payments-py/blob/7b5f157a5658daa93b535b05b4d8662c9209e12b/payments_py/mcp/types/http_types.py#L96-L99)) — read as `options.get("oauthUrls")` in [`oauth_router.py#L113-L116`](https://github.com/nevermined-io/payments-py/blob/7b5f157a5658daa93b535b05b4d8662c9209e12b/payments_py/mcp/http/oauth_router.py#L113-L116). All three are `total=False` TypedDicts consumed with `.get`, so a dict carrying `oauth_urls` is dropped in silence: no error, no warning, and the document keeps the bare URL — the exact state the operator was following this sentence to leave. The new bullet in `docs/api/09-mcp-integration.md:413` repeats the same spelling, on a page whose own options table is camelCase throughout (`serverName`, `baseUrl`).

This is one of only two places the twins tell the operator different things: payments#455's paragraph names `oauthUrls.authorizationUri`, which is correct for that SDK.

Rename to `oauthUrls` here and at `docs/api/09-mcp-integration.md:413`.

docs/api/09-mcp-integration.md
```diff
@@ -410,7 +410,7 @@ For backward compatibility the server still falls back to reading the access tok
 
 The MCP server exposes:
 
-- `/.well-known/oauth-authorization-server` - OAuth 2.1 discovery
+- `/.well-known/oauth-authorization-server` - OAuth 2.1 discovery. Its `authorization_endpoint` names the API tier your server runs against — `https://nevermined.app/oauth/authorize?network=sandbox` for `sandbox`, `?network=live` for `live` — because one Nevermined web app serves the consent screens for both tiers and boots on whatever tier the user's browser last chose. Clients keep that query string when they add their own parameters (RFC 6749 §3.1); if you configure a client by hand, copy the endpoint from the discovery document, query string included. A `custom` environment derives the tier from its backend host (`api.sandbox.…` / `api.live.…`, branded subdomains included); behind a host the SDK cannot classify — `localhost`, a proxy — no tier is stamped, so set `oauth_urls={"authorizationUri": "<webapp>/oauth/authorize?network=<tier>"}` yourself.
```
eruizgar91: **MEDIUM** — spelled correctly, the override still cannot be supplied on the path this page documents, so the fallback this sentence promises is unreachable.

`ServerManager.start()` builds the `create_oauth_router({...})` options dict by hand and never forwards `oauthUrls` ([`server_manager.py#L407-L432`](https://github.com/nevermined-io/payments-py/blob/7b5f157a5658daa93b535b05b4d8662c9209e12b/payments_py/mcp/core/server_manager.py#L407-L432)); `McpServerConfig` — the config `payments.mcp.start()` accepts — has no such field at all ([`server_types.py#L173-L190`](https://github.com/nevermined-io/payments-py/blob/7b5f157a5658daa93b535b05b4d8662c9209e12b/payments_py/mcp/types/server_types.py#L173-L190)); and the decorator API is narrower still, `PaymentsMCP.start()` taking four explicit keyword arguments ([`payments_mcp.py#L524-L532`](https://github.com/nevermined-io/payments-py/blob/7b5f157a5658daa93b535b05b4d8662c9209e12b/payments_py/mcp/payments_mcp.py#L524-L532)). The override therefore only works on the advanced `create_oauth_router(...)` / `build_*_metadata(...)` path, not on either managed entry point.

That is precisely the population this sentence addresses. A `custom` deployment behind a host `resolve_oauth_tier` cannot classify gets the bare URL by design, and then has no supported way to state its tier — so it keeps landing on whatever tier the browser last chose and fails as #277 described, with nothing in "Connector not authorized" naming the tier. Omitting rather than guessing is the right call, but its safety argument rests on this knob existing.

The TS twin does forward it: `startManagedServer` destructures `oauthUrls` and passes it to `createOAuthRouter` ([`managed-server.ts#L74-L76`](https://github.com/nevermined-io/payments/blob/55a982bdff2c1ee5be7adaa7c9ea245f2f72147f/src/mcp/http/managed-server.ts#L74-L76), and again for `createMcpApp` at [`#L238-L265`](https://github.com/nevermined-io/payments/blob/55a982bdff2c1ee5be7adaa7c9ea245f2f72147f/src/mcp/http/managed-server.ts#L238-L265)).

Two lines in one adjacent file: add `oauthUrls: Optional[Dict[str, str]]` to `McpServerConfig`, and `"oauthUrls": config.get("oauthUrls"),` to the dict at `server_manager.py:409`.

payments_py/mcp/http/oauth_metadata.py
```diff
@@ -54,26 +138,43 @@ def _build_oauth_urls(frontend_url: str, backend_url: str) -> OAuthUrls:
 
     return {
         "issuer": frontend,
-        "authorizationUri": f"{frontend}/oauth/authorize",
+        "authorizationUri": _with_tier_param(f"{frontend}/oauth/authorize", tier),
```
eruizgar91, line 143: **LOW** (follow-up — cannot land here) — `issuer`, one line above, is still tier-blind.

`Environments["sandbox"].frontend` and `Environments["live"].frontend` are the same string, `https://nevermined.app` ([`environments.py#L36-L61`](https://github.com/nevermined-io/payments-py/blob/7b5f157a5658daa93b535b05b4d8662c9209e12b/payments_py/environments.py#L36-L61)). After this change the two tiers' RFC 8414 documents differ in `authorization_endpoint`, `token_endpoint`, `jwks_uri` and `userinfo_endpoint` but publish an **identical** `issuer` — the field RFC 8414 §2 designates as the authorization server's identifier, and the one an OAuth client keys AS identity and its token cache on. Same root cause as #277: one webapp origin standing in for two authorization servers.

Not introduced here, and not fixable here: `issuer` is a published value that would have to move together with the Nevermined API's own document and the TypeScript SDK, which is a design decision this PR has not made. Flagging so it gets an issue (with a TS twin) rather than being lost — it is the last field in this dict that still cannot tell the tiers apart.

eruizgar91 (commented):
**Bot-mode review — no verdict here**; `act.py` issues it. Reviewed at `7b5f157a`, side by side with the TypeScript twin nevermined-io/payments#455 at `bef9839e`. Three findings, all inline: **2 MEDIUM, 1 LOW** (the LOW is explicitly deferred, see below). Nothing overlaps the author's own commented panel.

### Do the two twins actually agree?

**On the part that decides which backend a human reaches — yes, and I verified it rather than reading the two PR bodies.** Both derive the tier identically: named environments short-circuit; `custom` is matched on the `api.<tier>` **label pair** taken from the parsed hostname (first `api` label, never a bare `sandbox` anywhere in the host); an unclassifiable host **omits** the tier rather than guessing; and both take the tier from the backend the document will actually publish, i.e. an `oauthUrls.tokenUri` override. Driving the Python public surface gives byte-for-byte the strings the TS suite pins:

```
sandbox          https://nevermined.app/oauth/authorize?network=sandbox
live             https://nevermined.app/oauth/authorize?network=live
staging_sandbox  https://nevermined.dev/oauth/authorize?network=sandbox
staging_live     https://nevermined.dev/oauth/authorize?network=live
custom/localhost http://localhost:3000/oauth/authorize          <- bare, as intended
unknown env      ...?network=sandbox beside a sandbox token_endpoint
```

I went looking specifically for the dangerous asymmetry — a host one side classifies **wrong** where the other omits — and there is none. Across a 28-shape host matrix run through both implementations, exactly two rows diverge, and both are malformed URLs on which the two parsers differ: a scheme-less protocol-relative `//api.sandbox.…` (`urlsplit` finds a hostname → Python derives `sandbox`; `new URL` throws → TS falls back to bare) and a triple-slash `https:///api.sandbox.…` (WHATWG collapses the extra slash → TS derives `sandbox`; `urlsplit` takes it literally, empty netloc → Python falls back to bare). In both rows the non-omitting side produces the **correct** tier, so each divergence is in the safe direction — neither side ever names the opposite tier. Neither suite covers either shape.

**Where they diverge and it matters is the escape hatch they both document for the unclassifiable case** — findings 1 and 2. TS names `oauthUrls.authorizationUri`, a key its managed server forwards. Python names `oauth_urls=…`, which exists nowhere, and the key that does exist is accepted by neither Python managed-server entry point. So the bare-URL fallback is operator-recoverable in TS and not in Python.

### Verified, not taken on trust

- `tests/unit/mcp_tests/test_oauth_metadata.py` — **48 passed** on this head. Two mutants killed: `_with_tier_param` returning the URL unchanged → 12 failed; replacing the `api.<tier>` anchor with "any tier label anywhere in the host" → 1 failed. Both halves have teeth.
- `black --check` clean on all four touched Python files.
- CI on this head: `unit_integration` green (1111 passed / 1 skipped), `lint` green. The red `e2e` check is `test_organizations_e2e.py::test_per_call_organization_id_override_targets_org_without_mutating_pin` — unrelated; every `test_mcp_oauth_server.py` case in that same run passed, including the two new exact-value assertions.
- **RFC shape is fine.** `authorization_endpoint` may carry a query: RFC 8414 §2 defines it by reference to RFC 6749 §3.1, which explicitly permits an `application/x-www-form-urlencoded` query component that clients MUST retain when adding their own; OIDC Discovery inherits that. RFC 9728 carries no `authorization_endpoint` at all (only `authorization_servers`), and this PR does not touch that document.
- **No downstream re-parse hazard in this SDK.** The three builders read `oauth_urls["authorizationUri"]` verbatim, and nothing in `payments_py/mcp/` constructs a redirect or re-joins that URL. The one consumer that must tolerate a concatenating client is the webapp's consent page, which is the nvm-monorepo#3515 gate the PR body cites.

### Considered and discarded

- **Query re-encoding and parameter order vs `URLSearchParams`.** Real but degenerate: when the authorize URL already carries a query, `parse_qsl` → `urlencode` diverges from TS three ways — an existing `network` moves to the **end** (TS keeps its position), `~` survives as `~` (TS emits `%7E`), and TS additionally lowercases the host via `URL.toString()`. Only reachable through an `NVM_FRONTEND_URL` that carries a query; no `Environments` frontend does. Worth recording that Python's append-last is the *safer* of the two under the consent page's tolerant parser, which recovers exactly one swallowed `key=value` and only from the last parameter. The author already noted this as "param order aside".
- **The `except ValueError` concat fallback placing the param after a fragment** (`…/authorize#x` → `#x?network=…`): identical in both twins and unreachable outside a malformed IPv6 `NVM_FRONTEND_URL`.
- **A named environment with a cross-tier `tokenUri` override** (`environment="live"` + a sandbox `tokenUri`) publishes a sandbox `token_endpoint` beside `?network=live`. The named-env short-circuit is deliberate and identical in TS; that is the operator's own contradiction, not the SDK's.
- **The Python e2e hardcoding `https://nevermined.app/oauth/authorize?network=sandbox`** where the TS e2e derives it from `TEST_ENVIRONMENT`: the Python fixture pins `PaymentsMock._environment_name = "sandbox"`, so the exact value is stable and is the stronger of the two assertions.
- **Export surface.** `payments_py.mcp` never re-exported `get_oauth_urls`, so exporting the new symbols from `payments_py.mcp.http` alone — where TS exports from both barrels — follows this SDK's own convention rather than missing a step.

**Deferred, needs an issue:** the tier-blind `issuer` (finding 3), with the reason it cannot land in this PR stated inline. Nothing else is deferred — findings 1 and 2 are each a couple of lines and belong here.

eruizgar91 (approved):
🟡 **Approved with comments** — no blockers; 2 should-fix, 1 nit(s) left inline.

Inline review: https://github.com/nevermined-io/payments-py/pull/286#pullrequestreview-5247891485

payments_py/mcp/http/oauth_metadata.py
r-marques: Folded in `b5df31a`: the docstring and `docs/api/09-mcp-integration.md` now name `oauthUrls` — the camelCase key `McpServerConfig` / `HttpRouterConfig` read — and say where it goes (`payments.mcp.start()` or `create_oauth_router()`). The Server Configuration table gained an `oauthUrls` row so the option is discoverable from the page's own reference, not only from this sentence.

docs/api/09-mcp-integration.md
r-marques: Folded in `b5df31a`, your two lines plus a test: `McpServerConfig.oauthUrls: Optional[Dict[str, str]]` (documented in the Args) and `"oauthUrls": config.get("oauthUrls")` in the hand-built router options. `tests/unit/mcp_tests/test_server_manager_oauth_urls.py` drives the real `start()` up to the router — `create_oauth_router` is bound in the module namespace, so a spy records the options and raises a sentinel, no port bound — and pins both that the override arrives verbatim and that absent stays `None` (a default here would be a second place inventing an authorize URL). Mutant: dropping the forward line fails both tests. The decorator `PaymentsMCP.start()` (four explicit kwargs) is untouched — widening that signature is a separate public-API question you did not raise.

payments_py/mcp/http/oauth_metadata.py
r-marques: Agreed, and it is worse than an identity nit since 2026-09-15: nvm-monorepo#3532 (deployed in v1.41.0) has the webapp return RFC 9207 `iss` = the **API origin**, and the API's own document publishes `issuer` = that same origin per tier — so a client that discovers via this server's own well-known (issuer = `https://nevermined.app`) now rejects every authorization response on the simple-string compare. RFC 9728 clients follow `authorization_servers` to the API's document and are fine.

Decision made (Rodolphe, 2026-09-21): **`issuer` = the API origin per tier**, the origin `authorization_servers` already carries. Filed with that decision, in lockstep: nevermined-io/payments-py#291 (this SDK) + nevermined-io/payments#463 (TS twin). Not in this PR, as you said — a changed published identifier ships as its own minor with the changelog naming it. Resolving.

PR #287, opened 2026-09-16
ci: sweep main into open Dependabot branches, as nevermined-io/payments does

eruizgar91:
Ports `dependabot-update-branches.yml` from `nevermined-io/payments` so the two repositories' Dependabot pipelines behave the same, and turns on `strict: true` on `main` so they behave the same for the same reason. On every push to `main` — plus a Monday 09:00 UTC fallback and `workflow_dispatch` — the workflow merges `main` into up to `MAX_MERGES` open Dependabot branches through `POST /repos/{owner}/{repo}/merges`.

## What this is for

`main` now requires branches to be up to date before merging, so a Dependabot PR that has fallen behind **cannot merge** until something merges `main` into it. Unattended, that something is this workflow. It is load-bearing rather than a convenience: if it breaks, the queue stalls silently — nothing merges, no check goes red, the PRs just sit. That is the failure mode the `payments` copy actually had (`--method POST` on a PUT endpoint, 404 on all 21 open PRs, green job — nevermined-io/payments#449), and it is why the logic is tested and why a failed `gh pr list` fails the job.

**The first draft of this PR got this wrong**, and review caught it. It ran against `strict: false` and claimed to deliver "the safety property `strict: true` would enforce". Under `strict: false` the sweep gated nothing: no required check depended on it, `dependabot-auto-merge.yml` arms a patch/minor PR the moment it opens so one green on its original head merged before any push woke the sweep, and `MAX_MERGES` left a wave's overflow unswept and still mergeable. `strict` was turned on rather than the claim softened. Every comment that carried the old framing now records that it was wrong, so the next reader does not re-derive it.

## Branch protection change

Applied with `PATCH /repos/nevermined-io/payments-py/branches/main/protection/required_status_checks` so no other setting was touched. Diffing the full protection object before and after, `required_status_checks.strict: False -> True` is the only field that moved.

Full diff against `payments` (snapshot 2026-09-16) — `strict` was the only setting that differed:

| | `payments` | `payments-py` |
|---|---|---|
| `strict` | `true` | `false` → **`true`** |
| `required_approving_review_count` | 1 | 1 |
| `dismiss_stale_reviews` | `false` | `false` |
| `require_code_owner_reviews` | `false` | `false` |
| `require_last_push_approval` | `false` | `false` |
| `bypass_pull_request_allowances` | `null` | `null` |
| `enforce_admins` | `true` | `true` |
| `required_linear_history` / `allow_force_pushes` / `allow_deletions` / `block_creations` / `required_conversation_resolution` / `lock_branch` | `false` | `false` |
| `restrictions` | none | none |

Required check contexts differ by name only (`e2e` / `lint` / `unit_integration` vs `E2E` / `Lint & Build` / `Unit & Integration`) — job naming, not policy.

One repository setting is left differing on purpose: `allow_update_branch` is `true` here and `false` on `payments`. It only surfaces GitHub's manual "Update branch" button on a behind PR; it performs no update, so it neither replaces this workflow nor weakens anything.

No collateral damage from the flip: both open PRs are `behind_by: 0`, so nothing became newly blocked. #285 is held by a failing `e2e` from an earlier run, #287 by the review — neither caused by this change.

## Deltas from the `payments` copy

All marked `payments-py:` in the files.

| | `payments` | here |
|---|---|---|
| `MAX_MERGES` | 5 | **3** — `test.yaml` sets no `concurrency` group and `e2e` runs the real staging suite on one shared pair of accounts, so the cap is also the number of concurrent staging suites this job can start. It bounds drain rate, not whether a PR can slip through unswept |
| Logic location | inline in the workflow | **`.github/scripts/update-dependabot-branches.sh`**, so it can be tested |
| `gh pr list` failure | unguarded | **fails the job** — see below |
| Approved-first ordering | `gh pr list --jq` | **`jq` inside the script**, so a test can exercise it |
| Shell | no `set` | `set -uo pipefail`, and `[ -z "$pr" ] && continue` rewritten as an `if` so the guard cannot become an exit under a future `set -e` |
| DST arithmetic in the cron comment | "two hours in summer and one in winter" | **three and two** — 08:00 Madrid is 06:00 UTC under CEST and 07:00 UTC under CET. The `payments` copy still carries the slip; worth a one-liner there |
| Conflict comment | `pnpm-lock.yaml` | `poetry.lock` |

`.github/dependabot.yml` now pins both ecosystems to Monday 08:00 Europe/Madrid, matching `payments`, so the cron fallback has a known run to clear.

## Failure that was silently green

`set -uo pipefail` does not stop a bare `prs=$(gh pr list ...)` from yielding `""` on error. Reproduced with a stubbed 401 before fixing:

```
Looking for open Dependabot PRs...
gh: HTTP 401: Bad credentials
No open Dependabot PRs found.
exit=0
```

A green job reporting a successful sweep of nothing — the same shape as the `--method POST` bug above. Both the `gh pr list` call and the `jq` pass now fail the job explicitly, with stderr captured into the annotation.

## Tests

`tests/unit/test_dependabot_sweeper.py` runs the real script with a stubbed `gh` on PATH. It lands in `unit_integration`, a required check. The workflow triggers on a push to `main`, so it cannot be exercised before it merges — this suite is the only pre-merge cover the selection, ordering, status handling and cap get.

Verified by mutation rather than by the suite being green. Five mutants, five kills, each failing its intended test and no other:

| mutant | test that fails |
|---|---|
| drop the approved-first partition | `test_approved_prs_are_served_first` |
| count a 204 as a merge | `test_already_current_branch_does_not_consume_a_merge_slot` |
| drop the conflict case | `test_conflict_warns_and_the_sweep_continues` |
| drop the `MAX_MERGES` cap | `test_merge_cap_stops_the_wave` |
| restore the bare `prs_json=$(...)` | `test_list_failure_fails_the_job` |

## Checkout

The workflow now checks out the repository to reach the script. Called out in a comment because `dependabot-auto-merge.yml` deliberately never checks out: that one runs on `pull_request` with the App token in scope, so a checkout would sit untrusted Dependabot branch content next to a privileged credential. Every trigger on this job — push to `main`, `schedule`, `workflow_dispatch` — resolves to the default branch, so no Dependabot branch is ever checked out and nothing from one is executed; those branches are only ever named in an API call.

## After merge

Dispatch the workflow by hand once. Its `push` trigger only fires post-merge, so this is the first run that exercises the App identity against `POST /merges` in this repository.


.github/workflows/dependabot-update-branches.yml
```diff
@@ -0,0 +1,220 @@
...
+#
+# On `payments`, `main` sets `strict: true`, so a branch that has fallen behind
+# literally cannot merge and this job is what unblocks it. Here `main` sets
+# `strict: false` (verified 2026-09-16, `gh api
+# repos/nevermined-io/payments-py/branches/main/protection`), so nothing is
+# blocked — which is exactly the problem. `dependabot-auto-merge.yml` approves
+# and queues patch/minor bumps unattended, so without this job a bump merges
+# green having been tested only against whatever `main` looked like when
+# Dependabot opened it. The required checks (`lint`, `unit_integration`, `e2e`)
+# re-run on the merge commit this job pushes, which is the safety property
```
aaitor: **blocker**: The asynchronous sweep does not enforce the strict-check property claimed here. main has strict: false, so an already-armed Dependabot PR can merge as soon as its old-head checks pass; neither the push trigger nor the weekly cron makes this job a prerequisite, and MAX_MERGES=3 can leave branches unswept while they remain mergeable. The core stale-branch race therefore still exists. Gate merging on an up-to-date branch (or on this update completing) before treating this as equivalent to strict checks.

.github/workflows/dependabot-update-branches.yml
```diff
@@ -0,0 +1,220 @@
...
+          # Measured against the real API on `payments` on 2026-09-14:
+          #   201 Created  — #356, commit 8dbd041 carrying the marker verbatim
+          #   204 No Content — same call repeated, branch already current
+          #   409 Merge conflict — #39, which is DIRTY
+          echo "Looking for open Dependabot PRs..."
+          # Approved first, least-recently-updated within each half. A slot spent
+          # on a PR nobody has approved buys nothing: it cannot merge, so it
+          # cannot produce the push that starts the next wave. Serving the
+          # approved ones first means every wave that has one ends in a merge.
+          prs=$(gh pr list \
```
aaitor: **should-fix**: set -uo pipefail does not stop when gh pr list fails. An API/auth failure in this command substitution leaves prs empty, after which the job prints No open Dependabot PRs found and exits 0. That silently reports a successful sweep when nothing was enumerated; propagate the list failure explicitly.

aaitor (commented):
Inline findings from the fallback review.

.github/workflows/dependabot-update-branches.yml
aaitor: **should-fix**: The selection and status-handling logic is only covered by a manually extracted script described in the PR body. No checked-in test would fail if the approved-first ordering, 204 handling, conflict continuation, or merge cap were deleted. Because this workflow writes Dependabot branches and cannot run pre-merge, keep the stub harness as an executable regression test (or move the logic into a tested script).

.github/workflows/dependabot-update-branches.yml
```diff
@@ -0,0 +1,220 @@
...
+  # A push is the normal trigger, and it chains: each merge is itself a push,
+  # so one wave feeds the next. That chain only holds while the PRs being served
+  # actually merge — a wave spent entirely on majors awaiting human review ends
+  # it, and nothing wakes the job again until someone lands something.
+  #
+  # Monday 09:00 UTC is the fallback for exactly that. Cron has no timezone, so
+  # it has to clear Dependabot's own weekly run (Monday 08:00 Europe/Madrid, see
+  # dependabot.yml) in both halves of the year: that is 06:00 UTC under CEST and
+  # 07:00 UTC under CET, so an hour's margin in summer is none at all in winter.
+  # 09:00 UTC leaves two hours in summer and one in winter, either side of the
```
aaitor: **should-fix**: The schedule arithmetic contradicts the times immediately above it: 08:00 Europe/Madrid is 06:00 UTC in CEST and 07:00 UTC in CET, so a 09:00 UTC cron leaves three hours in summer and two in winter, not two and one. Correct the comment because this timing is the stated basis for the fallback schedule.

aaitor (changes requested):
🔴 **Blocked** — 1 blocker(s) to address before merge. Details are inline on the changed lines.

Inline review: https://github.com/nevermined-io/payments-py/pull/287#pullrequestreview-5220744671

.github/workflows/dependabot-update-branches.yml
eruizgar91: Correct, and the claim is withdrawn — 5df5f9f.

The mechanism is as you describe: there is no required check for this job, `dependabot-auto-merge.yml` arms a patch/minor PR the moment it opens, so a PR whose checks go green on its original head merges before any push to `main` wakes the sweep, and `MAX_MERGES` leaves a wave's overflow unswept while still mergeable. Nothing in a workflow can gate a merge on an up-to-date branch; only branch protection can.

The header now states what the job does buy — re-testing the PRs that *sit* (held for review, conflicted, queued behind others) against a current `main`, which is the backlog case this repo actually hits — and states outright that it enforces nothing, with the three bypass paths enumerated. The same correction is in `dependabot-auto-merge.yml`'s header (it is the job that arms the PR, so it is where the race is visible) and in CLAUDE.md, which now says not to describe this workflow as equivalent to strict checks.

On the real fix: `strict: true` on `main` is one settings change away and it is with Enrique to call, since it makes every PR in the repo — human ones included — require an up-to-date branch before merging. The sweeper is what makes that flip cheap rather than a pile of stale branches to unblock by hand, so the order is this PR first either way. Happy to do it as a follow-up the moment he says go, at which point this paragraph gets replaced by the `payments` one.

.github/workflows/dependabot-update-branches.yml
eruizgar91: Fixed in 5df5f9f, and it reproduced exactly as you said before the fix — stubbed `gh pr list` returning `gh: HTTP 401: Bad credentials`, exit 1:

```
Looking for open Dependabot PRs...
gh: HTTP 401: Bad credentials
No open Dependabot PRs found.
exit=0
```

Green job, nothing enumerated. That is the same shape as the `--method POST` bug documented in the script header (404 on all 21 PRs under a green job), which is what made it worth a guard rather than a style note — so the comment on the guard says so, to stop it being "simplified" back.

`gh pr list` and the `jq` pass are both now `if ! ... then ::error; exit 1; fi` with stderr captured to a temp file so the message reaches the annotation. `test_list_failure_fails_the_job` holds it; restoring the bare `prs_json=$(...)` fails that test and only that test (`assert 0 == 1`).

eruizgar91: Done in 5df5f9f — the logic moved to `.github/scripts/update-dependabot-branches.sh` and `tests/unit/test_dependabot_sweeper.py` drives it with a stubbed `gh` on PATH. It runs in `unit_integration`, which is a required check, so it is real cover rather than a harness in a PR body.

One thing your comment made me change beyond keeping the harness: approved-first ordering was in `gh pr list --jq`, and through `--jq` the expression is just an argument handed to the stub — the stub's own output decides the order, so a test would have asserted nothing. The script now takes `--json` and pipes through `jq` itself, so the partition is code the test exercises.

Verified by mutation rather than by the suite being green — five mutants, five kills, each failing its intended test and no other:

| mutant | test that fails |
|---|---|
| drop the approved-first partition | `test_approved_prs_are_served_first` |
| count a 204 as a merge | `test_already_current_branch_does_not_consume_a_merge_slot` |
| drop the conflict case | `test_conflict_warns_and_the_sweep_continues` |
| drop the `MAX_MERGES` cap | `test_merge_cap_stops_the_wave` |
| restore the bare `prs_json=$(...)` | `test_list_failure_fails_the_job` |

The workflow now needs `actions/checkout` to reach the script. Called out in a comment because `dependabot-auto-merge.yml` deliberately never checks out: that one runs on `pull_request` with the App token in scope, so a checkout would sit untrusted Dependabot branch content next to a privileged credential. Every trigger on this job — push to `main`, `schedule`, `workflow_dispatch` — resolves to the default branch, so no Dependabot branch is ever checked out and nothing from one is executed; those branches are only ever named in an API call.

.github/workflows/dependabot-update-branches.yml
eruizgar91: Right, and it was arithmetic, not a typo in one direction: 08:00 Europe/Madrid is 06:00 UTC under CEST and 07:00 UTC under CET, so 09:00 UTC leaves **three** hours in summer and **two** in winter. Checked with `zoneinfo` against real 2026 dates rather than by re-reading it. Corrected in 5df5f9f.

The conclusion is unchanged and the margin is wider than was claimed, so the schedule stays at `0 9 * * 1`. The slip came in verbatim from the `payments` copy, which still carries it — noted in the comment so the next person diffing the two files does not "fix" ours back to match. Worth a one-line follow-up over there.

.github/workflows/dependabot-update-branches.yml
```diff
@@ -0,0 +1,139 @@
...
+      # Checking out is safe HERE, and the distinction is worth stating because
+      # `dependabot-auto-merge.yml` makes a point of never doing it. That file
+      # runs on `pull_request` with the App token in scope, so a checkout would
+      # put an untrusted Dependabot branch's code next to a privileged
+      # credential. This job's triggers — push to `main`, `schedule`,
+      # `workflow_dispatch` — all resolve to the default branch, so what lands
+      # in the working tree is `main` and nothing else. No Dependabot branch is
+      # ever checked out, and nothing from one is executed: the branches this
+      # job writes to are only ever named in an API call.
+      - uses: actions/checkout@v7
```
aaitor, line 136: **should-fix:** `workflow_dispatch` does not always resolve to the default branch. GitHub lets the dispatcher select any branch/tag (`gh workflow run ... --ref BRANCH`), and `actions/checkout` without an explicit `ref` checks out that dispatch ref. Consequently this can execute the selected branch's copy of `.github/scripts/update-dependabot-branches.sh` with the release-App token in `GH_TOKEN`, contradicting lines 110–114 and the security argument for adding checkout. Please pin checkout (and therefore the script) to the default branch, or remove/manual-dispatch harden this path.

.github/scripts/update-dependabot-branches.sh
```diff
@@ -0,0 +1,167 @@
...
+# push that starts the next wave. Serving the approved ones first means every
+# wave that has one ends in a merge.
+#
+# Applied here rather than passed to `gh --jq` so that a test can exercise it:
+# through `--jq` the ordering is an argument to the stub and the stub's own
+# output decides the order, so deleting the expression would not fail anything.
+if ! prs=$(printf '%s' "$prs_json" | jq -r '
+      [.[] | select(.reviewDecision == "APPROVED")]
+      + [.[] | select(.reviewDecision != "APPROVED")]
+      | .[] | "\(.number) \(.headRefName)"' 2>"$err_file"); then
```
aaitor, line 122: **should-fix:** This new parse-failure branch is not pinned by the new suite. `GH_STUB` always emits `json.dumps(prs)`, and the only failure test covers `gh pr list`; deleting this `if`/`exit 1` guard would therefore leave every test green while malformed/truncated JSON again becomes `No open Dependabot PRs found.` with exit 0. Add a bad-JSON/`jq`-failure case so this error path cannot silently regress.

aaitor (commented):
Round 2 review: the previous findings are fixed; two new should-fix findings are inline.

.github/workflows/dependabot-update-branches.yml
eruizgar91: Follow-up: Enrique said go, so `strict: true` is now **on** for `payments-py` `main` — 0cf7ce1. The blocker is closed at the root rather than by wording.

Applied with `PATCH .../protection/required_status_checks` so nothing else moved; diffing the full protection object before and after, `required_status_checks.strict: False -> True` is the only field that changed. Diffed against `payments` too, and `strict` was the only setting that differed at all — reviews, `enforce_admins`, `dismiss_stale_reviews`, `require_last_push_approval`, force-push/deletion/linear-history flags and `restrictions` were already identical. Contexts differ by name only (`e2e`/`lint`/`unit_integration` vs `E2E`/`Lint & Build`/`Unit & Integration`).

So the sweeper is now load-bearing rather than advisory, and the comments say that instead of the old claim: a silent failure stalls the queue without turning anything red, and `MAX_MERGES` bounds how fast it drains rather than whether a PR can slip through unswept. Each of the four places that carried the wrong framing — workflow header, script header, `dependabot-auto-merge.yml` header and its `--auto` step, CLAUDE.md — now records that the first draft got it wrong, so nobody re-derives it.

Two things I checked rather than assumed:

- **No collateral damage.** Both open PRs are `behind_by: 0`, so the flip blocked nothing new. #285 is held by a failing `e2e` from an earlier run and this one by your review; neither is from the change.
- **`allow_update_branch`** is `true` here and `false` on `payments` — the one repo setting still differing, left alone on purpose. It only surfaces GitHub's manual "Update branch" button on a behind PR; it performs no update, so it neither replaces this job nor weakens it. Disabling it for symmetry would lose a convenience and buy nothing. Say if you would rather have them identical anyway.

.github/workflows/dependabot-update-branches.yml
eruizgar91: Right, and it is the same class of error as the round-1 blocker: a comment asserting a property the mechanism does not have. Fixed in 21a9782.

Checked the path you describe rather than taking it on the wording — `gh workflow run ... --ref dependabot/pip/whatever` is dispatchable because a Dependabot branch carries this workflow once it has been branched from `main`, and `actions/checkout` with no `ref` would have put that branch's copy of the script in the tree with the release-App token in `GH_TOKEN`.

```yaml
- uses: actions/checkout@v7
  with:
    ref: ${{ github.event.repository.default_branch }}
    persist-credentials: false
```

Same expression as `BASE` on purpose, so the two cannot drift; if it ever resolved empty the script's `: "${BASE:?}"` guard fails the job rather than the checkout and the merge target silently disagreeing about which branch this is. `persist-credentials: false` because every write here goes through the API and nothing uses git, so there is no reason to leave a token in `.git/config`.

What I did not claim, and the comment now says so explicitly: this does not make `workflow_dispatch` safe in general. GitHub reads the workflow DEFINITION from the dispatched ref, so a dispatch from a branch runs that branch's version of this file, pin and all — no workflow can defend against that from the inside, and the real boundary is who may dispatch, i.e. write access. The pin closes the easy path, an unmodified dispatch of an existing branch. That is the whole of the claim.

.github/scripts/update-dependabot-branches.sh
eruizgar91: Correct — the stub only ever emitted `json.dumps(prs)`, so the guard was unpinned. Fixed in 21a9782: `_run` takes `raw_json`, and `test_malformed_json_fails_the_job` hands it `[{"number": 1, "headRefName"`.

Measuring that case turned up a second hole in the same guard, so it is closed here rather than left known-and-unmentioned. `jq` does **not** treat an empty body as an error:

```
$ printf '' | jq -r '.[]' ; echo $?
0
$ printf '[{"number":1' | jq -r '.[]' ; echo $?
5
```

So a truncated body fails (exit 5, caught by the guard) but an EMPTY one with a zero exit does not — and since `--json` always yields at least `[]`, that would be a `gh` malfunction reaching the same silent green by a route your test case does not pass through. It gets its own `[ -z "$prs_json" ]` guard and `test_empty_response_body_fails_the_job`.

Mutation re-run over the whole suite — seven mutants, seven kills, each failing its intended test and no other:

| mutant | test that fails |
|---|---|
| drop the approved-first partition | `test_approved_prs_are_served_first` |
| count a 204 as a merge | `test_already_current_branch_does_not_consume_a_merge_slot` |
| drop the conflict case | `test_conflict_warns_and_the_sweep_continues` |
| drop the `MAX_MERGES` cap | `test_merge_cap_stops_the_wave` |
| restore the bare `prs_json=$(...)` | `test_list_failure_fails_the_job` |
| drop the `jq` parse guard | `test_malformed_json_fails_the_job` |
| drop the empty-body guard | `test_empty_response_body_fails_the_job` |

aaitor (commented):
Fallback-bot pass on `21a9782`: reviewed the full diff plus CI status directly (this session hit a local permission gate that blocks the deterministic run-dir this bot normally uses for its verdict, so this is a plain comment, not the automated approve/request-changes).

No new findings. Both should-fix items from the prior round are confirmed fixed in this diff: `actions/checkout` now pins `ref: ${{ github.event.repository.default_branch }}` with `persist-credentials: false` (dependabot-update-branches.yml), and `test_malformed_json_fails_the_job` / `test_empty_response_body_fails_the_job` now cover the two silent-green parse paths. CI (`deepagents_compat`, `e2e`, `lint`, `unit_integration`) is green on this head.

aaitor (approved):

PR #292, opened 2026-09-21
feat(mcp): publish the API origin per tier as the OAuth `issuer` (changed published identifier — minor)

### LINKED ISSUE #291: fix(mcp): publish the API origin per tier as the OAuth `issuer` — it is tier-blind and now contradicts the RFC 9207 `iss` the webapp returns
## Context

The MCP server the SDK hosts publishes its own RFC 8414 / OIDC discovery documents (`buildAuthorizationServerMetadata` / `buildOidcConfiguration`). Their `issuer` is the **frontend** origin — `Environments[*].frontend`, which is the SAME string (`https://nevermined.app`) for `sandbox` and `live` — while every other field now tells the tiers apart (`authorization_endpoint?network=<tier>` since nevermined-io/payments-py#286, `token_endpoint`, `jwks_uri`, `userinfo_endpoint`, and `authorization_servers` in the RFC 9728 document, which already points at the **API** origin).

Two facts make this more than an identity nit (verified 2026-09-18/21):

- The Nevermined API's own RFC 8414 document publishes `issuer` = **the API origin per tier** — `https://api.sandbox.nevermined.app` / `https://api.live.nevermined.app` (`apps/api/src/well-known/well-known.service.ts`).
- Since nvm-monorepo#3532 (merged 2026-09-15, deployed in v1.41.0) the webapp returns **RFC 9207 `iss`** = that API origin on every authorization response, and the API document advertises `authorization_response_iss_parameter_supported`. An RFC 9207 client compares `iss` with the `issuer` of the document it discovered **by simple string comparison** and rejects the response on any difference. A client that discovers via the MCP server's own `/.well-known/oauth-authorization-server` (the pre-RFC-9728 MCP shape) therefore now rejects every code: `iss=https://api.sandbox.nevermined.app` ≠ `issuer=https://nevermined.app`. RFC 9728 clients follow `authorization_servers` to the API's document and are unaffected.

Raised by @eruizgar91 as a LOW on the tier PR (explicitly "cannot land here": a published identifier that must move together across both SDKs).

## Decision (Rodolphe, 2026-09-21)

**`issuer` = the API origin per tier** — the same backend origin `authorization_servers` already publishes and the API's own document uses. Options declined: keeping the frontend issuer (leaves every direct-discovery RFC 9207 client broken since #3532); investigating client populations first (delays the fix for a break that is live).

## Scope

- `buildOAuthUrls`: `issuer` from the **backend** origin (`new URL(backend).origin` / `urlsplit`), not the frontend. `custom` follows the backend the document publishes (an `oauthUrls.tokenUri` override), exactly as the tier and `resolveAuthorizationServer` already do. An explicit `oauthUrls.issuer` override is passed through unchanged.
- Both documents (`buildAuthorizationServerMetadata`, `buildOidcConfiguration`) and the root server-info agree; the e2e that pins `authorization_endpoint` exactly should pin `issuer` the same way.
- Docs: the SDK MCP integration page's endpoint bullet gains one sentence; note it is a **changed published identifier** — a client keying its AS cache on `issuer` re-discovers/re-consents once. Ship as a **minor**, with the changelog naming it.
- Land in lockstep with the twin so the two SDKs never disagree on the identifier.

Origin: nevermined-io/payments-py#286 review. Twin: see the linked issue in the other SDK repo.

r-marques:
## Description

Closes #291 — twin of nevermined-io/payments#464 (land in lockstep so the two SDKs never disagree on the identifier). **Stacked on #294** (the `_environment_name` bug the panel found; this PR's e2e assertion depends on it) — merge #294 first; GitHub retargets this one to `main`. ⚠️ **Changed published identifier → release as a MINOR**, with this title in the changelog.

The MCP server that `payments.mcp.start()` hosts publishes its own RFC 8414 / OIDC discovery documents. Their `issuer` was the **web app's origin** — `https://nevermined.app`, the same string for `sandbox` and `live` — while `authorization_servers` (RFC 9728), `token_endpoint`, `jwks_uri` and the Nevermined API's own RFC 8414 document all name the **API origin per tier**. Two facts made that a live break rather than an identity nit:

- The API's own document publishes `issuer` = `https://api.<tier>.nevermined.app` (`well-known.service.ts`, reduced to the origin).
- Since nvm-monorepo#3532 (deployed in v1.41.0) the web app returns **RFC 9207 `iss`** = that API origin on every authorization response, and RFC 9207 §2.4 has the client compare `iss` with the discovered `issuer` **by simple string comparison** and reject on any difference. A client that discovers via this server's own `/.well-known/oauth-authorization-server` therefore rejected every code (`iss=https://api.sandbox.nevermined.app` ≠ `issuer=https://nevermined.app`). RFC 9728 clients follow `authorization_servers` to the API's document and were unaffected.

### What changed (after the self-review panel — see the posted review)

- **`issuer` = the CANONICAL API origin of the tier** (`payments_py/mcp/http/oauth_metadata.py`, `_issuer_for`): a **named environment** publishes its canonical backend origin whatever `tokenUri` override fronts it — the environment is the identity, as it already is for the tier. **`custom`** follows the backend it publishes: canonicalised to `api.<tier>.nevermined.<tld>` when the host classifies on a `*.nevermined.(app|dev)` host (branded `acme.api.live…` / `mcp.api.sandbox…` — the API's own document anchors on `API_HOST` and the web app's `iss` on fixed per-tier config, so the canonical origin is the only value that passes the RFC 9207 compare), else the host's own origin, else the environment backend's origin, else the raw string minus its slash — **warned once** through the module logger. `_origin_of` returns `None` for anything without an origin (scheme-less, scheme-only such as `localhost:3001`, bad port). A metadata endpoint never raises.
- **Override hygiene**: only non-empty strings override. `{"issuer": os.getenv("X")}` with `X` unset used to publish `"issuer": null` in a REQUIRED field.
- Docstrings: the "byte for byte `new URL().origin`" claim is scoped to the ASCII hosts Nevermined serves and the third fallback trigger (a bad port) is listed; the false "can never name different servers" property is gone; "every client" → "every RFC 9207 client"; the module doctest is expression-form and now passes. `docs/api/09-mcp-integration.md` names the canonical rule, the `custom` / override behaviour and re-consent.
- ⚠️ **Twin divergence, deliberately NOT touched here:** this SDK's protected-resource documents publish `authorization_servers: [baseUrl]` — the MCP server itself — whereas the TS SDK points at the API origin. That makes this SDK's AS document the *primary* discovery path for RFC 9728 clients (so this fix matters more here). Decided 2026-09-21: tracked as nevermined-io/payments#465 (decision owner Rodolphe; CIMD has landed, so it is unblocked).

### Verification (state the runner)

- `poetry run pytest -m "not slow"`: **1167 passed, 1 skipped, 102 deselected** on `d30d29d` (includes #294's tests underneath); `tests/unit/mcp_tests/test_oauth_metadata.py` 99 passed; the module's doctests 7 passed (`pytest --doctest-modules`). Pins: literal per-environment issuers (never the frontend); the two tiers differ; named-env precedence under a same-tier proxy, cross-tier and malformed `tokenUri` overrides for all four environments; canonical Nevermined hosts vs foreign hosts; unparsable override → environment origin; nothing parsable → raw string, warned once (caplog); unknown-environment fallback; `_origin_of` table incl. per-scheme default ports, userinfo, IPv6, scheme-only and bad-port inputs; override hygiene (`None`, `""`).
- **Mutation-tested (new derivation)**: named env follows the override → 10 red; no canonicalisation → 2; suffix dropped → 3; `port != 443` → 2; `port not in (80, 443)` → 2 (the two the panel found surviving before); netloc verbatim (credentials) → 4; overrides not cleaned → 1; fallback silent → 1. Earlier round: frontend issuer → 11; no origin reduction → 2; override ignored → 4.
- `tests/e2e/mcp_tests/test_mcp_oauth_server.py` (in-process server on a sandbox double that now carries the REAL `environment_name` attribute, via #294): pins the exact issuer `https://api.sandbox.nevermined.app` and that `token_endpoint` hangs off the issuer; 11 passed, 3 pre-existing SSE skips.
- `black --check` clean; `mkdocs build` green.

### Not changed

- RFC 8414 §3.3 (issuer == the URL the document was fetched from) is still not met by a document served from `<mcp>/…` — it was not met by the frontend value either, and the RFC 9728 route is the conformant one and already correct.
- `custom` tier derivation, the `?network=` param, warn-once behaviour (#455) — untouched.


tests/e2e/mcp_tests/test_mcp_oauth_server.py
```diff
@@ -208,6 +208,10 @@ async def test_oauth_authorization_server_metadata(self):
 
             # Required fields per RFC 8414
             assert "issuer" in data
+            # payments-py#291: the issuer is the API ORIGIN of the tier this server runs
+            # against — the value the web app returns as RFC 9207 ``iss``, which clients
+            # string-compare against this field. Exact, on the sandbox mock.
+            assert data["issuer"] == "https://api.sandbox.nevermined.app"
```
r-marques, line 214: **🔴 BLOCKER — The exact-issuer assertion is green only because the double sets an attribute the real class never has**

`server_manager.py:404-406` (untouched) picks the router's environment as `config.get("environment") or getattr(self._payments, "_environment_name", "staging_sandbox")`. The real `Payments` sets **`environment_name`** (`base_payments.py:100`) — every other MCP reader uses that — so a real server started without an explicit `environment` advertises `staging_sandbox` documents whatever key built it. Reproduced on the real class with a production-sandbox key: `environment_name == "sandbox"`, router received `"staging_sandbox"`. This test passes because `PaymentsMock._environment_name = "sandbox"` models the bug. **Split out and fixed first as payments-py#293 / PR #294** (one line + both doubles corrected, unit double `spec`-bound); this PR should rebase onto it so this assertion means what it says.

r-marques (commented):
## 🤖 Automated PR review — 🔴 Blocked

Twin of payments#464: the MCP server's RFC 8414/OIDC `issuer` becomes the origin of the published backend instead of the tier-blind frontend. The change itself is right and the pins are literal; two things block. First, the panel found a **pre-existing bug this PR's e2e depends on**: `server_manager.py` reads `payments._environment_name`, which the real class never sets, so a managed server started without an explicit `environment` advertises `staging_sandbox` whatever key built it — reproduced on the real class with a production-sandbox key; the e2e passes only because its double sets the phantom attribute. That is split out as payments-py#293 / PR #294 and this PR must rebase onto it. Second, as on the TS twin, the issuer follows the *request* host while the web app's RFC 9207 `iss` and the API's own document anchor on the *canonical* tier host, so a branded per-org backend or a same-tier proxy override still fails the compare.

**What I reviewed**
- Diff range `origin/main...HEAD` — 5 files, +191/−12, head `7e0d1c77`.
- Panel: code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer (all delivered by file; worktree clean).
- Re-ran the suite in the review worktree (67 unit / 5 e2e discovery green) and re-measured the author's mutants; found two surviving default-port mutants.
- Note, out of scope and parked as payments#465: this SDK's PRM still points `authorization_servers` at the MCP server itself, unlike TS — so here the SDK document is the *primary* discovery path, which is why the fixes above matter more on this side.

**Verdict:** 🔴 Blocked — 1 blocker (rebase onto #294), 4 should-fix, 5 nits (+1 praise).

### 🧭 Inline comments (11)

- 🔴 **BLOCKER** — The exact-issuer assertion is green only because the double sets an attribute the real class never has — `tests/e2e/mcp_tests/test_mcp_oauth_server.py:214`
- 🟡 **SHOULD FIX** — Any non-canonical backend — branded per-org host, proxy, same-tier tokenUri override — publishes an issuer the web app never returns — `payments_py/mcp/http/oauth_metadata.py:187`
- 🟡 **SHOULD FIX** — `{"issuer": None}` publishes `"issuer": null` in a REQUIRED field — `payments_py/mcp/http/oauth_metadata.py:244`
- 🟡 **SHOULD FIX** — Every `_issuer_of` fallback fires silently — `payments_py/mcp/http/oauth_metadata.py:134`
- 🟡 **SHOULD FIX** — The `_issuer_of` table never exercises the per-scheme default-port branch — `tests/unit/mcp_tests/test_oauth_metadata.py:657`
- 💡 **NIT** — The override half of the invariant test is vacuous for the `live` row — `tests/unit/mcp_tests/test_oauth_metadata.py:628`
- 💡 **NIT** — "byte for byte `new URL(...).origin`" overstates `urlsplit` — `payments_py/mcp/http/oauth_metadata.py:129`
- 💡 **NIT** — "the identifier and the token endpoint can never name different servers" is not a property of this function — `payments_py/mcp/http/oauth_metadata.py:174`
- 💡 **NIT** — "made every client … reject" over-claims; and the module doctest still fails — `payments_py/mcp/http/oauth_metadata.py:167`
- 💡 **NIT** — Doc sentence: hedge the client prediction, and name the `custom` / override behaviour — `docs/api/09-mcp-integration.md:414`
- ✅ **NICE** — Mutant claims re-measured and confirmed — `payments_py/mcp/http/oauth_metadata.py:220`

payments_py/mcp/http/oauth_metadata.py
```diff
@@ -139,7 +184,7 @@ def _build_oauth_urls(
     backend = backend_url.rstrip("/")
 
     return {
-        "issuer": frontend,
+        "issuer": _issuer_of(issuer_backend_url or backend_url),
```
r-marques: **🟡 SHOULD FIX — Any non-canonical backend — branded per-org host, proxy, same-tier tokenUri override — publishes an issuer the web app never returns**

`_issuer_of` reduces the *published* backend to its origin. For a `custom` server on `https://acme.api.live.nevermined.app`, or `sandbox` + `tokenUri: https://gw.corp.com/oauth/token`, the document says `issuer: https://acme.api.live.nevermined.app` / `https://gw.corp.com` — but the web app returns RFC 9207 `iss = https://api.live.nevermined.app` from its fixed per-tier config and the API's own document anchors on `API_HOST`, not the request host. The simple-string compare fails on every response for those configurations, the #291 symptom. **Fix (twin of payments#464's blocker):** a named environment publishes its canonical backend origin (the environment's identity wins, as it already does for the tier); `custom` follows the published backend but, when that classifies to a tier on a `*.nevermined.(app|dev)` host, uses the canonical environment origin for that tier; only a non-Nevermined host keeps its own origin. Pin `acme.api.live.nevermined.app → https://api.live.nevermined.app` and the proxy case.

payments_py/mcp/http/oauth_metadata.py
```diff
@@ -194,7 +241,7 @@ def get_oauth_urls(
     Examples:
         >>> urls = get_oauth_urls("staging_sandbox")
         >>> urls["issuer"]
-        'https://nevermined.dev'
+        'https://api.sandbox.nevermined.dev'
```
r-marques, line 350: **🟡 SHOULD FIX — `{"issuer": None}` publishes `"issuer": null` in a REQUIRED field**

`get_oauth_urls(env, {"issuer": os.getenv("X")})` with the variable unset merges `None` over the computed value and FastAPI serialises `"issuer": null`; `{"tokenUri": ""}` likewise publishes an empty token endpoint. Drop `None` / empty-string values from `overrides` before merging (the TS twin drops the key entirely, which is also wrong — both should keep the computed value), plus a test.

payments_py/mcp/http/oauth_metadata.py
```diff
@@ -117,19 +117,64 @@ def _with_tier_param(authorize_url: str, tier: Optional[OAuthTier]) -> str:
...
+    Nevermined API's own document uses exactly this (``well-known.service.ts``
+    reduces ``API_HOST`` to its origin), and since nvm-monorepo#3532 the web app
+    returns RFC 9207 ``iss`` = that origin on every authorization response — which
+    an RFC 9207 client compares with the ``issuer`` it discovered by SIMPLE STRING
+    comparison and rejects on any difference. So the reduction has to match the web
+    app's ``new URL(...).origin`` byte for byte. A backend ``urlsplit`` cannot read,
+    or one without a scheme/host, keeps the pre-existing "strip the trailing slash"
+    shape rather than crashing a metadata endpoint.
+    """
+    try:
```
r-marques, line 132: **🟡 SHOULD FIX — Every `_issuer_of` fallback fires silently**

A scheme-less or malformed backend (`api.sandbox.nevermined.app`, `not a url`) is served as `issuer` verbatim, 200, `Cache-Control: public, max-age=3600`, with no log anywhere in this module — the TS twin at least warns once on the tierless-`custom` and cross-tier cases. Warn once when the fallback fires, naming the value and `oauthUrls.issuer` as the remedy (same module-level flag shape the TS twin uses), and fall back to the environment's canonical backend origin rather than the raw string where one exists.

<sub>🤖 pr-review-toolkit · silent-failure-hunter · medium</sub>

tests/unit/mcp_tests/test_oauth_metadata.py
```diff
@@ -552,3 +553,121 @@ def test_unknown_environment_falls_back_to_sandbox_including_the_tier(self):
...
+        [
+            # RFC 9207 §2.4 is a simple string comparison against the ``iss`` the web app
+            # returns — ``new URL(backendUrl).origin`` — so this side reduces the same way.
+            (
+                "HTTPS://API.Sandbox.Nevermined.app/oauth/token",
+                "https://api.sandbox.nevermined.app",
+            ),
+            ("https://api.live.nevermined.app/", "https://api.live.nevermined.app"),
+            (
+                "https://api.live.nevermined.app:443/x",
```
r-marques, line 786: **🟡 SHOULD FIX — The `_issuer_of` table never exercises the per-scheme default-port branch**

Mutants `port != 443` and `port not in (80, 443)` both survive all 67 tests (measured). Add `https://host:443/` → dropped, `https://host:80/` → **kept**, `http://host:80/` → dropped, and a userinfo row (`https://user:pw@host/` → credentials stripped — a regression to `parts.netloc` would publish them).

tests/unit/mcp_tests/test_oauth_metadata.py
```diff
@@ -552,3 +553,121 @@ def test_unknown_environment_falls_back_to_sandbox_including_the_tier(self):
...
+                frontend="https://nevermined.app",
+                backend="http://localhost:3001",
+                proxy="",
+                helicone_url="",
+            ),
+        )
+        urls = get_oauth_urls(environment)
+        parts = urlsplit(urls["tokenUri"])
+        assert urls["issuer"] == f"{parts.scheme}://{parts.netloc}"
+        overridden = get_oauth_urls(
```
r-marques: **💡 NIT — The override half of the invariant test is vacuous for the `live` row**

The override is `https://api.live.nevermined.app/oauth/token`, which for `live` equals the environment's own backend, so "override ignored" is invisible on that parametrization. Use a cross-tier override per row — which, under the canonical-issuer fix above, is where the named-environment precedence gets pinned.

payments_py/mcp/http/oauth_metadata.py
```diff
@@ -117,19 +117,64 @@ def _with_tier_param(authorize_url: str, tier: Optional[OAuthTier]) -> str:
...
+def _issuer_of(backend_url: str) -> str:
+    """The RFC 8414 issuer identifier of the authorization server a document describes.
+
+    It is the ORIGIN of the backend whose ``token_endpoint`` the document publishes:
+    scheme + lower-cased host + non-default port, no path, no trailing slash. The
+    Nevermined API's own document uses exactly this (``well-known.service.ts``
+    reduces ``API_HOST`` to its origin), and since nvm-monorepo#3532 the web app
+    returns RFC 9207 ``iss`` = that origin on every authorization response — which
+    an RFC 9207 client compares with the ``issuer`` it discovered by SIMPLE STRING
+    comparison and rejects on any difference. So the reduction has to match the web
```
r-marques: **💡 NIT — "byte for byte `new URL(...).origin`" overstates `urlsplit`**

Python does no WHATWG host canonicalisation — an IDN host stays Unicode where `new URL` would punycode it, a trailing space survives, `https:///host` differs — and the fallback has a third trigger the docstring omits (`parts.port` raising on `:99999`). Say "for the ASCII hosts Nevermined serves" and list the third trigger; the realistic inputs (userinfo, default ports, trailing dot, IPv6) were probed equal.

payments_py/mcp/http/oauth_metadata.py
```diff
@@ -117,19 +117,64 @@ def _with_tier_param(authorize_url: str, tier: Optional[OAuthTier]) -> str:
...
+    RFC 8414 document named the backend. That was a tier-blind identifier, and once
+    the web app started returning RFC 9207 ``iss`` = the API origin
+    (nvm-monorepo#3532) it made every client that discovered through this server's
+    well-known reject its authorization responses (payments-py#291).
 
     Args:
         frontend_url: The frontend URL (e.g., https://nevermined.app).
         backend_url: The backend URL (e.g., https://api.sandbox.nevermined.app).
         tier: The API tier to stamp on the authorize URL, or ``None`` to omit it.
+        issuer_backend_url: The backend the document will actually publish as
```
r-marques: **💡 NIT — "the identifier and the token endpoint can never name different servers" is not a property of this function**

`tokenUri` here is built from `backend_url` (the environment's), not from `issuer_backend_url`; it is `get_oauth_urls`'s override merge that makes them agree. Word it as a property of the published document, or move the sentence there.

payments_py/mcp/http/oauth_metadata.py
```diff
@@ -117,19 +117,64 @@ def _with_tier_param(authorize_url: str, tier: Optional[OAuthTier]) -> str:
...
+    - authorizationUri uses the frontend (the user-facing consent page) and carries
+      the API tier (see :func:`resolve_oauth_tier`)
+    - issuer, tokenUri, jwksUri, userinfoUri use the backend — the API is the
+      authorization server
+
+    ``issuer`` used to be the FRONTEND origin — identical for both tiers, since one
+    web app serves both consent screens — while ``token_endpoint`` and the API's own
+    RFC 8414 document named the backend. That was a tier-blind identifier, and once
+    the web app started returning RFC 9207 ``iss`` = the API origin
+    (nvm-monorepo#3532) it made every client that discovered through this server's
```
r-marques: **💡 NIT — "made every client … reject" over-claims; and the module doctest still fails**

Only an RFC 9207-validating client rejects on `iss`; say so here and in the test class docstring. Separately the module doctest at l.15-16 pairs `print(...)` with a quoted expected value — pre-existing shape, but this PR rewrote the line; drop the quotes or the `print`.

docs/api/09-mcp-integration.md
```diff
@@ -411,7 +411,7 @@ For backward compatibility the server still falls back to reading the access tok
...
...network=live` for `live` — because one Nevermined web app serves the consent screens for both tiers and boots on whatever tier the user's browser last chose. Clients keep that query string when they add their own parameters (RFC 6749 §3.1); if you configure a client by hand, copy the endpoint from the discovery document, query string included. A `custom` environment derives the tier from its backend host (`api.sandbox.…` / `api.live.…`, branded subdomains included); behind a host the SDK cannot classify — `localhost`, a proxy — no tier is stamped, so state it yourself with the `oauthUrls` option: `oauthUrls={"authorizationUri": "<webapp>/oauth/authorize?network=<tier>"}` on `payments.mcp.start()` (or `create_oauth_router()`). Its `issuer` is the **API origin of that tier** — `https://api.sandbox.nevermined.app` for `sandbox`, `https://api.live.nevermined.app` for `live` — the same identifier the Nevermined API's own discovery document publishes and the value the consent page returns as the RFC 9207 `iss` parameter on every authorization response. (Earlier releases published the web app's origin here, identical for both tiers; a client that cached that identifier re-discovers once.)
```
r-marques: **💡 NIT — Doc sentence: hedge the client prediction, and name the `custom` / override behaviour**

"a client that cached that identifier re-discovers once" states a client prediction as fact and drops the user-visible half ("re-discovers / re-consents once"). The sentence also documents `sandbox`/`live` only; that a `custom` backend and a `tokenUri` override move the issuer is documented only in a private docstring — one clause here.

payments_py/mcp/http/oauth_metadata.py
```diff
@@ -172,10 +217,12 @@ def _get_oauth_urls_for_environment(
         environment if environment in Environments else "sandbox"
     )
     env_config = Environments[effective]
+    backend = backend_for_tier or env_config.backend
```
r-marques, line 326: **✅ NICE — Mutant claims re-measured and confirmed**

Issuer back to frontend → 11 unit red plus 2 e2e red; override backend ignored → 4 red; the literal per-environment pins cannot be satisfied by mirroring the implementation.

eruizgar91:

docs/api/09-mcp-integration.md
```diff
@@ -411,7 +411,7 @@ For backward compatibility the server still falls back to reading the access tok
...
...s included); behind a host the SDK cannot classify — `localhost`, a proxy — no tier is stamped, so state it yourself with the `oauthUrls` option: `oauthUrls={"authorizationUri": "<webapp>/oauth/authorize?network=<tier>"}` on `payments.mcp.start()` (or `create_oauth_router()`). Its `issuer` (and the OpenID configuration's) is the **canonical API origin of that tier** — `https://api.sandbox.nevermined.app` for `sandbox`, `https://api.live.nevermined.app` for `live` — the same identifier the Nevermined API's own discovery document publishes and the value the consent page returns as the RFC 9207 `iss` parameter on every authorization response (which clients compare with the discovered `issuer` by simple string comparison). A named environment keeps its canonical origin whatever `tokenUri` proxy sits in front of it; a `custom` environment follows the backend it publishes, reduced to the canonical `api.<tier>.nevermined.…` origin for a branded or MCP subdomain and to the host's own origin otherwise. (Earlier releases published the web app's origin here, identical for both tiers; a client that keyed its authorization-server state on that identifier re-discovers, and may re-consent, once.)
```
eruizgar91: **MEDIUM** — This bullet now says two things that do not line up for the one deployment shape that needs them to.

The pre-existing half tells an operator behind a host the SDK cannot classify — "`localhost`, a proxy" — to fix the tier-blind authorize URL with `oauthUrls={"authorizationUri": "<webapp>/oauth/authorize?network=<tier>"}`. The half this PR appends then says a `custom` environment's `issuer` falls "to the host's own origin otherwise".

For the proxy case those combine badly. A `custom` server on, say, `https://gw.corp.com` that fronts the **real** sandbox API classifies to no tier, so `_canonical_nevermined_origin` returns `None` and `_issuer_for` publishes `https://gw.corp.com` — while the web app returns `iss = https://api.sandbox.nevermined.app` from its fixed per-tier config (`apps/webapp/src/lib/oauth-issuer.ts`). Simple string comparison fails on every authorization response: exactly the #291 symptom, for the deployment the doc is walking the operator through, and silently — the new warn-once fires only in the last-resort branch, which this case never reaches (pinned by `test_custom_follows_the_published_backend_canonicalised_for_nevermined_hosts`, row `"https://gw.corp.com/oauth/token" → "https://gw.corp.com"`).

The derivation is not wrong — the SDK genuinely cannot know what sits behind the proxy — but the remediation advice is now incomplete. Extend the same sentence to name the other override, e.g.: *"…and when that host is a proxy or gateway in front of a Nevermined API rather than the API itself, set `oauthUrls.issuer` (and `oauthUrls.tokenUri`) to that API's real origin too — otherwise the document publishes your proxy's origin as `issuer`, which will not match the `iss` the authorization server returns."* One clause, in a line this PR already edits.

payments_py/mcp/http/oauth_metadata.py
```diff
@@ -117,19 +118,133 @@ def _with_tier_param(authorize_url: str, tier: Optional[OAuthTier]) -> str:
...
+        return _origin_of(env_backend) or env_backend.rstrip("/")
+    issuer = (
+        _canonical_nevermined_origin(published_backend)
+        or _origin_of(published_backend)
+        or _origin_of(env_backend)
+    )
+    if issuer:
+        return issuer
+    raw = published_backend.rstrip("/")
+    if not _issuer_fallback_warned:
```
eruizgar91: **LOW** — Two things about this warning, both small and both fixable here.

**It is warn-once per *process*, not per *value*.** A process serving several `custom` configurations with different unparsable backends logs the first and swallows every later distinct one, while still publishing a wrong raw-string `issuer` for each. The test pins only that the *same* value does not re-warn. The shape used for the equivalent knob in nvm-monorepo (`router-fee.ts`) is once per **distinct** bad value, so a changed typo re-alerts — a module-level `set()` of already-warned raw strings instead of the bool is the same few lines.

**It can fire at an operator who has already done what it asks.** `_get_oauth_urls_for_environment` (l.330) computes the issuer *before* `base_urls.update(clean)` (l.332) applies the overrides, so someone who worked around an unparsable `NVM_BACKEND_URL` by setting `oauthUrls.issuer` correctly still gets told, once per process, to "Set oauthUrls.issuer". Skipping the warning when `clean` carries a non-empty `issuer` is a one-line guard.

(The check-then-set race on the flag is benign — worst case a duplicate log line; the return value never depends on it. Not worth changing.)

payments_py/mcp/http/oauth_metadata.py
```diff
@@ -117,19 +118,133 @@ def _with_tier_param(authorize_url: str, tier: Optional[OAuthTier]) -> str:
...
+    """
+    tier = resolve_oauth_tier("custom", backend_url)
+    if tier is None:
+        return None
+    try:
+        hostname = urlsplit(backend_url).hostname or ""
+    except ValueError:
+        return None
+    environment: Optional[EnvironmentName]
+    if hostname.endswith(".nevermined.app"):
```
eruizgar91, line 171: **LOW** — A trailing-dot FQDN defeats these suffix checks and republishes a non-canonical issuer.

`urlsplit("https://api.sandbox.nevermined.app./oauth/token").hostname` keeps the dot — `"api.sandbox.nevermined.app."` — so `resolve_oauth_tier` still classifies it (`api` + `sandbox` label pair, the dot lands in a trailing empty label) while `hostname.endswith(".nevermined.app")` is `False`. `_canonical_nevermined_origin` returns `None`, `_issuer_for` falls through to `_origin_of(published_backend)`, and the document publishes `https://api.sandbox.nevermined.app.` — one byte off the `iss` the web app returns, which is the whole failure mode this PR exists to close.

Narrow trigger (an operator writing an absolute FQDN in `NVM_BACKEND_URL` or a `tokenUri` override), and the host *is* a Nevermined API host, so canonicalising is the right answer rather than a special case: `hostname = (urlsplit(backend_url).hostname or "").rstrip(".")` before the two checks, which is how DNS treats it anyway. Verified by executing this logic, not inferred.

tests/unit/mcp_tests/test_oauth_metadata.py
```diff
@@ -552,3 +553,235 @@ def test_unknown_environment_falls_back_to_sandbox_including_the_tier(self):
...
+            # A branded per-org subdomain and the Commerce MCP host serve the same API as
+            # ``api.<tier>.nevermined.<tld>``; the API's own document and the web app's ``iss``
+            # both say the canonical origin — the only value that passes the RFC 9207 compare.
+            (
+                "https://acme.api.live.nevermined.app/oauth/token",
+                "https://api.live.nevermined.app",
+            ),
+            (
+                "https://mcp.api.sandbox.nevermined.dev/oauth/token",
+                "https://api.sandbox.nevermined.dev",
```
eruizgar91, line 655: **LOW** — The `.nevermined.dev` + `live` arm of `_canonical_nevermined_origin`'s ternary is never exercised.

`environment = "staging_live" if tier == "live" else "staging_sandbox"` (l.172) only ever runs its false arm in this suite: this row covers `.dev` + `sandbox`, the `acme.…` row covers `.app` + `live`, and no row feeds a `.dev` host with a `live` tier through `get_oauth_urls`. A mutant that collapsed the ternary to `"staging_sandbox"` would survive every test in the file — the one surviving-mutant shape in a table that is otherwise mutation-pinned throughout.

The behaviour is correct today (`get_oauth_urls("custom", {"tokenUri": "https://mcp.api.live.nevermined.dev/oauth/token"})["issuer"]` → `https://api.live.nevermined.dev`, executed). One more row here pins it:

```python
(
    "https://mcp.api.live.nevermined.dev/oauth/token",
    "https://api.live.nevermined.dev",
),
```

eruizgar91 (commented):
Reviewed the diff of #292 against **its own base** (`fix/mcp-start-reads-environment-name-293`, merge-base `4c2b5589f`) — 386/-18 across 5 files — so #294's one-line fix underneath is out of scope here, as is the TS twin payments#464.

The derivation is correct and the load-bearing invariant holds. Four findings, all inline, all small: one on the documented remediation for a proxied `custom` deployment, one on the quality of the new fallback warning, one narrow host-form gap, one unpinned branch.

### What was verified (not just read)

- **The named-environment invariant holds for all four shipped environments, not only the one in the test.** `_issuer_for` returns `_origin_of(env_backend)` for every non-`custom` environment, and `Environments[*].backend` is already a bare origin for `sandbox` / `live` / `staging_sandbox` / `staging_live` — so `_origin_of` is the identity on each and no `tokenUri` proxy can move the issuer. Cross-checked against the other side of the compare rather than taken on trust: nvm-monorepo `apps/webapp/src/lib/oauth-issuer.ts` builds `iss` as `new URL(backendUrl).origin` from **fixed per-tier config**, never the request host, and `WellKnownService.issuerIdentifier` anchors the API's own `issuer` on `API_HOST`. That is exactly what makes the branded/`mcp.` canonicalisation in `_canonical_nevermined_origin` right, and what makes ignoring the override under a named environment right.
- **The `custom` host that does NOT match `*.api.<tier>.nevermined.*` fails to the host's own origin — it never falls back to the old tier-blind web-app origin.** So the bug this PR fixes is not silently reproduced for a self-hosted stack, where that origin genuinely is the authorization server. It *is* reproduced for one shape — a proxy fronting a **real** Nevermined API — which is finding 1 below, and it is a doc gap rather than a wrong derivation (the SDK cannot know).
- **Nothing anywhere in the repo still asserts the old value.** Swept the whole tree at head: the only surviving `https://nevermined.{app,dev}` literals in the test file are the `frontend` fixtures used to assert `issuer != frontend`, which is the point. No README, example, mkdocs page or fixture publishes an OAuth `issuer`.
- **All 18 doctests in the module pass** (executed against the extracted head, not eyeballed), including the rewritten module-level one now in expression form.
- **The missing version bump is correct.** `.github/workflows/prepare-release.yml` owns `pyproject.toml` + `CHANGELOG.md` via `workflow_dispatch`; every recent release lands as its own `chore: bump version to …` PR. A `minor` is the right call for a changed published identifier — that is a release-time input, not something this diff can carry.
- **Nothing in this repo consumes `issuer`.** Grepped every non-test module: no JWT `iss` validation, no state keyed on it, no comparison outside `oauth_metadata.py`. The change is publish-only.
- **The override-hygiene change breaks no caller.** Every site that builds `oauthUrls` (`oauth_router.py:115/140`, `server_manager.py:449`) forwards the operator dict unmodified, and the declared type was `Dict[str, str]` already.

### Considered and deliberately not raised

- **The Python/TypeScript divergence on `authorization_servers` is deliberate, and this PR does not make it worse.** The TS twin makes `authorization_servers === issuer`; this SDK keeps `[baseUrl]`, so an RFC 9728 client is routed back to this server's own AS document, whose `issuer` is now a different origin from the URL it was fetched from (RFC 8414 §3.3). That was equally untrue before — the old value was `https://nevermined.app`, also ≠ `baseUrl` — so it is a pre-existing divergence, not a regression, and the PR body records the decision and parks it as payments#465 with an owner. Correctly out of scope.
- **Adjacent to that, and worth knowing while payments#465 is open:** the TS SDK also warns once on a tierless `custom` host and once on a cross-tier `tokenUri` override (`getOAuthUrlsForEnvironment`, from payments#455); this SDK has neither. Pre-existing and listed under "Not changed", so not filed here — but #292 raises the stakes on the first of the two, because that same unclassifiable host now also decides a published security identifier. That is the context for finding 1.
- The eight items the self-review panel raised are all addressed in `a367787`; none is re-filed.
- RFC 8414 §3.3 non-conformance of a document served from `<mcp>/…`: acknowledged in the PR body, unchanged by this diff.

docs/api/09-mcp-integration.md
r-marques: Folded in `a7b5651`, both halves: the docs sentence now carries your clause (a proxy/gateway in front of a Nevermined API → set `oauthUrls.issuer` and `tokenUri` to the real origin, otherwise the issuer will not match the `iss`), and the code no longer stays mute on it — the branch that publishes the host's own origin warns once per distinct value, naming the derived origin and `oauthUrls.issuer` (twin of payments#464's fix). Pinned for `gw.corp.com` and `api.live.example.com`; suppressed by an `issuer` override; silent for loopback (a local stack, which is not a proxy). Mutant (never fires) → 1 red.

payments_py/mcp/http/oauth_metadata.py
r-marques: Folded in `a7b5651`: a module-level `set()` of already-warned values (a changed typo re-alerts; mutant to a process-wide bool → 1 red), and `get_oauth_urls` passes `"issuer" in clean` down so an operator who already set `oauthUrls.issuer` is never told to (mutant → 1 red).

payments_py/mcp/http/oauth_metadata.py
r-marques: Folded in `a7b5651`: `.rstrip(".")` on the hostname before the suffix checks, with the row `https://api.sandbox.nevermined.app./oauth/token → https://api.sandbox.nevermined.app` (mutant → 1 red). Same fix on the TS twin, where `new URL().hostname` keeps the dot too.

tests/unit/mcp_tests/test_oauth_metadata.py
r-marques: Folded in `a7b5651`: your row, verbatim; collapsing the ternary is now 1 red.

eruizgar91:
👀 pr-review-bot (fallback) picking this up — round 2.

eruizgar91 (commented):
**Round 2 — confirmation pass** on `a7b565139a1695776b1157d8560dfdbd44d38764`. Base is now `main`: payments-py#294 merged and GitHub retargeted this PR, which resolves the stacked-base rebase that moved the head out from under round 1.

Round 1 (`a367787e`, [review](https://github.com/nevermined-io/payments-py/pull/292#pullrequestreview-5267810992)) raised 1 MEDIUM + 3 LOW. **All four are folded at this head**, each confirmed by executing the code rather than by reading the diff. No new findings — this review carries no inline comments.

### Disposition of the four

1. **MEDIUM — `docs/api/09-mcp-integration.md:414`** (proxy remediation named only `authorizationUri`). **Folded.** The bullet now names the actual remedy for the proxy case: *"when that host is a proxy or gateway in front of a Nevermined API rather than the API itself, set `oauthUrls.issuer` (and `oauthUrls.tokenUri`) to that API's real origin too"*, with the RFC 9207 consequence spelled out. The commit went further than the clause asked for and closed the code half too: the warning now fires in the `own` branch (`_issuer_for`, l.235), which is exactly the branch the round-1 warn-once could never reach — so a proxied `custom` deployment is no longer silent.

2. **LOW — `oauth_metadata.py:211`** (once-per-*process*, and fired at an operator who already set `oauthUrls.issuer`). **Folded, and the key is the right one.** `_issuer_fallback_warned: bool` is replaced by `_issuer_warned: set` keyed on the **derived origin** (`own`) in the proxy branch and on `raw` in the last-resort branch. That key determines the message — the message embeds `urlsplit(published_backend).netloc`, constant for a given derived origin — so no distinct message can be swallowed by a colliding key, which is the failure mode a "once per value" rewrite usually still has. Executed: two distinct hosts → 2 warnings, set holds `{'https://gw.corp.com', 'https://gw2.corp.com'}`; same host twice → 1. The override suppression is wired correctly *after* the override hygiene rather than before it — `get_oauth_urls` computes `clean` first and passes `"issuer" in clean` (l.370-372), so only a genuinely non-empty override suppresses: executed, an `issuer=""` override (which does **not** override) still warns and still publishes the derived origin. Loopback is excluded too.

3. **LOW — `oauth_metadata.py:169`** (trailing-dot FQDN republished a one-byte-off issuer). **Folded.** `.rstrip(".")` before the two suffix checks. Executed: `https://api.sandbox.nevermined.app./oauth/token` → `https://api.sandbox.nevermined.app`, byte-identical to the no-dot control.

4. **LOW — `tests/unit/mcp_tests/test_oauth_metadata.py:655`** (the `.dev` + `live` ternary arm was never exercised). **Folded — and the row was verified to actually pin the arm, by executing the mutation.** Collapsing `environment = "staging_live" if tier == "live" else "staging_sandbox"` to `"staging_sandbox"` now takes the suite from **102 passed** to **1 failed / 101 passed**, and the single red test is precisely the new row (`[https://mcp.api.live.nevermined.dev/oauth/token-https://api.live.nevermined.dev]`), asserting `api.live.…` against the mutant's `api.sandbox.…`. The mutant is killed by the added row and by nothing else. File restored; working tree clean.

### Checked and discarded

- **`_issuer_warned` unbounded growth.** A module-level set that never evicts is normally a leak. Every caller (`build_authorization_server_metadata`, `build_oidc_configuration`, `build_server_info_response`, the two protected-resource builders) passes `config["environment"]` and `config.get("oauthUrls")` from the server's fixed startup `OAuthConfig` — never from the HTTP request — so cardinality is bounded by configuration, not by traffic. Not a finding.
- **`_is_loopback` covers only `localhost` / `127.0.0.1` / `::1`.** `127.0.0.2` or `0.0.0.0` would draw a spurious advisory warning. The warning changes no published value, so this is cosmetic; below the bar.
- **Key collision between the two warn branches.** `own` is always a `scheme://host` origin and `raw` is only reached when `_origin_of` failed for both backends, so the two key spaces do not realistically overlap.
- **f-string built eagerly before the set is consulted.** Negligible; not worth the lazy-`%s` form back.
- **New round-2 test isolation.** `test_custom_non_nevermined_host_warns_once_per_value_unless_overridden` monkeypatches `_issuer_warned` to a fresh set and filters `caplog` on `"proxy or gateway"`, so it neither leaks nor picks up the sibling fallback warning. It pins all four properties (once-per-value, distinct-value re-alerts, override suppresses, loopback exempt).

### Suite

`tests/unit/mcp_tests/test_oauth_metadata.py` **102 passed**; doctests on `oauth_metadata.py` **7 passed**; full `tests/unit` **1100 passed, 7 failed**. The 7 are all `tests/unit/langsmith/test_spans.py` and reproduce on `origin/main` at the merge-base — pre-existing, unrelated to this PR. Two further collection errors (`tests/unit/x402/test_{deepagents_compat,langchain_decorator}.py`) are a missing optional `langchain_core` extra in the local env, not a repo defect.

Items round 1 verified and cleared are not re-examined here (backend-derived issuer for a non-matching `custom` host, the named-environment invariant under any `tokenUri` override, the deliberately absent version bump, no stale assertions of the old value, no repo consumer of `issuer`). The Python/TypeScript `authorization_servers` asymmetry stays parked as payments#465.

<sub>Automated review · round 2 · verdict issued separately.</sub>

eruizgar91 (approved):

Inline review: https://github.com/nevermined-io/payments-py/pull/292#pullrequestreview-5268370890

PR #294, opened 2026-09-21
fix(mcp): `payments.mcp.start()` read a phantom `_environment_name` and advertised staging_sandbox discovery for every key

### LINKED ISSUE #293: fix(mcp): payments.mcp.start() ignores the key's environment and advertises staging_sandbox discovery documents (reads a phantom `_environment_name`)
## Bug

`McpServerManager.start()` picks the environment for the OAuth discovery documents as

```python
environment = config.get("environment") or getattr(self._payments, "_environment_name", "staging_sandbox")
```

(`payments_py/mcp/core/server_manager.py:404-406`, since `2bdece2`, 2026-01-26). **The real `Payments` class has no `_environment_name`** — it sets `environment_name` (`payments_py/api/base_payments.py:100`), which every other MCP reader uses (`core/auth.py`, `core/paywall.py`). So unless the caller passes `environment` explicitly, `getattr` returns the default and **every managed server advertises `staging_sandbox`** — `token_endpoint`, `jwks_uri`, `userinfo_endpoint` and (since #286) `authorization_endpoint?network=sandbox` all pointing at `nevermined.dev`, whatever key the server was constructed with.

Neither managed entry point passes `environment`: the decorator `PaymentsMCP.start()` builds its config without it (`payments_mcp.py:562-572`), and `payments.mcp.start(config)` only carries what the caller wrote. The documented shape (`await mcp.start(port=3002)`) hits the bug.

**Reproduced 2026-09-21** on the real class with a production-sandbox key and no `environment` anywhere: `payments.environment_name == "sandbox"`, router received `environment == "staging_sandbox"`. (With a staging-sandbox key the bug is invisible, which is presumably how it survived: the fallback happens to be right.)

Why the tests did not catch it: both doubles set the phantom attribute — `PaymentsMock._environment_name = "sandbox"` in `tests/e2e/mcp_tests/test_mcp_oauth_server.py` and the `MagicMock` in `tests/unit/mcp_tests/test_server_manager_oauth_urls.py` — so they modelled the bug rather than the class. Surfaced by the `/review-pr` panel on payments-py#292 (silent-failure-hunter), which noticed the e2e's exact-issuer assertion was green only because of the double.

## Fix

Read `environment_name` (the attribute the class has), keep `staging_sandbox` as the last-resort default only for an object that has neither; correct both doubles to the real attribute; add a unit test that a payments object with `environment_name="sandbox"` and no `_environment_name` yields a router `environment == "sandbox"`, and that an explicit `config["environment"]` still wins. The TS twin already reads the real `environmentName` (`managed-server.ts:227`).

Impact statement for the changelog: a Python-hosted MCP server that relied on its key's environment has been advertising staging discovery documents; after this fix it advertises the key's tier. Operators who passed `environment` explicitly are unaffected.

r-marques:
## Description

Closes #293. **A one-line bug with a wide blast radius, so it ships on its own** rather than inside #292 where the review panel found it.

`McpServerManager.start()` chose the environment for the OAuth discovery documents as `config.get("environment") or getattr(payments, "_environment_name", "staging_sandbox")`. The real `Payments` never sets `_environment_name` — it sets `environment_name` (`payments_py/api/base_payments.py:100`), the attribute every other MCP reader uses (`core/auth.py`, `core/paywall.py`). So unless the caller passed `environment` explicitly, the default won and **every managed server advertised `staging_sandbox`** — `token_endpoint`, `jwks_uri`, `userinfo_endpoint` and (since #286) `authorization_endpoint?network=sandbox` on `nevermined.dev` — whatever key it was built with. Neither managed entry point passes `environment` (`PaymentsMCP.start()` builds its config without it; `payments.mcp.start(config)` carries only what the caller wrote), and the documented `await mcp.start(port=3002)` hits it.

**Reproduced on the real class, 2026-09-21**: production-sandbox key, no `environment` anywhere → `payments.environment_name == "sandbox"`, router received `"staging_sandbox"`. With a staging-sandbox key the bug is invisible (the fallback happens to be right), which is how it survived since `2bdece2` (2026-01-26).

### What changed

- `server_manager.py`: read `environment_name`; `staging_sandbox` remains the last resort only for an object that has neither attribute. The TS twin already reads the real `environmentName` (`managed-server.ts:227`).
- Both test doubles modelled the bug — `PaymentsMock._environment_name` in the e2e and the `MagicMock` in `test_server_manager_oauth_urls.py` — they now carry the real attribute, and the unit double is `spec`-bound so a permissive mock can never again answer a misspelled attribute read. **And one test no longer trusts any double**: it starts the manager on a **real `Payments`** built offline from the repo's unsigned test JWT (`sandbox:` and `live:` prefixes) and asserts the router gets the prefix's environment — the check against the class itself that #293 lacked.
- The last-resort `staging_sandbox` (unreachable on a real instance — the constructor always sets `environment_name`) is now **logged** through the module logger and `onLog`, naming `config['environment']` as the remedy; a silent default is the shape that hid this for eight months.
- Visible behaviour change worth knowing: a key with an unrecognised prefix and no `environment` option resolves to `custom`, so such a server now advertises `Environments["custom"]` documents (`NVM_BACKEND_URL` / `NVM_FRONTEND_URL`) instead of staging — correct, and the tier-omission for an unclassifiable host is already documented.

### Verification (runner named)

- `poetry run pytest -m "not slow"`: **1119 passed, 1 skipped, 102 deselected** on the folded head; the two touched files: 17 passed, 3 pre-existing SSE skips. Tests: key-derived environment reaches the router (double **and** real `Payments`, sandbox + live); an explicit `config["environment"]` still wins; a bare object falls back to `staging_sandbox` and says so.
- **Mutant** (phantom name restored): **6 red** — the two real-instance rows, the double-based test, the `spec`-bound forward test, and both e2e discovery tests through the corrected double. `black --check` clean.

### Release note

A Python-hosted MCP server that relied on its API key's environment has been advertising **staging** discovery documents; after this fix it advertises the key's tier. Operators who passed `environment` explicitly are unaffected. Ship as a **patch** is defensible, but it lands in the same release as #292 (minor) either way.


tests/unit/mcp_tests/test_server_manager_oauth_urls.py
```diff
@@ -37,9 +37,11 @@ def _base_config(**extra):
...
 def _payments(environment_name):
     # ``start()`` configures the paywall integration before it builds the router; a mock with
-    # the environment name the manager falls back to is all that path needs.
-    payments = MagicMock()
-    payments._environment_name = environment_name
+    # the environment name the manager falls back to is all that path needs. ``spec`` keeps the
+    # double honest: a ``MagicMock()`` without it answers ANY attribute, so a manager reading a
+    # misspelled one (the #293 bug read ``_environment_name``) would still get a value here.
+    payments = MagicMock(spec=["environment_name", "mcp"])
+    payments.environment_name = environment_name
```
r-marques, line 53: **🟡 SHOULD FIX — Every mutant-killing test checks the manager against another double — anchor one on a REAL Payments**

The `spec=["environment_name", "mcp"]` double here and the e2e `PaymentsMock` are both written by the hand that wrote the fix: the `spec` list is a second copy of the belief "the attribute is `environment_name`", not a check against `BasePaymentsAPI`. #293 was exactly a double and a reader agreeing on a name the class never had. A real `Payments` builds offline — `Payments.get_instance(PaymentOptions(nvm_api_key="sandbox:" + <the unsigned JWT from tests/unit/test_base_payments_http.py>))` — so add a test that starts the manager on the real class and asserts the router gets the prefix's environment.

payments_py/mcp/core/server_manager.py
```diff
@@ -400,9 +400,19 @@ async def start(self, config: McpServerConfig) -> McpServerResult:
...
+            # `environment_name`, the attribute the class actually sets (base_payments.py)
+            # and every other MCP reader uses (core/auth.py, core/paywall.py). This line
+            # read a phantom `_environment_name` from January to September 2026 and so
+            # advertised staging_sandbox for every server that did not pass `environment`
+            # explicitly (payments-py#293). The default stays only for an object that has
+            # neither attribute.
+            environment = (
+                config.get("environment")
+                or getattr(self._payments, "environment_name", None)
+                or "staging_sandbox"
```
r-marques: **🟡 SHOULD FIX — The last-resort `staging_sandbox` still fires in silence**

Unreachable on a real `Payments` (its constructor always sets `environment_name`), so only a double or a foreign object lands here — but a silent default is the exact shape that hid #293 for eight months. Log it (module logger, plus `onLog` when present), naming the default and `config['environment']` as the remedy.

tests/unit/mcp_tests/test_server_manager_oauth_urls.py
```diff
@@ -76,3 +78,50 @@ async def test_start_without_oauth_urls_forwards_none_not_a_default(monkeypatch)
...
+
+@pytest.mark.asyncio
+async def test_start_defaults_to_staging_sandbox_only_when_the_object_has_no_environment(
+    monkeypatch,
+):
+    captured = {}
+    monkeypatch.setattr(sm, "create_oauth_router", _spy(captured))
+    bare = MagicMock(
+        spec=["mcp"]
+    )  # neither attribute, as before the key-derived environment existed
```
r-marques: **💡 NIT — Comment is historically wrong: `environment_name` predates key derivation**

`git log -S'self._environment_name' origin/main -- payments_py/` is empty and `2bdece2:base_payments.py` already sets `environment_name` — the class never had the phantom, so "as before the key-derived environment existed" describes a state that never was. Say "a foreign object, never a real Payments".

payments_py/mcp/core/server_manager.py
```diff
@@ -400,9 +400,19 @@ async def start(self, config: McpServerConfig) -> McpServerResult:
...
+            # must come from the key the Payments object was built with — that is
+            # `environment_name`, the attribute the class actually sets (base_payments.py)
+            # and every other MCP reader uses (core/auth.py, core/paywall.py). This line
+            # read a phantom `_environment_name` from January to September 2026 and so
+            # advertised staging_sandbox for every server that did not pass `environment`
+            # explicitly (payments-py#293). The default stays only for an object that has
+            # neither attribute.
+            environment = (
+                config.get("environment")
+                or getattr(self._payments, "environment_name", None)
```
r-marques: **💡 NIT — State the `custom` behaviour change in the PR body**

A key with an unrecognised prefix and no `environment` option resolves to `custom`, so such a server now advertises `Environments["custom"]` documents (`NVM_BACKEND_URL` / `NVM_FRONTEND_URL`) instead of staging — correct, and the tier-omission for an unclassifiable host is already documented, but it is a visible change for that population.

tests/unit/mcp_tests/test_server_manager_oauth_urls.py
```diff
@@ -76,3 +78,50 @@ async def test_start_without_oauth_urls_forwards_none_not_a_default(monkeypatch)
...
+    captured = {}
+    monkeypatch.setattr(sm, "create_oauth_router", _spy(captured))
+    manager = sm.McpServerManager(_payments("sandbox"))
+
+    with pytest.raises(_StopAtRouter):
+        await manager.start(_base_config())
+
+    assert captured["environment"] == "sandbox"
+
+
```
r-marques, line 109: **✅ NICE — Precedence `config > environment_name > default` is pinned in the right order**

The swap mutant and the drop-config mutant are each killed by exactly the explicit-config test; the e2e's exact `issuer` assertion is genuinely driven by `environment_name` (2 e2e tests go red under the phantom mutant).

r-marques (commented):

One-line bug with a wide blast radius: `McpServerManager.start()` read `payments._environment_name`, which the real class never had, so every managed Python MCP server started without an explicit `environment` advertised `staging_sandbox` discovery documents since January. The fix reads the attribute the class sets and corrects both test doubles that had modelled the bug. Two should-fix items, both about not repeating the failure shape: anchor at least one test on a real `Payments` rather than a third double, and never let the last-resort default fire silently again.

**What I reviewed**
- Diff range `origin/main...HEAD` — 3 files, head `55723b48`.
- Panel: code-reviewer, pr-test-analyzer (both delivered by file; worktree clean).
- Verified `git log -S` that the class never set the phantom attribute, that `environment_name` is always set by the constructor (so the default is unreachable on a real instance), and that a real `Payments` is constructible offline with the repo's unsigned test JWT.

**Verdict:** 🟡 Mergeable with nits — 0 blockers, 2 should-fix, 2 nits (+1 praise).

### 🧭 Inline comments (5)

- 🟡 **SHOULD FIX** — Every mutant-killing test checks the manager against another double — anchor one on a REAL Payments — `tests/unit/mcp_tests/test_server_manager_oauth_urls.py:44`
- 🟡 **SHOULD FIX** — The last-resort `staging_sandbox` still fires in silence — `payments_py/mcp/core/server_manager.py:415`
- 💡 **NIT** — Comment is historically wrong: `environment_name` predates key derivation — `tests/unit/mcp_tests/test_server_manager_oauth_urls.py:121`
- 💡 **NIT** — State the `custom` behaviour change in the PR body — `payments_py/mcp/core/server_manager.py:414`
- ✅ **NICE** — Precedence `config > environment_name > default` is pinned in the right order — `tests/unit/mcp_tests/test_server_manager_oauth_urls.py:100`

eruizgar91:

payments_py/mcp/core/server_manager.py
```diff
@@ -400,10 +400,32 @@ async def start(self, config: McpServerConfig) -> McpServerResult:
...
+            # Create OAuth router. The environment decides which API the discovery
+            # documents name (token_endpoint, jwks, the tier on the authorize URL), so it
+            # must come from the key the Payments object was built with — that is
+            # `environment_name`, the attribute the class actually sets (base_payments.py)
+            # and every other MCP reader uses (core/auth.py, core/paywall.py). This line
+            # read a phantom `_environment_name` from January to September 2026 and so
+            # advertised staging_sandbox for every server that did not pass `environment`
+            # explicitly (payments-py#293). The default stays only for an object that has
+            # neither attribute.
             environment = config.get("environment") or getattr(
```
eruizgar91: **MEDIUM** — `config["environment"]` is forwarded to the document builders **unvalidated**, and an unrecognised value is silently coerced to `sandbox` — the same wrong-tier-in-silence failure this PR is fixing, on the input the new log line recommends as the remedy.

`_get_oauth_urls_for_environment` substitutes `sandbox` for any name not in `Environments`, without a log ([`oauth_metadata.py#L170-L174`](https://github.com/nevermined-io/payments-py/blob/4c2b5589f05d0a446ea07235aff839fc95e84a98/payments_py/mcp/http/oauth_metadata.py#L170-L174)). Measured on this head, driving `McpServerManager.start()` on a real `Payments` built from a **`live:`** key:

```
config.environment='production'      -> forwarded='production'
    token_endpoint=https://api.sandbox.nevermined.app/oauth/token
    authorize=https://nevermined.app/oauth/authorize?network=sandbox
config.environment='staging-sandbox' -> forwarded='staging-sandbox'   # hyphen, not underscore
    token_endpoint=https://api.sandbox.nevermined.app/oauth/token
    authorize=https://nevermined.app/oauth/authorize?network=sandbox
```

Nothing was emitted at any level; I ran it with `logging.basicConfig(level=DEBUG)`. A live operator who writes `environment="production"` gets a server advertising production **sandbox** endpoints with a `?network=sandbox` authorize URL — exactly the "Connector not authorized" shape #286 fixed.

The type says `Optional[EnvironmentName]`, but that is a `Literal` in a `TypedDict`: nothing checks it at runtime, and this is an SDK consumed from untyped code. And the SDK's own convention for an unknown environment name is to **refuse**, not to substitute — `get_environment()` raises `ValueError(f"Environment '{name}' is not defined.")` ([`environments.py#L124-L127`](https://github.com/nevermined-io/payments-py/blob/4c2b5589f05d0a446ea07235aff839fc95e84a98/payments_py/environments.py#L124-L127)), which is why a real `Payments` can never hold an invalid `environment_name`.

Validate it here, where the value is assembled, so the coercion cannot happen downstream:

```python
configured = config.get("environment")
if configured and configured not in Environments:
    raise ValueError(
        f"Unknown MCP environment {configured!r}; expected one of {sorted(Environments)}"
    )
environment = configured or getattr(self._payments, "environment_name", None)
```

Raising matches `get_environment`; if a hard failure at `start()` is too strong, log at WARNING and fall through to `environment_name` — either is better than publishing sandbox documents for a live key with no signal at all.

payments_py/mcp/core/server_manager.py
```diff
@@ -400,10 +400,32 @@ async def start(self, config: McpServerConfig) -> McpServerResult:
...
-                self._payments, "_environment_name", "staging_sandbox"
+                self._payments, "environment_name", None
```
eruizgar91: **MEDIUM** — an explicit `config["environment"]` silently outranks the key, and **nothing checks the two agree** — so a mismatch produces a server whose discovery documents name one tier while its x402 and paywall halves use the other.

`config.get("environment")` wins unconditionally, but no other MCP reader consults it: `auth.py` ([L103-L106](https://github.com/nevermined-io/payments-py/blob/4c2b5589f05d0a446ea07235aff839fc95e84a98/payments_py/mcp/core/auth.py#L103-L106), and L180) and `paywall.py` ([L318-L322](https://github.com/nevermined-io/payments-py/blob/4c2b5589f05d0a446ea07235aff839fc95e84a98/payments_py/mcp/core/paywall.py#L318-L322), and L364/L412) all take `getattr(self._payments, "environment_name", None)` directly. Measured on this head with a **`sandbox:`** key and `config["environment"]="live"`: the served documents publish `https://api.live.nevermined.app/oauth/token` and `?network=live`, while `build_payment_required_for_plans` still declares the sandbox network and the facilitator still verifies against sandbox. The client is sent to the live authorization server for a token the server will verify on sandbox — one broken server, no signal.

The precedence is also inverted relative to the SDK's own decision. `BasePaymentsAPI._resolve_environment` makes the **key prefix always win**, ignores a disagreeing `environment` option, *and warns when it does* ([`base_payments.py#L140-L157`](https://github.com/nevermined-io/payments-py/blob/4c2b5589f05d0a446ea07235aff839fc95e84a98/payments_py/api/base_payments.py#L140-L157)) — the option is deprecated precisely because it could disagree with the key. This line resurrects it with the opposite precedence and no warning.

Keeping config-first is defensible (it is the documented escape hatch, and `oauthUrls` covers the unclassifiable-host case separately), so the bounded fix is to make the disagreement loud rather than to change the order — a few lines next to the block you just added, in the same spirit as it:

```python
key_environment = getattr(self._payments, "environment_name", None)
if configured and key_environment and configured != key_environment:
    message = (
        f"[MCP] config['environment']={configured!r} overrides the API key's "
        f"{key_environment!r}; the OAuth discovery documents will name "
        f"{configured!r} while payment verification stays on {key_environment!r}."
    )
    logging.getLogger(__name__).warning(message)
    if self._log:
        self._log(message)
```

Same rationale as the fallback you made loud below: the failure this PR is closing was invisible for eight months because the wrong tier was chosen without saying so.

eruizgar91 (commented):
Reviewed the one-line environment fix and its test scaffolding. The fix itself is right, and I verified it rather than taking the PR body's word for it.

**Verified here, on this head (`4c2b558`):**

- **The new tests really do kill the mutant.** Restoring `_environment_name` in `server_manager.py` turns `tests/unit/mcp_tests/test_server_manager_oauth_urls.py` from `8 passed` to `4 failed, 4 passed` — including both `test_start_reads_the_environment_off_a_REAL_payments_instance[sandbox|live]` rows. The real-`Payments` anchor is the part that matters: the `spec`-bound doubles encode the same belief a second time, the real instance checks it against the class. `MagicMock(spec=[...])` is what makes the double-based row able to fail at all — a bare `MagicMock()` answers any attribute and would have stayed green.
- **The end-to-end behaviour.** Driving `McpServerManager.start()` on a real `Payments` built from a `live:` key with no `environment` in the config now forwards `live` and the document publishes `https://api.live.nevermined.app/oauth/token` + `…/oauth/authorize?network=live`. Before the fix it was `staging_sandbox`.
- **Explicit `environment` is untouched.** `config.get("environment")` is still evaluated first, so a server that passed one sees no change. Confirmed by probe, not just by reading.
- **No other double still carries the phantom.** `_environment_name` survives only in three comments. I also swept every attribute the suite assigns onto a `payments`-shaped double against `dir(Payments)`: the only two names that do not exist on the real class are `function_name` (a Lambda context mock) and `organization_id` (a `PaymentOptions` mock, where it is real) — neither is the same class of defect.
- **The other MCP readers were already correct.** `core/auth.py:105,180` and `core/paywall.py:320,364,412` all read `environment_name`; `server_manager` was the lone divergent reader, and `create_oauth_router` has exactly one caller.
- Full fast suite on this venv: `1075 passed / 18 failed` vs `1069 passed / 18 failed` on `origin/main` — the same 18 pre-existing `langsmith`/`a2a` failures on both, +6 new passing rows. `black --check` clean on all three files.

**Two findings, both inline, both MEDIUM, both about the *other* silent default in the same path** — the PR's own thesis is that "a silent default is the shape that hid this for eight months", and the branch that the new log line points operators at (`config['environment']`) is still unvalidated and still silently outranks the key. Neither is a regression introduced here; both are one small block in a file this PR already touches.

**Considered and discarded:**

- *The `custom` behaviour change.* A key with an unrecognised prefix and no `environment` now resolves to `custom`, i.e. `NVM_*_URL` or `http://localhost:300x` discovery documents instead of staging ones. Real, flagged in the PR body, and correct: the only way to reach it is a non-`sandbox|live|sandbox-staging|live-staging` prefix, which in practice is a local/custom deployment where localhost *is* the right answer, and whose Payments object was already talking to that backend while the documents said staging.
- *`self._log(message)` passes one argument while `McpServerConfig.onLog` is typed `Callable[[str], Optional[str]]`-style two-arg.* Pre-existing and uniform — every one of the six `self._log(...)` call sites in this file is single-arg, `PaymentsMCP.start()` declares `on_log: Optional[Callable[[str], None]]`, and `oauth_router.log()` does the same. The new block copies the established shape at `server_manager.py:671-678` exactly (module logger unconditionally, `onLog` if present). Not this PR's to fix.
- *`docs/api/09-mcp-integration.md`.* It documents how the tier is stamped but never said where the environment came from, so nothing in it became false. `McpServerConfig.environment`'s own docstring already said "Defaults to Payments instance env" — the code had diverged from its own documented contract, which is now restored rather than changed.
- *Version / CHANGELOG.* Correctly absent — the release workflow owns both.

CI is green (`lint`, `unit_integration`, `e2e`, `deepagents_compat`).

eruizgar91 (approved):
🟡 **Approved with comments** — no blockers; 2 should-fix, 0 nit(s) left inline.

Inline review: https://github.com/nevermined-io/payments-py/pull/294#pullrequestreview-5267745924

payments_py/mcp/core/server_manager.py
r-marques: Folded in `e86c8dc` — raised, as `get_environment()` does: `ValueError("Unknown MCP environment 'production'; expected one of [...]")` at `start()`, before anything is built, and the manager rewinds to IDLE like any other start failure. Test pins it never reaches the router; mutant (validation dropped) → 1 red.

payments_py/mcp/core/server_manager.py
r-marques: Folded in `e86c8dc`, config-first kept (the documented escape hatch and the TS SDK's order), disagreement made loud: your message, verbatim in spirit, through the module logger and `onLog`. Tests: mismatch warns once and config wins; a matching config stays quiet. Mutants: warning dropped → 1 red; key-wins-over-config → 2 red. Thanks for measuring the split (documents on one tier, verification on the other) — that is the sentence in the comment now.

PR #296, opened 2026-09-21
feat(mcp): advertise `authorization_response_iss_parameter_supported` for the named environments (RFC 9207)

### LINKED ISSUE #295: feat(mcp): advertise `authorization_response_iss_parameter_supported` for the named environments (not `custom`)
## Context

RFC 9207 §3 lets an authorization server advertise `authorization_response_iss_parameter_supported: true` in its RFC 8414 metadata; §2.4 has a client that sees it **reject** any authorization response lacking `iss`, and lets a client that does not see it skip the check. The Nevermined API's own document advertises it (verified live on `api.sandbox.nevermined.app`, 2026-09-21), and the web app returns `iss` = the API origin on every authorization response since nvm-monorepo#3532 (deployed v1.41.0). The SDK-hosted AS documents (`build_authorization_server_metadata` / `buildAuthorizationServerMetadata` and the OIDC configuration) do not advertise it, so a client that gates the check on the advertisement gets no mix-up protection when it discovers via the SDK document — which on this SDK, whose protected-resource documents point at the MCP server itself, is the primary discovery path for RFC 9728 clients too. Raised by the self-review panel on the issuer PRs (payments#464 / payments-py#292).

## Decision (Rodolphe, 2026-09-21)

**Advertise it for the four named environments only** (`sandbox`, `live`, `staging_sandbox`, `staging_live` — the consent page is the Nevermined web app, which returns `iss`); **omit it for `custom`**, whose frontend may be an older or self-hosted web app that does not — advertising there would make RFC 9207 clients reject its responses for lacking `iss`. Declined: always advertising (the `custom` risk above); leaving it out (loses mix-up protection for gating clients).

## Scope

- Both AS-style documents (RFC 8414 and OIDC configuration) gain the field, `true` for named environments, absent for `custom`; the root server-info response is unchanged.
- Tests: the field is present and `true` for each named environment, absent for `custom`; the e2e that already pins `issuer` exactly pins this too.
- Docs: one clause in the discovery bullet of the MCP integration page.
- Land in lockstep with the twin (payments), after payments#464 / payments-py#292 merge (this is why it is not folded into them: both are approved and the change is additive).

r-marques:
## Description

Closes #295 — twin of nevermined-io/payments#466 (land in lockstep). Additive: one optional field on the two AS-style discovery documents.

RFC 9207 §3 lets an authorization server advertise `authorization_response_iss_parameter_supported: true` when it returns the `iss` parameter on every authorization response; §2.4 then has a client that sees the flag **reject** any response lacking `iss` (mix-up protection), and lets a client that does not see it skip the check. The Nevermined API's own document advertises it, and the web app has returned `iss` (= the canonical API origin these documents publish as `issuer` since #292) on every response since nvm-monorepo#3532 (deployed v1.41.0). The SDK-hosted AS / OIDC documents did not advertise it, so a client gating the check on the advertisement got no protection when discovering via this server — and on this SDK the SDK-hosted document is the *primary* discovery path for RFC 9728 clients (`authorization_servers` names the MCP server itself; payments#465).

### Decision (Rodolphe, 2026-09-21)

Advertise it **for the four named environments only**. **Omit** it (absent — never `false`) for `custom`, whose frontend may be an older or self-hosted web app that does not return `iss`, and whenever `oauthUrls.authorizationUri` is overridden, since the consent page is then an AS this SDK cannot vouch for. An unknown environment name serves the sandbox documents and advertises what sandbox advertises.

### Folded from the self-review panel (see the posted review)

- **The "overridden `authorizationUri` ⇒ omit" rule is my conservative addition** to the recorded decision (named ⇒ advertise, `custom` ⇒ omit): the SDK cannot vouch that an AS it did not choose returns `iss` — even when the override happens to point at the Nevermined web app. Stated in the docstrings and on `authorizationUri`'s own doc; pinned.
- RFC 9207 §3 makes an omitted flag **default to `false`**, so the wording is now "omitted (which the RFC defines as false); never an explicit `false`" rather than "absent, not false"; the wire type is the field's boolean.
- Only `custom` omits — the duplicated effective-environment rule was inert and is gone. Python: the override predicate is the same non-empty-string rule `get_oauth_urls` applies (`None` is no override).
- Pinned: `custom` on a Nevermined backend **and** frontend still omits (the line is the environment name — a "helpful" widening must be deliberate); an override pointing at the Nevermined web app omits, on both documents; `undefined`/`None`/`""` are no override; OIDC-side assertions for every override case; the e2e pins both documents.
- After the fold on `77abd33`: `poetry run pytest -m "not slow"`: **1185 passed, 1 skipped, 102 deselected**; doctests 7 passed; new mutants killed — custom-on-Nevermined advertises, webapp-override advertises, OIDC ignores overrides, `1` instead of `True`.

### What changed

- `payments_py/mcp/http/oauth_metadata.py` `_iss_parameter_support()`: a small helper decides the field; both `build…AuthorizationServerMetadata` and `build…OidcConfiguration` spread it. Types gain the optional field with a docstring saying when it is present.
- Docs: one clause in the MCP integration page's discovery bullet.

### Verification (runner named)

- `poetry run pytest -m "not slow"`: **1184 passed, 1 skipped, 102 deselected** on `865d156`; `tests/unit/mcp_tests/test_oauth_metadata.py` 110 passed; doctests 7 passed. Tests: all four named environments on both documents; `custom` absent; `authorizationUri` override absent while a `tokenUri`-only override stays advertised and an empty override is no override; unknown environment; the value is exactly `True` or absent.
- `black --check` clean; `mkdocs build` green.
- **Mutation-tested**: advertised for `custom` too → 1 red; consent override ignored → 1 red; `false` instead of absent → 3 red; OIDC document not advertising → 4 red.
- The e2e discovery test pins the served document (a named environment) advertises `true`.


payments_py/mcp/http/oauth_metadata.py
```diff
@@ -391,6 +391,31 @@ def get_oauth_urls(
...
+    older or self-hosted web app that does not return ``iss``) and whenever
+    ``oauthUrls.authorizationUri`` is overridden (the consent page is then an AS this
+    SDK knows nothing about). Absent, never ``False``: a client that sees no
+    advertisement skips the check rather than failing it. payments-py#295.
+    """
+    # Same effective-environment rule as ``_get_oauth_urls_for_environment``: an
+    # unknown name serves the sandbox documents, so it advertises what sandbox does.
+    effective = environment if environment in Environments else "sandbox"
+    consent_overridden = bool((overrides or {}).get("authorizationUri"))
+    if effective != "custom" and not consent_overridden:
```
r-marques: **🟡 SHOULD FIX — The documented `custom` recipe (localhost/proxy backend + `authorizationUri` pointed at the Nevermined web app**

The documented `custom` recipe (localhost/proxy backend + `authorizationUri` pointed at the Nevermined web app + `oauthUrls.issuer` set to the canonical origin) and the branded-subdomain `custom` (backend the SDK already classifies, frontend `https://nevermined.app`) can never advertise the flag and have no opt-in — on the SDK where the SDK-hosted document is the primary discovery path.

_(panel: code-reviewer; full detail was delivered to the author and is being folded.)_

<sub>🤖 pr-review-toolkit · code-reviewer · medium</sub>

tests/unit/mcp_tests/test_oauth_metadata.py
```diff
@@ -837,3 +837,121 @@ def test_override_hygiene_none_and_empty_never_replace_a_computed_value(self):
...
+        assert (
+            "authorization_response_iss_parameter_supported"
+            not in build_authorization_server_metadata(config)
+        )
+        assert (
+            "authorization_response_iss_parameter_supported"
+            not in build_oidc_configuration(config)
+        )
+
+    def test_overridden_consent_endpoint_is_not_vouched_for(self):
```
r-marques: **🟡 SHOULD FIX — The overridden-`authorizationUri` arm is pinned on the AS document only; the OIDC document's second call site**

The overridden-`authorizationUri` arm is pinned on the AS document only; the OIDC document's second call site (`oauth_metadata.py:573`) survives a mutant that drops the overrides argument, so a false RFC 9207 advertisement on `/.well-known/openid-configuration` would ship green.

_(panel: pr-test-analyzer; full detail was delivered to the author and is being folded.)_

<sub>🤖 pr-review-toolkit · pr-test-analyzer · medium</sub>

tests/unit/mcp_tests/test_oauth_metadata.py
```diff
@@ -837,3 +837,121 @@ def test_override_hygiene_none_and_empty_never_replace_a_computed_value(self):
...
+            is True
+        )
+
+    def test_custom_omits_the_key_entirely(self, monkeypatch):
+        monkeypatch.setitem(
+            Environments,
+            "custom",
+            EnvironmentInfo(
+                frontend="https://nevermined.app",
+                backend="http://localhost:3001",
```
r-marques: **🟡 SHOULD FIX — `custom` is pinned as "omitted" only with a loopback backend; a `custom` whose backend (or `tokenUri` override**

`custom` is pinned as "omitted" only with a loopback backend; a `custom` whose backend (or `tokenUri` override) classifies as a Nevermined API host has no row, so a mutant advertising the flag whenever the tier resolves survives the suite.

_(panel: pr-test-analyzer; full detail was delivered to the author and is being folded.)_

<sub>🤖 pr-review-toolkit · pr-test-analyzer · medium</sub>

payments_py/mcp/http/oauth_metadata.py
```diff
@@ -391,6 +391,31 @@ def get_oauth_urls(
...
+    advertises the same flag. It is OMITTED for ``custom`` (the frontend may be an
+    older or self-hosted web app that does not return ``iss``) and whenever
+    ``oauthUrls.authorizationUri`` is overridden (the consent page is then an AS this
+    SDK knows nothing about). Absent, never ``False``: a client that sees no
+    advertisement skips the check rather than failing it. payments-py#295.
+    """
+    # Same effective-environment rule as ``_get_oauth_urls_for_environment``: an
+    # unknown name serves the sandbox documents, so it advertises what sandbox does.
+    effective = environment if environment in Environments else "sandbox"
+    consent_overridden = bool((overrides or {}).get("authorizationUri"))
```
r-marques: **💡 NIT — `bool(overrides.get("authorizationUri"))` is a second definition of "overridden" that disagrees with `get_oaut**

`bool(overrides.get("authorizationUri"))` is a second definition of "overridden" that disagrees with `get_oauth_urls`'s `clean` predicate (`isinstance(v, str) and v != ""`) and with the TS twin: a truthy non-str value publishes the Nevermined authorize URL AND omits the flag.

_(panel: code-reviewer; full detail was delivered to the author and is being folded.)_

<sub>🤖 pr-review-toolkit · code-reviewer · medium</sub>

payments_py/mcp/http/oauth_metadata.py
```diff
@@ -391,6 +391,31 @@ def get_oauth_urls(
...
+    ``issuer``) on every response since nvm-monorepo#3532 — the API's own document
+    advertises the same flag. It is OMITTED for ``custom`` (the frontend may be an
+    older or self-hosted web app that does not return ``iss``) and whenever
+    ``oauthUrls.authorizationUri`` is overridden (the consent page is then an AS this
+    SDK knows nothing about). Absent, never ``False``: a client that sees no
+    advertisement skips the check rather than failing it. payments-py#295.
+    """
+    # Same effective-environment rule as ``_get_oauth_urls_for_environment``: an
+    # unknown name serves the sandbox documents, so it advertises what sandbox does.
+    effective = environment if environment in Environments else "sandbox"
```
r-marques: **💡 NIT — The effective-environment rule is now copied verbatim from `_get_oauth_urls_for_environment` (l.311-313) with**

The effective-environment rule is now copied verbatim from `_get_oauth_urls_for_environment` (l.311-313) with a comment promising they agree; a one-line `_effective_environment()` helper used by both removes the drift instead of documenting it.

_(panel: code-reviewer; full detail was delivered to the author and is being folded.)_

<sub>🤖 pr-review-toolkit · code-reviewer · medium</sub>

tests/unit/mcp_tests/test_oauth_metadata.py
r-marques: **💡 NIT — The override matrix pins `tokenUri`-only and `authorizationUri` but not `oauthUrls.issuer` on a named environm**

The override matrix pins `tokenUri`-only and `authorizationUri` but not `oauthUrls.issuer` on a named environment, where the flag stays `True` next to an `issuer` the web app's `iss` will not equal — defensible, but unpinned and undocumented in the helper's docstring, so the next reader can "fix" it either way silently.

_(panel: code-reviewer; full detail was delivered to the author and is being folded.)_

<sub>🤖 pr-review-toolkit · code-reviewer · medium</sub>

payments_py/mcp/types/http_types.py
```diff
@@ -273,6 +277,7 @@ class AuthorizationServerMetadata(TypedDict, total=False):
     scopes_supported: List[str]
     token_endpoint_auth_methods_supported: List[str]
     subject_types_supported: List[str]
+    authorization_response_iss_parameter_supported: bool
```
r-marques, line 290: **💡 NIT — Typed `bool` while the docstring, the tests and the TS twin (`?: true`) all say "never `False`" — `Literal[Tru**

Typed `bool` while the docstring, the tests and the TS twin (`?: true`) all say "never `False`" — `Literal[True]` states the contract the tests enforce.

_(panel: code-reviewer; full detail was delivered to the author and is being folded.)_

<sub>🤖 pr-review-toolkit · code-reviewer · medium</sub>

payments_py/mcp/types/http_types.py
```diff
@@ -308,6 +313,7 @@ class OidcConfiguration(TypedDict, total=False):
...
```
r-marques, line 329: **💡 NIT — `OidcConfiguration` gains the field but its `Attributes:` docstring block (l.282-300) was not updated, unlike**

`OidcConfiguration` gains the field but its `Attributes:` docstring block (l.282-300) was not updated, unlike `AuthorizationServerMetadata`'s (l.263-266).

_(panel: code-reviewer; full detail was delivered to the author and is being folded.)_

<sub>🤖 pr-review-toolkit · code-reviewer · medium</sub>

tests/unit/mcp_tests/test_oauth_metadata.py
```diff
@@ -837,3 +837,121 @@ def test_override_hygiene_none_and_empty_never_replace_a_computed_value(self):
...
+            build_authorization_server_metadata(token_only)[
+                "authorization_response_iss_parameter_supported"
+            ]
+            is True
+        )
+        # An empty override is no override.
+        empty = {
+            **self.BASE,
+            "environment": "sandbox",
+            "oauthUrls": {"authorizationUri": ""},
```
r-marques: **💡 NIT — `{"authorizationUri": None}` is not pinned, and nothing ties `_iss_parameter_support`'s override predicate (`o**

`{"authorizationUri": None}` is not pinned, and nothing ties `_iss_parameter_support`'s override predicate (`oauth_metadata.py:413`) to `get_oauth_urls`' `clean` filter — a mutant treating `None` as an override survives.

_(panel: pr-test-analyzer; full detail was delivered to the author and is being folded.)_

<sub>🤖 pr-review-toolkit · pr-test-analyzer · medium</sub>

r-marques (commented):

Twin of payments#467; same shape of findings: the `custom` boundary pinned only on loopback; the overridden-`authorizationUri` arm pinned on the AS document only (the OIDC call site survives a mutant); the override predicate is a second definition of "overridden" that disagrees with `get_oauth_urls`'s `clean` filter (`None`); the effective-environment rule is copied and inert for this predicate; the TypedDict says `bool` while the contract is "True or absent"; the `OidcConfiguration` docstring block was not updated; a `# type: ignore` names the wrong code.

**What I reviewed**
- Diff range `origin/main...HEAD` — 5 files, head `865d1568`.
- Panel: code-reviewer, pr-test-analyzer (delivered by file; worktree clean).

**Verdict:** 🟡 Mergeable with nits — 0 blockers, 3 should-fix, 9 nits.

### 🧭 Inline comments (12)

- 🟡 **SHOULD FIX** — The documented `custom` recipe (localhost/proxy backend + `authorizationUri` pointed at the Nevermined web app — `payments_py/mcp/http/oauth_metadata.py:414`
- 🟡 **SHOULD FIX** — The overridden-`authorizationUri` arm is pinned on the AS document only; the OIDC document's second call site — `tests/unit/mcp_tests/test_oauth_metadata.py:890`
- 🟡 **SHOULD FIX** — `custom` is pinned as "omitted" only with a loopback backend; a `custom` whose backend (or `tokenUri` override — `tests/unit/mcp_tests/test_oauth_metadata.py:875`
- 💡 **NIT** — `bool(overrides.get("authorizationUri"))` is a second definition of "overridden" that disagrees with `get_oaut — `payments_py/mcp/http/oauth_metadata.py:413`
- 💡 **NIT** — The effective-environment rule is now copied verbatim from `_get_oauth_urls_for_environment` (l.311-313) with — `payments_py/mcp/http/oauth_metadata.py:412`
- 💡 **NIT** — The override matrix pins `tokenUri`-only and `authorizationUri` but not `oauthUrls.issuer` on a named environm — `tests/unit/mcp_tests/test_oauth_metadata.py:890`
- 💡 **NIT** — Typed `bool` while the docstring, the tests and the TS twin (`?: true`) all say "never `False`" — `Literal[Tru — `payments_py/mcp/types/http_types.py:280`
- 💡 **NIT** — `OidcConfiguration` gains the field but its `Attributes:` docstring block (l.282-300) was not updated, unlike — `payments_py/mcp/types/http_types.py:316`
- 💡 **NIT** — `{"authorizationUri": None}` is not pinned, and nothing ties `_iss_parameter_support`'s override predicate (`o — `tests/unit/mcp_tests/test_oauth_metadata.py:918`
- 💡 **NIT** — An `authorizationUri` override that points AT the Nevermined web app (the documented escape-hatch shape) is no — `tests/unit/mcp_tests/test_oauth_metadata.py:895`
- 💡 **NIT** — `in (True, None)` is equality-based, so the test named "exactly True" passes on a `1` mutant — only the `is Tr — `tests/unit/mcp_tests/test_oauth_metadata.py:951`
- 💡 **NIT** — `# type: ignore[typeddict-item]` names the wrong code — mypy reports `arg-type` on that line and the ignore as — `tests/unit/mcp_tests/test_oauth_metadata.py:930`

tests/unit/mcp_tests/test_oauth_metadata.py
```diff
@@ -837,3 +837,121 @@ def test_override_hygiene_none_and_empty_never_replace_a_computed_value(self):
...
+            "authorization_response_iss_parameter_supported"
+            not in build_oidc_configuration(config)
+        )
+
+    def test_overridden_consent_endpoint_is_not_vouched_for(self):
+        overridden = {
+            **self.BASE,
+            "environment": "sandbox",
+            "oauthUrls": {
+                "authorizationUri": "https://custom-issuer.com/oauth/authorize"
```
r-marques: **💡 NIT — An `authorizationUri` override that points AT the Nevermined web app (the documented escape-hatch shape) is no**

An `authorizationUri` override that points AT the Nevermined web app (the documented escape-hatch shape) is not pinned; the docs promise omission there and a mutant that re-advertises for a known frontend survives.

_(panel: pr-test-analyzer; full detail was delivered to the author and is being folded.)_

<sub>🤖 pr-review-toolkit · pr-test-analyzer · medium</sub>

tests/unit/mcp_tests/test_oauth_metadata.py
```diff
@@ -837,3 +837,121 @@ def test_override_hygiene_none_and_empty_never_replace_a_computed_value(self):
...
+                backend="http://localhost:3001",
+                proxy="",
+                helicone_url="",
+            ),
+        )
+        for environment in ("sandbox", "custom"):
+            doc = build_authorization_server_metadata(
+                {**self.BASE, "environment": environment}
+            )
+            assert doc.get("authorization_response_iss_parameter_supported") in (
```
r-marques: **💡 NIT — `in (True, None)` is equality-based, so the test named "exactly True" passes on a `1` mutant — only the `is Tr**

`in (True, None)` is equality-based, so the test named "exactly True" passes on a `1` mutant — only the `is True` tests elsewhere kill it (measured); it also checks the AS document only.

_(panel: pr-test-analyzer; full detail was delivered to the author and is being folded.)_

<sub>🤖 pr-review-toolkit · pr-test-analyzer · medium</sub>

tests/unit/mcp_tests/test_oauth_metadata.py
```diff
@@ -837,3 +837,121 @@ def test_override_hygiene_none_and_empty_never_replace_a_computed_value(self):
...
+            build_authorization_server_metadata(empty)[
+                "authorization_response_iss_parameter_supported"
+            ]
+            is True
+        )
+
+    def test_unknown_environment_serves_sandbox_documents_flag_included(self):
+        config = {**self.BASE, "environment": "bogus"}
+        assert (
+            build_authorization_server_metadata(config)[  # type: ignore[typeddict-item]
```
r-marques: **💡 NIT — `# type: ignore[typeddict-item]` names the wrong code — mypy reports `arg-type` on that line and the ignore as**

`# type: ignore[typeddict-item]` names the wrong code — mypy reports `arg-type` on that line and the ignore as unused; sibling rows (:526, :773) use `arg-type`; cosmetic, no checker runs in CI.

_(panel: pr-test-analyzer; full detail was delivered to the author and is being folded.)_

<sub>🤖 pr-review-toolkit · pr-test-analyzer · medium</sub>

eruizgar91:

payments_py/mcp/http/oauth_metadata.py
```diff
@@ -502,6 +534,7 @@ def build_authorization_server_metadata(
         "scopes_supported": scopes,
         "token_endpoint_auth_methods_supported": ["client_secret_post"],
         "subject_types_supported": ["public"],
+        **_iss_parameter_support(config["environment"], config.get("oauthUrls")),
```
eruizgar91, line 585: **LOW** — `_iss_parameter_support` is annotated `-> Dict[str, bool]` (line 394), and spreading a non-literal `dict[str, bool]` into a `TypedDict` is something mypy cannot check. This adds **two new mypy errors** on the two spread sites:

```
payments_py/mcp/http/oauth_metadata.py:537: error: Unsupported type "dict[str, bool]" for ** expansion in TypedDict  [typeddict-item]
payments_py/mcp/http/oauth_metadata.py:580: error: Unsupported type "dict[str, bool]" for ** expansion in TypedDict  [typeddict-item]
```

Verified against this head with mypy 2.3.1 (`--python-version 3.10 --ignore-missing-imports`): the base commit `bd30292` reports exactly one error in this file (the pre-existing `endpoints` one at `build_server_info_response`); this head reports three. So these two are introduced here.

No runtime impact and `lint.yml` only runs `pre-commit` → `black`, so nothing goes red today — but it is the one place where the module loses the type-level link to the documents it builds, while `pyproject.toml` still declares `[tool.mypy]` with `disallow_untyped_defs`/`warn_return_any`. Every other builder in this file returns a properly typed `TypedDict` literal.

Bounded fix, verified to remove both errors while leaving the pre-existing one and keeping all 111 tests green:

```python
class _IssParameterSupport(TypedDict, total=False):
    authorization_response_iss_parameter_supported: bool


def _iss_parameter_support(
    environment: EnvironmentName, overrides: Optional[Dict[str, str]]
) -> _IssParameterSupport:
```

(`from typing_extensions import TypedDict`, the form `http_types.py:24` already uses.) It also makes the "the key is optional and only ever `True`" contract a type, which is currently only prose.

payments_py/mcp/http/oauth_metadata.py
```diff
@@ -391,6 +391,38 @@ def get_oauth_urls(
...
+    self-hosted web app and this SDK cannot tell; a ``custom`` server on a Nevermined
+    backend AND frontend is deliberately still omitted, the decision drew the line at
+    the environment name, not at a host guess — and whenever
+    ``oauthUrls.authorizationUri`` is overridden, since the consent page is then an AS
+    this SDK cannot vouch for, even when the value points at the Nevermined web app.
+    The override predicate is the same "non-empty string" rule ``get_oauth_urls``
+    applies to overrides.
+    """
+    raw = (overrides or {}).get("authorizationUri")
+    consent_overridden = isinstance(raw, str) and raw != ""
```
eruizgar91: **LOW** — `consent_overridden` re-derives "what counts as an override" (`isinstance(raw, str) and raw != ""`) as a second copy of `get_oauth_urls`'s `clean` filter, and the docstring above asserts the two agree ("the same *non-empty string* rule `get_oauth_urls` applies"). They do agree today — I checked `""`, `None` and a non-`str` all behave identically on both sides — so nothing fires; this is about the next edit, not this one.

The consequence of drift is specific and silent: if `clean` ever gains a `.strip()` (a whitespace-only `authorizationUri` is exactly the kind of value that shows up from an unset Helm/env template), the document would publish the *computed* Nevermined authorize URL while this predicate still reads the value as an override and withdraws the flag — or the reverse. The document and its RFC 9207 claim would disagree with no test failing, because both copies are individually correct.

One module-level predicate used by both removes the class instead of documenting it:

```python
def _is_override(value: object) -> bool:
    """An override replaces a computed URL only when it is a non-empty string (#292)."""
    return isinstance(value, str) and value != ""
```

then `clean = {k: v for k, v in (overrides or {}).items() if _is_override(v)}` and `consent_overridden = _is_override(raw)`. Same file, no new abstraction — and it pairs naturally with the typing fix above. (Your own self-review panel raised this as a nit on `865d156`; it is still open on `77abd33`, so flagging it as still-actionable rather than as new.)

eruizgar91 (commented):
Reviewed `77abd33` — the SDK-hosted AS / OIDC documents advertise RFC 9207 `authorization_response_iss_parameter_supported: true` for the four named environments and omit it for `custom` and an overridden `authorizationUri`. **The conditional-advertisement claim verifies end to end**: I checked the server actually returns `iss`, rather than taking the PR body's word for it. Two LOW findings inline, neither blocking behaviour.

### The correctness claim — verified against the deployed reality, not the repo

Advertising a capability the server does not implement is the failure mode that matters here (RFC 9207 §2.4: a client that sees the flag MUST reject a response without `iss`; and, independently of the flag, MUST reject any `iss` that does not match the discovered `issuer`). So:

- **All four Nevermined API discovery documents, fetched live today**, publish `authorization_response_iss_parameter_supported: true` and an `issuer` of `https://api.{sandbox,live}.nevermined.{app,dev}` — byte-identical to what this SDK publishes for `sandbox` / `live` / `staging_sandbox` / `staging_live` (I ran the builders and diffed the strings).
- **Both deployed web apps emit `iss` unconditionally.** Pulled and read the shipped bundles: `nevermined.app` (`_hooks-4w4SVqk9.js`) and `nevermined.dev` (`_hooks-DAe6iykV.js`) both contain the two redirect builders, and both set `iss` on the `?code=` success *and* the `?error=` denial, with the value = `new URL(backendUrl).origin` for the ceremony's tier. That is nvm-monorepo#3527/#3532 actually live on prod *and* staging — not inferred from `main`.

So the `true` arm is true for all four named environments, on both documents.

### The omission arms genuinely fire

Executed the real builders across the matrix (not just read the tests):

| config | `issuer` published | flag |
|---|---|---|
| `custom` (localhost) and `custom` on a Nevermined backend+frontend | backend origin | **absent** |
| `sandbox` + `authorizationUri` → third-party AS | `https://api.sandbox.nevermined.app` | **absent** |
| `sandbox` + `authorizationUri` → the Nevermined web app (the documented escape hatch) | `https://api.sandbox.nevermined.app` | **absent** |
| `sandbox` + `tokenUri`-only override | canonical | `True` |
| `sandbox` + `authorizationUri` `""` / `None` | canonical | `True` |
| unknown environment name | sandbox document | `True` (correct — the served document *is* the sandbox one) |

`server_manager` does pass `oauthUrls` through to `create_oauth_router` (`server_manager.py:476`), so the override arm fires on `payments.mcp.start()` too, not only on the direct-router path.

### Tests are not vacuous — mutation-tested

Ran four mutants against `tests/unit/mcp_tests/test_oauth_metadata.py` at this head; all killed: advertise for `custom` → 2 red; ignore the `authorizationUri` override → 1 red; publish an explicit `False` instead of omitting → 4 red; drop the spread from the OIDC builder → 5 red. `tests/unit/mcp_tests` 201 passed, doctests 7 passed, `black --check` clean. The e2e pair exercises the real `server_manager → create_oauth_router → JSONResponse` wiring, which is what proves the new key survives serialization (the handlers pass the raw dict, no `response_model`, so nothing can strip it).

### Considered and discarded

- **`oauthUrls.issuer` overridden on a named environment keeps `flag: true` next to an `issuer` the web app's `iss` will never equal.** Looks like a hole; it is not one *this PR opens*. RFC 9207 §2.4's comparison-on-presence is unconditional — the web app returns `iss` regardless of what this SDK advertises, so a conforming client rejects on the mismatch with or without the flag. Equally broken before and after, fails closed, and `oauthUrls` is operator configuration, not attacker input. Worth a docstring line and a pinned row (already on your own nit list), not a code change.
- **Twin parity (payments#467).** Same field, same two conditions, same predicate, same "omit, never explicit `false`", and the same `?: boolean` on the public interface — no semantic divergence between the SDKs.
- **Consistency with #292.** The override predicate here matches `get_oauth_urls`'s `clean` filter exactly, and this PR adds no URL construction or comparison, so #292's canonicalisation (`_origin_of` / `_canonical_nevermined_origin`, trailing-dot strip included) is neither bypassed nor reintroduced in an older shape.
- **CI.** `unit_integration`, `lint`, `deepagents_compat`, `lint-links`, `staged-mintlify-check` all pass. The red `e2e` is `test_a2a_client_e2e.py::test_send_message_and_stream` failing `402 Payment Required` on the shared test account — unrelated to this diff; the MCP OAuth e2e tests are in the 76 that passed.

eruizgar91 (approved):
🟡 **Approved with comments** — no blockers; 0 should-fix, 2 nit(s) left inline.

Inline review: https://github.com/nevermined-io/payments-py/pull/296#pullrequestreview-5269324219

payments_py/mcp/http/oauth_metadata.py
r-marques: Done as you wrote it (see the new head): `_IssParameterSupport(TypedDict, total=False)` from `typing_extensions`, returned by `_iss_parameter_support`. Verified with mypy 2.3.1 (`--python-version 3.10 --ignore-missing-imports`): this file went from 3 errors to 1 — the pre-existing `endpoints` one at `build_server_info_response` — so both `typeddict-item` errors on the spread sites are gone. Also folded the twin of the payments#467 finding you amended there: a named environment whose `oauthUrls.issuer` differs from the canonical origin is still published but `_warn_issuer_once`d (keyed on the value), with a test that goes red when the warning is removed — so the two SDKs stay value-for-value.

payments_py/mcp/http/oauth_metadata.py
r-marques: Folded: `_is_override(value)` is now the one predicate behind `get_oauth_urls`'s `clean` filter and `_iss_parameter_support`'s `consent_overridden`; the docstring points at it rather than asserting agreement. Thanks for flagging that the self-review nit was still open.

